PR #14, opened 2015-01-09
Finish support for list operator implementation

cmoesel:


Src/coffeescript/cql-execution/src/elm/query.coffee
```diff
@@ -28,6 +27,8 @@ module.exports.Without = class Without extends With
   exec: (ctx) ->
     !super(ctx)
 
+module.exports.Sort = class Sort extends UnimplementedExpression
+
```
rdingwell: Whats the purpose of putting this back in?  Are we waiting to see if Bryns translator changes bring this up to the surface?

Src/coffeescript/cql-execution/src/elm/list.coffee
```diff
@@ -60,59 +51,46 @@ module.exports.SingletonFrom = class SingletonFrom extends Expression
...
-      when typeIsArray container
-        return item in container
-      when container instanceof ValueSet
-        return container.hasCode item
-
-# TODO: Deconflict w/ definition in interval.coffee
-module.exports.Contains = class Contains extends UnimplementedExpression
+    src = @source.exec ctx
+    el = @element.exec ctx
+    if not src? or not el? then return null
```
rdingwell: nit picky but is it not possible to have a list with a null element in it?

cmoesel: Actually, that's a great point.  I was following the spec's guidance on `IndexOf` though:

> If either argument is null, the result is null.

So... we should leave it for now since it is according to spec, but I will add a ballot comment to consider changing it to allow for searching on null.  That said, I'm not sure there are too many use cases for it.

Src/coffeescript/cql-execution/src/elm/query.coffee
cmoesel: I'd like to leave it in so even if we don't implement it, we know what aspects of ELM are not surfaced in CQL and not supported in the engine.  Perhaps we should add comments to things like this?  Do you want me to do that before you merge?

PR #16, opened 2015-02-20
Interval operators

cmoesel:


Src/coffeescript/cql-execution/src/datatypes/interval.coffee
```diff
@@ -1,67 +1,212 @@
...
 
-  equals: (item) ->
-    if item instanceof Interval
-      [a, b] = [@toClosed(), item.toClosed()]
-      ThreeValuedLogic.and equals(a.low, b.low), equals(a.high, b.high)
+  areDateTimes = (x, y) ->
+    [x, y].every (z) -> z instanceof DateTime
+
+  areNumeric = (x, y) ->
+    [x, y].every (z) -> typeof z is 'number' or (z instanceof Uncertainty and typeof z.low is 'number')
```
rdingwell: I'm just being lazy and not looking but why does't areDateTimes look to see if z is an uncertainty with a low with type DateTime the way that areNumeric does. Does Uncertainty not support DateTime

cmoesel: Right.  Our implementation can never result in an Uncertainty containing DateTimes.  DateTimes in and of themselves contain uncertainty when they lack precision, but they're never represented using Uncertainty objects.

PR #21, opened 2015-05-18
Refactor gradle

cmoesel:
Refactored all of the Java projects to re-use as much gradle config as possible, hopefully making maintenance simpler.  In addition:
- Made static code analysis optional (default is off).  Turn it on for a build by passing `-Dsca` as an argument.
  - And fixed all static code analysis problems in all projects.
- Added support for uploading to Maven Central
- Changed group from _org.cqframework_ to _info.cqframework_
  - This was required by Maven Central (they won't let us post to _org.cqframework_ since we don't own the domain name).
- Changed version from _0.1_ to _0.1-SNAPSHOT_ to indicate it is a daily build (not a release).


cmoesel:
BTW-- you can see our first set of deployed snapshots here: https://oss.sonatype.org/content/groups/public/info/cqframework/

cmoesel:
I also just added support for incremental builds.  Now it only re-generates code if it notices the sources changed.  This improves the build-time on successive builds by a lot!

PR #25, opened 2015-08-25
Adding using definitions to Library object.  This information can be …

rdingwell:
…helpful to match an arbitray json document to a PatientSource provider that supports a given model.  For instance if the library in question is using QUICK then the included FHIR based patient source can be used , if a library is using model FOO then a PatientSource that knows how to deal with those types of records can be used.

### DIFF
--- a/Src/coffeescript/cql-execution/package.json
+++ b/Src/coffeescript/cql-execution/package.json
@@ -1,6 +1,6 @@
 {
   "name": "cql-execution",
-  "version": "0.0.1",
+  "version": "0.0.2",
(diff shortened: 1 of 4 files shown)


cmoesel:
When I run the unit tests, I get this error:

    1) CalculateAge should execute age in months:
    
      AssertionError: expected 422 to be 423

Is this just a bad unit test that is too dependent on timezone or daylight savings or something?  Do you get this?

cmoesel:
Nevermind regarding the broken test.  I see that you've fixed it in a subsequent PR.

PR #28, opened 2015-09-03
Very simplistic implimentation of Instance selectors .  This is only …

rdingwell:
…a stop gap until a more robust impl can be produced.  Instance type must come from either the System types such as Quantity or from types declared in models in the using statements.  The translator enforces this so you cannot just supply any type name.  Chances are that this will only work for a simple subset of property types within the definition.  The translator enforces types to be correct so if you attempt to use this in a nested fashion chances are it will break.

### DIFF
--- a/Src/coffeescript/cql-execution/src/elm/expressions.coffee
+++ b/Src/coffeescript/cql-execution/src/elm/expressions.coffee
@@ -7,6 +7,7 @@ conditional = require './conditional',
 datetime = require './datetime',
 declaration = require './declaration',
 external = require './external',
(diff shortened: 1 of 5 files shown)


cmoesel:
I will merge this, but suggest that a future implementation recognizes system types and constructs them using the existing system type implementations.  For example, Quantity should be recognized as a system type and constructed using the existing Quantity class.  This ensures that any special magic built around quantities continues to work as expected.

rdingwell:
I agree though I have been looking for the magic that should be around
Quantity and cannot find it.

Rob

PR #31, opened 2016-02-09
Fix issue where records were preserved across patients when processing queries

pkmitre:


cmoesel:
The Travis build seems to keep getting hung up either downloading the Rhino jar or executing the next step after that.  This didn't happen with the last PR (a couple of weeks ago).  I'll look into ways to fix it.

cmoesel:
BTW-- I fixed the travis issue on master by skipping the coffee script compilation and test steps (for cql-execution).  Not ideal, but there was no way to get around the timeouts (neither `travis_wait` nor `travis_retry`) had any effect.

Also, since there were changes on master, I had to rebase this branch on top of master.  But... it's all good now!

PR #33, opened 2016-02-17
Updates to support RESTful translation service

hadleynet:
Some of the commits are showing up twice, I think this is a result of my rebasing the translation_service branch on master today.


cmoesel:
I'll take a look.  Regarding the double commits, I expect this is because you mixed regular merges (see all the commits that start with "Merge branch") with rebase merges.  In general, you should try to avoid the regular merges (e.g., "git merge origin/master") and just pull with rebase instead (e.g., "git pull --rebase origin master").

Src/cql-lm/schema/elm/cqlannotations.xsd
```diff
@@ -22,4 +22,61 @@
...
+  </xs:simpleType>
+  <xs:complexType name="CqlToElmError">
+    <xs:annotation>
+      <xs:documentation>Represents CQL to ELM conversion errors</xs:documentation>
+    </xs:annotation>
+    <xs:complexContent>
+      <xs:extension base="Locator">
+        <xs:attribute name="library" type="xs:string" use="required">
+          <xs:annotation>
+            <xs:documentation>The line number in the CQL where the error started</xs:documentation>
```
cmoesel: This description ("The line number...") doesn't seem to match the attribute name ("library").

If it's really intended to correspond to the "line number", how does this differ from the base `Locator`'s `startLine` attribute?

If it's really intended to be "library", then I'd ask if it's really needed-- as the library will already be specified in the ELM document that contains the annotation (note that `Locator` already assumes current library).  If it's not intended to be the current library, then how does it differ from `targetIncludeLibraryId`?

Src/cql-lm/schema/elm/cqlannotations.xsd
```diff
@@ -22,4 +22,61 @@
...
+          </xs:annotation>
+        </xs:attribute>
+        <xs:attribute name="message" type="xs:string" use="required">
+          <xs:annotation>
+            <xs:documentation>A human-readable description of the error</xs:documentation>
+          </xs:annotation>
+        </xs:attribute>
+        <xs:attribute name="errorType" type="ErrorType" use="required">
+          <xs:annotation>
+            <xs:documentation>The character position in the line where the error
```
cmoesel: Clearly the wrong description ("The character position...")

Src/cql-lm/schema/elm/library.xsd
```diff
@@ -1,171 +1,176 @@
 <?xml version="1.0"?>
 <!-- This file defines the root library element and type for the ELM schema. -->
```
cmoesel: I can't tell what changed or did not change in this file (due to all the reformatting).  Did anything change?  If so, what?

hadleynet: I don't think anything changed, just the indenting and some line wrapping on the documentation.

PR #34, opened 2016-02-19
Adding support for QDM 4.2.

brynrhodes:
Added schema and model info file for QDM 4.2 to support measure creation for CQL-Based HQMF.


Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
```diff
@@ -0,0 +1,541 @@
...
+        <ns4:element name="reason" type="System.Concept"/>
+        <ns4:element name="method" type="System.Concept"/>
+        <ns4:element name="anatomicalLocationSite" type="System.Concept"/>
+    </ns4:typeInfo>
+    <ns4:typeInfo xsi:type="ns4:ClassInfo" name="QDM.ProcedureIntolerance" baseType="QDM.QDMBaseType">
+        <ns4:element name="startDatetime" type="System.DateTime"/>
+        <ns4:element name="stopDatetime" type="System.DateTime"/>
+        <ns4:element name="reaction" type="System.Concept"/>
+        <ns4:element name="ordinality" type="System.Concept"/>
+        <ns4:element name="ProcedurePerformed" type="System.Concept"/>
```
cmoesel: The `ProcedurePerformed` property looks out of place -- it doesn't reflect a property in QDM 4.2 and it is PascalCased (instead of camelCased).  Was this an intentional addition?  I think that the procedure to which the patient is intolerant would usually be represented by the default `code` property (part of `QDM.QDMBaseType`).  For example, CMS 147v5 uses "Occurrence A of Procedure, Intolerance: Influenza Vaccination".

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
```diff
@@ -0,0 +1,541 @@
...
+        <ns4:element name="method" type="System.Concept"/>
+    </ns4:typeInfo>
+    <ns4:typeInfo xsi:type="ns4:ClassInfo" name="QDM.LaboratoryTestPerformed" baseType="QDM.QDMBaseType"
+                  retrievable="true" label="Laboratory Test, Result" identifier="LaboratoryTestResult" primaryCodePath="code">
+        <ns4:element name="startDatetime" type="System.DateTime"/>
+        <ns4:element name="stopDatetime" type="System.DateTime"/>
+        <ns4:element name="negationRationale" type="System.Concept"/>
+        <ns4:element name="status" type="System.Concept"/>
+        <ns4:element name="method" type="System.Concept"/>
+        <ns4:element name="result" type="System.Concept"/>
```
cmoesel: Measures represent the `result` as both quantities _and_ concepts.  Examples:

- CMS 30v5: "Laboratory Test, Performed: LDL-c (result < 100 mg/dL)"
- CMS 188v5: "Laboratory Test, Performed: Bacterial culture (result: Positive Finding)"

So, to work with existing measures, we'll need `resultCode` *and* `resultQuantity` properties.

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
```diff
@@ -0,0 +1,541 @@
...
+        <ns4:element name="quantitativeResult" type="System.Quantity"/>
+        <ns4:element name="reason" type="System.Concept"/>
+        <ns4:element name="referenceRangeHigh" type="System.Concept"/>
+        <ns4:element name="referenceRangeLow" type="System.Concept"/>
```
cmoesel: Similar to `result`, the `referenceRangeHigh` and `referenceRangeLow` properties need to support both codes and quantities.

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
```diff
@@ -0,0 +1,541 @@
...
+    </ns4:typeInfo>
+    <ns4:typeInfo xsi:type="ns4:ClassInfo" name="QDM.MedicationAdministered" baseType="QDM.QDMBaseType">
+        <ns4:element name="startDatetime" type="System.DateTime"/>
+        <ns4:element name="stopDatetime" type="System.DateTime"/>
+        <ns4:element name="negationRationale" type="System.Concept"/>
+        <ns4:element name="dose" type="System.Quantity"/>
+        <ns4:element name="frequency" type="System.Concept"/>
+        <ns4:element name="route" type="System.Concept"/>
+        <ns4:element name="reason" type="System.Concept"/>
+        <ns4:element name="cumaltiveMedicationDuration" type="System.Quantity"/>
```
cmoesel: Typo.  Should be `cumulativeMedicationDuration`.

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
```diff
@@ -0,0 +1,541 @@
...
+        <ns4:element name="admissionDatetime" type="System.DateTime"/>
+        <ns4:element name="dischargeDatetime" type="System.DateTime"/>
+        <ns4:element name="lengthOfStay" type="System.Quantity"/>
+        <ns4:element name="dischargeStatus" type="System.Concept"/>
+        <ns4:element name="facilityLocation" type="System.Concept"/>
+        <ns4:element name="facilityLocationArrivalDatetime" type="System.DateTime"/>
+        <ns4:element name="facilityLocationDepartureDatetime" type="System.DateTime"/>
+        <ns4:element name="reason" type="System.Concept"/>
+        <ns4:element name="diagnosis" type="QDM.Diagnosis"/>
+        <ns4:element name="principalDiagnosis" type="QDM.Diagnosis"/>
```
cmoesel: Oooh.  You're breaking out of the flat model here.  I suppose that is intentional, but isn't really a true reflection of QDM 4.2.  In QDM 4.2, these would be `System.Concept`.  I don't think the MAT team has committed to supporting this yet...

Also, if you're going to break out of the flat model, why not do it elsewhere too (such as introducing a `FacilityLocation` type that contains `code`, `arrivalDatetime`, and `departureDatetime`)?

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
```diff
@@ -0,0 +1,541 @@
...
+    </ns4:typeInfo>
+    <ns4:typeInfo xsi:type="ns4:ClassInfo" name="QDM.DiagnosticStudyIntolerance" baseType="QDM.QDMBaseType">
+        <ns4:element name="startDatetime" type="System.DateTime"/>
+        <ns4:element name="stopDatetime" type="System.DateTime"/>
+        <ns4:element name="negationRationale" type="System.Concept"/>
+        <ns4:element name="reaction" type="System.Concept"/>
+        <ns4:element name="radiationDosage" type="System.Concept"/>
```
cmoesel: This isn't used in any measures right now, but it kind of seems more like a `System.Quantity` to me.

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
```diff
@@ -0,0 +1,541 @@
...
+        <ns4:element name="radiationDuration" type="System.Concept"/>
```
cmoesel: This isn't used in any measures right now, but it kind of seems more like a `System.Quantity` to me.

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
```diff
@@ -0,0 +1,541 @@
...
+    </ns4:typeInfo>
+    <ns4:typeInfo xsi:type="ns4:ClassInfo" name="QDM.PatientCharcteristicSex" baseType="System.Any">
```
cmoesel: Typo.  Should be "QDM.PatientCharacteristicSex".

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
```diff
@@ -0,0 +1,541 @@
...
+        <ns4:element name="startDatetime" type="System.DateTime"/>
+        <ns4:element name="stopDatetime" type="System.DateTime"/>
+    </ns4:typeInfo>
+    <ns4:typeInfo xsi:type="ns4:ClassInfo" name="QDM.PatientCharcteristic" baseType="System.Any"
```
cmoesel: Typo.  Should be "QDM.PatientCharacteristic"

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
```diff
@@ -0,0 +1,541 @@
...
+                  retrievable="true" label="Communication: From Provider to Patient" identifier="CommunicationFromProviderToPatient" primaryCodePath="code">
+        <ns4:element name="startDatetime" type="System.DateTime"/>
+        <ns4:element name="stopDatetime" type="System.DateTime"/>
+        <ns4:element name="negationRationale" type="System.Concept"/>
+    </ns4:typeInfo>
+    <ns4:typeInfo xsi:type="ns4:ClassInfo" name="QDM.TransferFrom" baseType="QDM.QDMBaseType">
+        <ns4:element name="startDatetime" type="System.DateTime"/>
+        <ns4:element name="stopDatetime" type="System.DateTime"/>
+        <ns4:element name="negationRationale" type="System.Concept"/>
+        <ns4:element name="reason" type="System.Concept"/>
```
cmoesel: The `reason` property is not in QDM 4.2.

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
```diff
@@ -0,0 +1,541 @@
...
+    </ns4:typeInfo>
+    <ns4:typeInfo xsi:type="ns4:ClassInfo" name="QDM.DiagnosticStudyPerformed" baseType="QDM.QDMBaseType">
+        <ns4:element name="startDatetime" type="System.DateTime"/>
+        <ns4:element name="stopDatetime" type="System.DateTime"/>
+        <ns4:element name="negationRationale" type="System.Concept"/>
+        <ns4:element name="reason" type="System.Concept"/>
+        <ns4:element name="result" type="System.Concept"/>
+        <ns4:element name="status" type="System.Concept"/>
+        <ns4:element name="radiationDosage" type="System.Concept"/>
```
cmoesel: This isn't used in any measures right now, but it kind of seems more like a `System.Quantity` to me.

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
```diff
@@ -0,0 +1,541 @@
...
+        <ns4:element name="radiationDuration" type="System.Concept"/>
```
cmoesel: This isn't used in any measures right now, but it kind of seems more like a `System.Quantity` to me.

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
```diff
@@ -0,0 +1,541 @@
...
+        <ns4:element name="method" type="System.Concept"/>
+        <ns4:element name="facilityLocation" type="System.Concept"/>
+    </ns4:typeInfo>
+    <ns4:typeInfo xsi:type="ns4:ClassInfo" name="QDM.InterventionOrder" baseType="QDM.QDMBaseType">
+        <ns4:element name="startDatetime" type="System.DateTime"/>
+        <ns4:element name="stopDatetime" type="System.DateTime"/>
+        <ns4:element name="negationRationale" type="System.Concept"/>
+        <ns4:element name="reason" type="System.Concept"/>
+    </ns4:typeInfo>
+    <ns4:typeInfo xsi:type="ns4:ClassInfo" name="QDM.PatientCharcteristicClinicalTrialParticipant" baseType="System.Any">
```
cmoesel: Typo.  Should be "PatientCharacteristicClinicalTrialParticipant".

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
pkmitre: I think that sticking with a flat model as close as possible to QDM 4.2 would align better with the general reasoning for sticking with QDM vs QUICK for this iteration. It would certainly simplify implementation in Bonnie.

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
```diff
@@ -0,0 +1,541 @@
...
+    <ns4:typeInfo xsi:type="ns4:ClassInfo" name="QDM.DeviceAllergy" baseType="QDM.QDMBaseType">
+        <ns4:element name="startDatetime" type="System.DateTime"/>
+        <ns4:element name="stopDatetime" type="System.DateTime"/>
+        <ns4:element name="reaction" type="System.Concept"/>
+    </ns4:typeInfo>
+    <ns4:typeInfo xsi:type="ns4:ClassInfo" name="QDM.ProcedureAdverseEvent" baseType="QDM.QDMBaseType">
+        <ns4:element name="startDatetime" type="System.DateTime"/>
+        <ns4:element name="stopDatetime" type="System.DateTime"/>
+        <ns4:element name="reaction" type="System.Concept"/>
+        <ns4:element name="ProcedurePerformed" type="System.Concept"/>
```
cmoesel: The `ProcedurePerformed` property looks out of place -- it doesn't reflect a property in QDM 4.2 and it is PascalCased (instead of camelCased).  Was this an intentional addition?  Like `Procedure, Intolerance`, I think that the procedure in which the adverse event occurred would usually be represented by the default `code` property (part of `QDM.QDMBaseType`) -- and the event itself would be the `reaction`.  That said, I agree that it's not clear in this case.

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
```diff
@@ -0,0 +1,541 @@
...
+    </ns4:typeInfo>
+    <ns4:typeInfo xsi:type="ns4:ClassInfo" name="QDM.FamilyHistory" baseType="QDM.QDMBaseType">
+        <ns4:element name="onsetAge" type="System.Quantity"/>
+        <ns4:element name="recordedDatetime" type="System.DateTime"/>
+        <ns4:element name="relationship" type="System.String"/>
```
cmoesel: This should be a `System.Concept`.  From the QDM 4.2 specification:

> If a relationship is specified, codes from the HL7 Personal Relationship Role Type value set (2.16.840.1.113883.1.11.19563) should be used to ensure compatibility with QRDA reporting constraints.

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
```diff
@@ -0,0 +1,541 @@
...
+        <ns4:element name="negationRationale" type="System.Concept"/>
+    </ns4:typeInfo>
+    <ns4:typeInfo xsi:type="ns4:ClassInfo" name="QDM.MedicationActive" baseType="QDM.QDMBaseType"
+                  retrievable="true" label="Medication, Active" identifier="MedicationActive" primaryCodePath="code">
+        <ns4:element name="startDatetime" type="System.DateTime"/>
+        <ns4:element name="stopDatetime" type="System.DateTime"/>
+        <ns4:element name="dose" type="System.Quantity"/>
+        <ns4:element name="frequency" type="System.Concept"/>
+        <ns4:element name="route" type="System.Concept"/>
+        <ns4:element name="cumaltiveMedicationDuration" type="System.Quantity"/>
```
cmoesel: Typo.  Should be "cumulativeMedicationDuration".

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
```diff
@@ -0,0 +1,541 @@
...
+        <ns4:element name="startDatetime" type="System.DateTime"/>
+        <ns4:element name="stopDatetime" type="System.DateTime"/>
+        <ns4:element name="reaction" type="System.Concept"/>
+    </ns4:typeInfo>
+    <ns4:typeInfo xsi:type="ns4:ClassInfo" name="QDM.DiagnosticStudyOrder" baseType="QDM.QDMBaseType">
+        <ns4:element name="startDatetime" type="System.DateTime"/>
+        <ns4:element name="stopDatetime" type="System.DateTime"/>
+        <ns4:element name="negationRationale" type="System.Concept"/>
+        <ns4:element name="reason" type="System.Concept"/>
+        <ns4:element name="radiationDosage" type="System.Concept"/>
```
cmoesel: This isn't used in any measures right now, but it kind of seems more like a `System.Quantity` to me.

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
```diff
@@ -0,0 +1,541 @@
...
+        <ns4:element name="radiationDuration" type="System.Concept"/>
```
cmoesel: This isn't used in any measures right now, but it kind of seems more like a `System.Quantity` to me.

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
```diff
@@ -0,0 +1,541 @@
...
+    <ns4:typeInfo xsi:type="ns4:ClassInfo" name="QDM.SubstanceOrder" baseType="QDM.QDMBaseType">
+        <ns4:element name="startDatetime" type="System.DateTime"/>
+        <ns4:element name="stopDatetime" type="System.DateTime"/>
+        <ns4:element name="negationRationale" type="System.Concept"/>
+        <ns4:element name="reason" type="System.Concept"/>
+        <ns4:element name="reaction" type="System.Concept"/>
+        <ns4:element name="dose" type="System.Quantity"/>
+        <ns4:element name="frequency" type="System.Concept"/>
+        <ns4:element name="method" type="System.Concept"/>
+        <ns4:element name="refills" type="System.Concept"/>
```
cmoesel: The `refills` property represents the number of refills, so is best expressed as a `System.Integer`.

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
```diff
@@ -0,0 +1,541 @@
...
+        <ns4:element name="route" type="System.Concept"/>
+    </ns4:typeInfo>
+    <ns4:typeInfo xsi:type="ns4:ClassInfo" name="QDM.LaboratoryTestAdverseEvent" baseType="QDM.QDMBaseType">
+        <ns4:element name="startDatetime" type="System.DateTime"/>
+        <ns4:element name="stopDatetime" type="System.DateTime"/>
+        <ns4:element name="reaction" type="System.Concept"/>
+        <ns4:element name="LaboratoryTestPerformed" type="QDM.LaboratoryTestPerformed"/>
```
cmoesel: Since I've seen a few things like this now, I'm starting to think this is intentional.  But here are a few notes:

- The `LaboratoryTestPerformed` property is not in QDM
- The use of PascalCase instead of camelCase seems a bit inconsistent
- Normally the test that had the adverse event would be represented by the `code` (although I admit that feels strange)
- In the other place you did this (with `ProcedureAdverseEvent` and `ProcedureIntolerance`), the property was represented as a `System.Concept`.  Here it is a reference to another type -- so there is some inconsistency here.

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
```diff
@@ -0,0 +1,541 @@
...
+    </ns4:typeInfo>
+    <ns4:typeInfo xsi:type="ns4:ClassInfo" name="QDM.SubstanceAdministered" baseType="QDM.QDMBaseType"
+                  retrievable="true" label="Substance, Administered" identifier="SubstanceAdministered" primaryCodePath="code">
+        <ns4:element name="startDatetime" type="System.DateTime"/>
+        <ns4:element name="stopDatetime" type="System.DateTime"/>
+        <ns4:element name="negationRationale" type="System.Concept"/>
+        <ns4:element name="dose" type="System.Quantity"/>
+        <ns4:element name="frequency" type="System.Concept"/>
+        <ns4:element name="route" type="System.Concept"/>
+        <ns4:element name="reason" type="System.Concept"/>
```
cmoesel: There is no `reason` property in QDM 4.2.

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
```diff
@@ -0,0 +1,541 @@
...
+        <ns4:element name="stopDatetime" type="System.DateTime"/>
+        <ns4:element name="negationRationale" type="System.Concept"/>
+        <ns4:element name="reason" type="System.Concept"/>
+        <ns4:element name="facilityLocation" type="System.Concept"/>
+    </ns4:typeInfo>
+    <ns4:typeInfo xsi:type="ns4:ClassInfo" name="QDM.LaboratoryTestIntolerance" baseType="QDM.QDMBaseType">
+        <ns4:element name="startDatetime" type="System.DateTime"/>
+        <ns4:element name="stopDatetime" type="System.DateTime"/>
+        <ns4:element name="reaction" type="System.Concept"/>
+        <ns4:element name="LaboratoryTestPerformed" type="QDM.LaboratoryTestPerformed"/>
```
cmoesel: I won't repeat myself-- but basically the same comment as `QDM.LaboratoryTestAdverseEvent`.

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
```diff
@@ -0,0 +1,541 @@
...
+        <ns4:element name="stopDatetime" type="System.DateTime"/>
+        <ns4:element name="negationRationale" type="System.Concept"/>
+        <ns4:element name="reason" type="System.Concept"/>
+        <ns4:element name="method" type="System.Concept"/>
+        <ns4:element name="result" type="System.Concept"/>
+        <ns4:element name="status" type="System.Concept"/>
+        <ns4:element name="anatomicalApproachSite" type="System.Concept"/>
+        <ns4:element name="anatomicalLocationSite" type="System.Concept"/>
+        <ns4:element name="ordinality" type="System.Concept"/>
+        <ns4:element name="radiationDuration" type="System.Concept"/>
```
cmoesel: This isn't used in any measures right now, but it kind of seems more like a `System.Quantity` to me.

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
```diff
@@ -0,0 +1,541 @@
...
+        <ns4:element name="radiationDosage" type="System.Concept"/>
```
cmoesel: This isn't used in any measures right now, but it kind of seems more like a `System.Quantity` to me.

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
```diff
@@ -0,0 +1,541 @@
...
+        <ns4:element name="reason" type="System.Concept"/>
+        <ns4:element name="method" type="System.Concept"/>
+    </ns4:typeInfo>
+    <ns4:typeInfo xsi:type="ns4:ClassInfo" name="QDM.DeviceRecommended" baseType="QDM.QDMBaseType">
+        <ns4:element name="startDatetime" type="System.DateTime"/>
+        <ns4:element name="stopDatetime" type="System.DateTime"/>
+        <ns4:element name="negationRationale" type="System.Concept"/>
+        <ns4:element name="reason" type="System.Concept"/>
+    </ns4:typeInfo>
+    <ns4:typeInfo xsi:type="ns4:ClassInfo" name="QDM.PatientCharcteristicExpired" baseType="System.Any">
```
cmoesel: Typo. Should be "QDM.PatientCharacteristicExpired"

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
```diff
@@ -0,0 +1,541 @@
...
+    </ns4:typeInfo>
+    <ns4:typeInfo xsi:type="ns4:ClassInfo" name="QDM.MedicationDischarge" baseType="QDM.QDMBaseType">
+        <ns4:element name="startDatetime" type="System.DateTime"/>
+        <ns4:element name="stopDatetime" type="System.DateTime"/>
+        <ns4:element name="negationRationale" type="System.Concept"/>
+        <ns4:element name="refills" type="System.Integer"/>
+        <ns4:element name="dose" type="System.Quantity"/>
+        <ns4:element name="frequency" type="System.Concept"/>
+        <ns4:element name="route" type="System.Concept"/>
+        <ns4:element name="reason" type="System.Concept"/>
```
cmoesel: The `reason` property is not in QDM 4.2.

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
```diff
@@ -0,0 +1,541 @@
...
+        <ns4:element name="cumaltiveMedicationDuration" type="System.Quantity"/>
```
cmoesel: Typo: Should be "cumulativeMedicationDuration".

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
```diff
@@ -0,0 +1,541 @@
...
+        <ns4:element name="status" type="System.Concept"/>
+    </ns4:typeInfo>
+    <ns4:typeInfo xsi:type="ns4:ClassInfo" name="QDM.LaboratoryTestRecommended" baseType="QDM.QDMBaseType">
+        <ns4:element name="startDatetime" type="System.DateTime"/>
+        <ns4:element name="stopDatetime" type="System.DateTime"/>
+        <ns4:element name="negationRationale" type="System.Concept"/>
+        <ns4:element name="method" type="System.Concept"/>
+        <ns4:element name="reason" type="System.Concept"/>
+    </ns4:typeInfo>
+    <ns4:typeInfo xsi:type="ns4:ClassInfo" name="QDM.PatientCharcteristicRace" baseType="System.Any"/>
```
cmoesel: Typo.  Should be "QDM.PatientCharcteristicRace".

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
```diff
@@ -0,0 +1,541 @@
...
+        <ns4:element name="stopDatetime" type="System.DateTime"/>
+        <ns4:element name="negationRationale" type="System.Concept"/>
+        <ns4:element name="refills" type="System.Integer"/>
+        <ns4:element name="dose" type="System.Quantity"/>
+        <ns4:element name="frequency" type="System.Concept"/>
+        <ns4:element name="route" type="System.Concept"/>
+        <ns4:element name="reason" type="System.Concept"/>
+        <ns4:element name="cumulativeMedicationDuration" type="System.Quantity"/>
+    </ns4:typeInfo>
+    <ns4:typeInfo xsi:type="ns4:ClassInfo" name="QDM.PatientCharcteristicPayer" baseType="System.Any">
```
cmoesel: Typo.  Should be "QDM.PatientCharacteristicPayer".

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
```diff
@@ -0,0 +1,541 @@
...
+        <ns4:element name="startDatetime" type="System.DateTime"/>
+        <ns4:element name="stopDatetime" type="System.DateTime"/>
+    </ns4:typeInfo>
+    <ns4:typeInfo xsi:type="ns4:ClassInfo" name="QDM.DiagnosticStudyAdverseEvent" baseType="QDM.QDMBaseType">
+        <ns4:element name="startDatetime" type="System.DateTime"/>
+        <ns4:element name="stopDatetime" type="System.DateTime"/>
+        <ns4:element name="negationRationale" type="System.Concept"/>
+        <ns4:element name="reaction" type="System.Concept"/>
+        <ns4:element name="radiationDosage" type="System.Concept"/>
```
cmoesel: This isn't used in any measures right now, but it kind of seems more like a `System.Quantity` to me.

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
```diff
@@ -0,0 +1,541 @@
...
+        <ns4:element name="radiationDuration" type="System.Concept"/>
```
cmoesel: This isn't used in any measures right now, but it kind of seems more like a `System.Quantity` to me.

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
```diff
@@ -0,0 +1,541 @@
...
+    </ns4:typeInfo>
+    <ns4:typeInfo xsi:type="ns4:ClassInfo" name="QDM.DiagnosticStudyRecommended" baseType="QDM.QDMBaseType">
+        <ns4:element name="startDatetime" type="System.DateTime"/>
+        <ns4:element name="stopDatetime" type="System.DateTime"/>
+        <ns4:element name="negationRationale" type="System.Concept"/>
+        <ns4:element name="reason" type="System.Concept"/>
```
cmoesel: The `reason` property is not in QDM 4.2.

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
```diff
@@ -0,0 +1,541 @@
...
+        <ns4:element name="radiationDosage" type="System.Concept"/>
+        <ns4:element name="radiationDuration" type="System.Concept"/>
+        <ns4:element name="method" type="System.Concept"/>
+    </ns4:typeInfo>
+    <ns4:typeInfo xsi:type="ns4:ClassInfo" name="QDM.ImmunizationOrder" baseType="QDM.QDMBaseType">
```
cmoesel: This is missing `signedDatetime`.

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
```diff
@@ -0,0 +1,541 @@
...
+    </ns4:typeInfo>
+    <ns4:typeInfo xsi:type="ns4:ClassInfo" name="QDM.PatientCareExperience" baseType="QDM.QDMBaseType">
+        <ns4:element name="startDatetime" type="System.DateTime"/>
+        <ns4:element name="stopDatetime" type="System.DateTime"/>
+    </ns4:typeInfo>
+    <ns4:typeInfo xsi:type="ns4:ClassInfo" name="QDM.TransferTo" baseType="QDM.QDMBaseType">
+        <ns4:element name="startDatetime" type="System.DateTime"/>
+        <ns4:element name="stopDatetime" type="System.DateTime"/>
+        <ns4:element name="negationRationale" type="System.Concept"/>
+        <ns4:element name="reason" type="System.Concept"/>
```
cmoesel: The `reason` property is not in QDM 4.2.

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
```diff
@@ -0,0 +1,541 @@
...
+    </ns4:typeInfo>
+    <ns4:typeInfo xsi:type="ns4:ClassInfo" name="QDM.ProviderCareExperience" baseType="QDM.QDMBaseType">
+        <ns4:element name="startDatetime" type="System.DateTime"/>
+        <ns4:element name="stopDatetime" type="System.DateTime"/>
+    </ns4:typeInfo>
+    <ns4:typeInfo xsi:type="ns4:ClassInfo" name="QDM.PatientCharcteristicEthnicity" baseType="System.Any"/>
```
cmoesel: Typo.  Should be "QDM.PatientCharacteristicEthnicity".

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
```diff
@@ -0,0 +1,541 @@
...
+    <ns4:typeInfo xsi:type="ns4:ClassInfo" name="QDM.SubstanceIntolerance" baseType="QDM.QDMBaseType">
+        <ns4:element name="startDatetime" type="System.DateTime"/>
+        <ns4:element name="stopDatetime" type="System.DateTime"/>
+        <ns4:element name="reaction" type="System.Concept"/>
+        <ns4:element name="frequency" type="System.Concept"/>
```
cmoesel: The `frequency` property is not in QDM 4.2.

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
```diff
@@ -0,0 +1,541 @@
...
+    <ns4:typeInfo xsi:type="ns4:ClassInfo" name="QDM.ProcedureOrder" baseType="QDM.QDMBaseType">
+        <ns4:element name="startDatetime" type="System.DateTime"/>
+        <ns4:element name="stopDatetime" type="System.DateTime"/>
+        <ns4:element name="negationRationale" type="System.Concept"/>
+        <ns4:element name="reason" type="System.Concept"/>
+        <ns4:element name="method" type="System.Concept"/>
+        <ns4:element name="anatomicalApproachSite" type="System.Concept"/>
+        <ns4:element name="anatomicalLocationSite" type="System.Concept"/>
+        <ns4:element name="ordinality" type="System.Concept"/>
+        <ns4:element name="radiationDuration" type="System.Concept"/>
```
cmoesel: This isn't used in any measures right now, but it kind of seems more like a `System.Quantity` to me.

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
```diff
@@ -0,0 +1,541 @@
...
+    </ns4:typeInfo>
+    <ns4:typeInfo xsi:type="ns4:ClassInfo" name="QDM.MedicationOrder" baseType="QDM.QDMBaseType">
```
cmoesel: This is missing `signedDatetime`.

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
```diff
@@ -0,0 +1,541 @@
...
+        <ns4:element name="startDatetime" type="System.DateTime"/>
+        <ns4:element name="stopDatetime" type="System.DateTime"/>
+        <ns4:element name="negationRationale" type="System.Concept"/>
+        <ns4:element name="refills" type="System.Integer"/>
+        <ns4:element name="dose" type="System.Quantity"/>
+        <ns4:element name="frequency" type="System.Concept"/>
+        <ns4:element name="route" type="System.Concept"/>
+        <ns4:element name="method" type="System.Concept"/>
+        <ns4:element name="reason" type="System.Concept"/>
+        <ns4:element name="cumaltiveMedicationDuration" type="System.Quantity"/>
```
cmoesel: Typo.  Should be "cumulativeMedicationDuration".

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
```diff
@@ -0,0 +1,541 @@
...
+    </ns4:typeInfo>
+    <ns4:typeInfo xsi:type="ns4:ClassInfo" name="QDM.PatientCharcteristicBirthdate" baseType="System.Any"
```
cmoesel: Typo.  Should be "QDM.PatientCharacteristicBirthdate".  Also, hopefully this is not needed anymore since `Patient` has a `birth datetime`.

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
```diff
@@ -0,0 +1,541 @@
...
+    </ns4:typeInfo>
+    <ns4:typeInfo xsi:type="ns4:ClassInfo" name="QDM.SubstanceRecommended" baseType="QDM.QDMBaseType">
+        <ns4:element name="startDatetime" type="System.DateTime"/>
+        <ns4:element name="stopDatetime" type="System.DateTime"/>
+        <ns4:element name="negationRationale" type="System.Concept"/>
+        <ns4:element name="reason" type="System.Concept"/>
+        <ns4:element name="dose" type="System.Quantity"/>
+        <ns4:element name="frequency" type="System.Concept"/>
+        <ns4:element name="method" type="System.Concept"/>
+        <ns4:element name="refills" type="System.Concept"/>
```
cmoesel: The `refills` property represents the number of refills, so is best expressed as a `System.Integer`.

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
```diff
@@ -0,0 +1,541 @@
...
+    </ns4:typeInfo>
+    <ns4:typeInfo xsi:type="ns4:ClassInfo" name="QDM.PhysicalExamPerformed" baseType="QDM.QDMBaseType"
+                  retrievable="true" label="Physical Exam, Performed" identifier="PhysicalExamPerformed" primaryCodePath="code">
+        <ns4:element name="startDatetime" type="System.DateTime"/>
+        <ns4:element name="stopDatetime" type="System.DateTime"/>
+        <ns4:element name="negationRationale" type="System.Concept"/>
+        <ns4:element name="reason" type="System.Concept"/>
+        <ns4:element name="method" type="System.Concept"/>
+        <ns4:element name="result" type="System.Concept"/>
+        <ns4:element name="quantitativeResult" type="System.Quantity"/>
```
cmoesel: OK, cool.  I see that you support both the concept and quantity versions of `result` here.  Personally, my preference would be to align more with how FHIR differentiates typed properties (via a suffix), so... `resultCode` (or `resultConcept`) and `resultQuantity`.

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
```diff
@@ -0,0 +1,541 @@
...
+    </ns4:typeInfo>
+    <ns4:typeInfo xsi:type="ns4:ClassInfo" name="QDM.FunctionalStatusRecommended" baseType="QDM.QDMBaseType">
+        <ns4:element name="startDatetime" type="System.DateTime"/>
+        <ns4:element name="stopDatetime" type="System.DateTime"/>
+        <ns4:element name="negationRationale" type="System.Concept"/>
+        <ns4:element name="reason" type="System.Concept"/>
+        <ns4:element name="method" type="System.Concept"/>
+    </ns4:typeInfo>
+    <ns4:typeInfo xsi:type="ns4:ClassInfo" name="QDM.LaboratoryTestPerformed" baseType="QDM.QDMBaseType"
+                  retrievable="true" label="Laboratory Test, Result" identifier="LaboratoryTestResult" primaryCodePath="code">
```
cmoesel: I've noticed that a _few_ of the definitions include the `retrievable`, `label`, `identifier`, and `primaryCodePath` attributes, but _most_ do not.  Shouldn't these be defined for all of them?

cmoesel:
In addition to the individual comments on the lines of code, I have one overarching comment for you to consider.  As the QDM classes are defined now, _most_ have a `negationRationale` property to indicate _the reason the thing was not done_.  While this matches the way the QDM 4.2 spec defines the data types, it results in an unexpected behavior: inclusion of _not done_ events in normal event queries.

For example, the retrieve `[ProcedurePerformed: "Amputation of Limb"]` would result in both _done_ and _not done_ procedures.  So in almost _every_ case, measure developers will have to add clauses to their queries to exclude _not done_ things (e.g., `[ProcedurePerformed: "Amputation of Limb"] P where P.code is null`).  Real QDM did not require this as it was _assumed_ that you only wanted positive assertions unless you stated otherwise -- so this will throw measure developers for a loop.

While we could provide a library of convenience methods to shield developers from this (e.g., `ProcedurePerformed("Amputation of Limb")` and `ProcedurePerformedNotDone("Amputation of Limb", "Patient Refused")), this is not ideal.  It's not very pretty, it's inefficient (since it can't bound the retrieves by dates), and it puts measure developers one step away from the real CQL.

Instead, I'd suggest that each QDM class that has a `negationRationale` _remove_ the `negationRationale` and that a corresponding _NotDone_ class be defined (with a `reason` property).  This will make authoring measures less error prone and more readable.

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
ChrisMarkle: There was a typo. The attribute was to be named procedurePerformed of type ProcedurePerformed.

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
ChrisMarkle: Thank You

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
ChrisMarkle: As I reviewed both QDM and QDM-based-HQMF, I found QDM-based-HQMF seemed to allow and REQUIRE these embedded datatypes as attributes. This appears to be a conflict between the two specs.

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
ChrisMarkle: What are you referring to when you say "this"?

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
ChrisMarkle: I'm not sure. So I left it as a concept.

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
ChrisMarkle: Thank You

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
cmoesel: It is true that the QDM-based HQMF used a non-flat representation, but the authoring format (QDM) has always been flat.  My assumption was that the QDM 4.2 based data model would stick to the same basic constraints as the QDM 4.2 authoring language.

If you choose to make the data model non-flat, you should make sure that the MAT and Bonnie teams  are aware and on-board, as they need to develop a UI and implementation that supports it.

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
ChrisMarkle: Thank You

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
ChrisMarkle: Thank You

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
ChrisMarkle: Thank You

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
ChrisMarkle: Another QDM mismatch with QDM-based-HQMF

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
ChrisMarkle: Another QDM mismatch with QDM-based-HQMF

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
ChrisMarkle: Another QDM mismatch with QDM-based-HQMF

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
ChrisMarkle: Thank You

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
ChrisMarkle: Another QDM mismatch with QDM-based-HQMF

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
ChrisMarkle: Another QDM mismatch with QDM-based-HQMF

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
ChrisMarkle: Thank You

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
ChrisMarkle: Thank You

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
ChrisMarkle: The creation of the 4.2 modelinfo was not intended to create a standard or
re-define anything. The MAT will not need to change anything. We just
wanted to create a representation of what is so that we could write and
verify existing measures using CQL and QDM. Any official measures will be
written using a different version of QDM (tentatively called QDM 5.0).

In the mean time we are investing the standards to see if these
discrepancies need to be addressed by a change in the standards.

On Mon, Feb 22, 2016 at 12:56 PM, Chris Moesel <notifications@github.com>
wrote:

> In Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
> <https://github.com/cqframework/clinical_quality_language/pull/34#discussion_r53663881>
> :
>
> > +        <ns4:element name="anatomicalApproachSite" type="System.Concept"/>
> > +        <ns4:element name="anatomicalLocationSite" type="System.Concept"/>
> > +        <ns4:element name="ordinality" type="System.Concept"/>
> > +    </ns4:typeInfo>
> > +    <ns4:typeInfo xsi:type="ns4:ClassInfo" name="QDM.EncounterPerformed" baseType="QDM.QDMBaseType"
> > +                  retrievable="true" label="Encounter, Performed" identifier="EncounterPerformed" primaryCodePath="code">
> > +        <ns4:element name="admissionDatetime" type="System.DateTime"/>
> > +        <ns4:element name="dischargeDatetime" type="System.DateTime"/>
> > +        <ns4:element name="lengthOfStay" type="System.Quantity"/>
> > +        <ns4:element name="dischargeStatus" type="System.Concept"/>
> > +        <ns4:element name="facilityLocation" type="System.Concept"/>
> > +        <ns4:element name="facilityLocationArrivalDatetime" type="System.DateTime"/>
> > +        <ns4:element name="facilityLocationDepartureDatetime" type="System.DateTime"/>
> > +        <ns4:element name="reason" type="System.Concept"/>
> > +        <ns4:element name="diagnosis" type="QDM.Diagnosis"/>
> > +        <ns4:element name="principalDiagnosis" type="QDM.Diagnosis"/>
>
> It is true that the QDM-based HQMF used a non-flat representation, but the
> authoring format (QDM) has always been flat. My assumption was that the QDM
> 4.2 based data model would stick to the same basic constraints as the QDM
> 4.2 authoring language.
>
> If you choose to make the data model non-flat, you should make sure that
> the MAT and Bonnie teams are aware and on-board, as they need to develop a
> UI and implementation that supports it.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/cqframework/clinical_quality_language/pull/34/files#r53663881>
> .
>



-- 
Chris Markle
ESAC Inc Analyst
618.979.4421

cmoesel: If that is the case, then I still suggest that using `System.Concept` would be a more accurate reflection of QDM 4.2, even for experimental QDM-on-CQL authoring.

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
cmoesel: By "this", I mean the `radiationDosage` in the line above my comment.  I would think that things like dosage and duration would most naturally be represented as a `System.Quantity` (e.g., a dosage of _100 mSv_, a duration of _5 days_).

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
cmoesel: It may not matter if you're just looking to convert existing measures, but see my reasoning above for why I think `System.Quantity` is better.

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
cmoesel: Yeah. I _think_ that the `reason` element in HQMF is mainly used in conjuction with the negation indicator.  If a measure uses the "not done for a reason" concept (a.k.a., _negation rationale), then HQMF sets the `negInd` flag to `true` and then puts the reason not done in the `reason`.

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
cmoesel: See my previous comment about `reason` and the negation indicator.

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
cmoesel: See my previous comment about `reason` and the negation indicator.

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
cmoesel: See my previous comment about `reason` and the negation indicator.

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.2.xml
cmoesel: OK, but the point is that this idea of intolerance _frequency_ is not available to measure authors now (since they author in QDM, not HQMF).

cmoesel:
@brynrhodes and @ChrisMarkle: I'll try to take a look at this pull request and merge it tomorrow.  Sorry for the long wait...

cmoesel:
OK.  I've reviewed it.  Some of my original comments still haven't been addressed-- but I will assume that it is intentional, so I won't let it hold up the merge into master.  After all, you guys own this effort-- I was just adding my feedback.  Just for clarity, however, here is a summary of my comments that still apply to the current code (just in case you didn't mean to miss some of them):

* You've broken out of the flat QDM model -- allowing nested complex types.  I _like_ this overall, but it doesn't reflect QDM 4.2.  If we know that the version that goes to production will support nested types, then fine -- but if it won't, then you'll confuse measure developers by giving them this power and then taking it away.  Note that there is also a question of vendor support: for example, in QDM 4.2, the only thing an author can say about a procedure to which a patient is intolerant is the code of the procedure.  In the nested model, an author can say all sorts of things about the procedure to which the patient is intolerant to.  The question is, what of those "all sorts of things" does the vendor support?
* The `radiationDosage` and `radiationDuration` and properties are still concepts in your model, although I think they should be quantities.
* The `referenceRangeHigh` and `referenceRangeLow` properties are still concepts in your model, even though I think they should be quantities.
* The following classes have a `reason` attribute even though QDM 4.2 doesn't specify one: `DiagnosticStudyRecommended`,`MedicationDischarge`,`SubstanceAdministered`,`TransferFrom`, `TransferTo`.  I know that the QDM-based HQMF templates support _reason_, but that it is only because it is needed to represent `negation rationale`.  Since you already have `negationRationale` in these class definitions, there is no need for `reason` too.
* `SubstanceIntolerance` does not have a `frequency` in QDM 4.2.

In regards to _negation rationale_, I see that you updated the examples to check for negation rationale being null.  Was that the final decision?  It still concerns me because it is very _unlike_ what authors are used to, and I expect they will find it an unwelcome complication and/or they won't remember to do it and will cause errors.  CQL is supposed to be _easier_, _more readable_, and _less error prone_.  Personally, I don't think that forcing this defensive use of negation rationale helps us achieve that goal -- in fact, I believe it detracts from it.  That said, I won't hold up the merge for it.

There is one thing, however, that is keeping me from merging this into master.  Yesterday Chris checked in some updated CQL documents reflecting the new style -- but the corresponding ELM XML documents were not updated.  As a result, the CQL and ELM examples are out of sync.  Can @ChrisMarkle or @brynrhodes please update the matching ELM documents so there is consistency?

Also, I noticed this new style of prefixing variables with '$'.  You've applied that style to the QDM CQL _and_ the QUICK CQL.  Is this intended to be a general CQL style or a QDM-specific CQL style?  If QDM-specific then, while I'm not personally a fan, I can see the argument that was probably made for it.  That said, if it's a QDM-specific style, it _shouldn't_ be applied to non-QDM CQL.  If it's being suggested as a more general CQL style, then I'd like to be a part of that conversation if possible.

ChrisMarkle:
Regarding the bulleted points, there was lengthy discussion and
cross-checking but you are right. I will correct the xsd right away.

With regards to negationRationale, after lengthy discussion with many
groups the current consensus was to require explicit semantics.

I'll let Bryn address the rest.

On Thu, Mar 24, 2016 at 8:57 AM, Chris Moesel <notifications@github.com>
wrote:

> OK. I've reviewed it. Some of my original comments still haven't been
> addressed-- but I will assume that it is intentional, so I won't let it
> hold up the merge into master. After all, you guys own this effort-- I was
> just adding my feedback. Just for clarity, however, here is a summary of my
> comments that still apply to the current code (just in case you didn't mean
> to miss some of them):
>
>    - You've broken out of the flat QDM model -- allowing nested complex
>    types. I *like* this overall, but it doesn't reflect QDM 4.2. If we
>    know that the version that goes to production will support nested types,
>    then fine -- but if it won't, then you'll confuse measure developers by
>    giving them this power and then taking it away. Note that there is also a
>    question of vendor support: for example, in QDM 4.2, the only thing an
>    author can say about a procedure to which a patient is intolerant is the
>    code of the procedure. In the nested model, an author can say all sorts of
>    things about the procedure to which the patient is intolerant to. The
>    question is, what of those "all sorts of things" does the vendor support?
>    - The radiationDosage and radiationDuration and properties are still
>    concepts in your model, although I think they should be quantities.
>    - The referenceRangeHigh and referenceRangeLow properties are still
>    concepts in your model, even though I think they should be quantities.
>    - The following classes have a reason attribute even though QDM 4.2
>    doesn't specify one: DiagnosticStudyRecommended,MedicationDischarge,
>    SubstanceAdministered,TransferFrom, TransferTo. I know that the
>    QDM-based HQMF templates support *reason*, but that it is only because
>    it is needed to represent negation rationale. Since you already have
>    negationRationale in these class definitions, there is no need for
>    reason too.
>    - SubstanceIntolerance does not have a frequency in QDM 4.2.
>
> In regards to *negation rationale*, I see that you updated the examples
> to check for negation rationale being null. Was that the final decision? It
> still concerns me because it is very *unlike* what authors are used to,
> and I expect they will find it an unwelcome complication and/or they won't
> remember to do it and will cause errors. CQL is supposed to be *easier*, *more
> readable*, and *less error prone*. Personally, I don't think that forcing
> this defensive use of negation rationale helps us achieve that goal -- in
> fact, I believe it detracts from it. That said, I won't hold up the merge
> for it.
>
> There is one thing, however, that is keeping me from merging this into
> master. Yesterday Chris checked in some updated CQL documents reflecting
> the new style -- but the corresponding ELM XML documents were not updated.
> As a result, the CQL and ELM examples are out of sync. Can @ChrisMarkle
> <https://github.com/ChrisMarkle> or @brynrhodes
> <https://github.com/brynrhodes> please update the matching ELM documents
> so there is consistency?
>
> Also, I noticed this new style of prefixing variables with '$'. You've
> applied that style to the QDM CQL *and* the QUICK CQL. Is this intended
> to be a general CQL style or a QDM-specific CQL style? If QDM-specific
> then, while I'm not personally a fan, I can see the argument that was
> probably made for it. That said, if it's a QDM-specific style, it
> *shouldn't* be applied to non-QDM CQL. If it's being suggested as a more
> general CQL style, then I'd like to be a part of that conversation if
> possible.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/cqframework/clinical_quality_language/pull/34#issuecomment-200820595>
>



-- 
Chris Markle
ESAC Inc Analyst
618.979.4421

PR #35, opened 2016-03-29
Types operator support

hadleynet:
The convert function is now implemented in CoffeeScript to match the changes to the CQL-to-ELM converter. I also bumped the artifact versions to 0.2-SNAPSHOT to provide a stable target for the cql-to-elm translation service code.

### DIFF
--- a/Src/coffeescript/cql-execution/src/elm/quantity.coffee
+++ b/Src/coffeescript/cql-execution/src/elm/quantity.coffee
@@ -10,6 +10,10 @@ module.exports.Quantity = class Quantity extends Expression
 
   exec: (ctx) ->
     @
+  
+  toString: () ->
+    "#{@value}#{@unit}"
+  
   sameOrBefore: (other) ->
     if other instanceof Quantity and other.unit == @unit then @value <= other.value else null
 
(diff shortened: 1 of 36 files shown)


Src/coffeescript/cql-execution/src/elm/quantity.coffee
```diff
@@ -10,6 +10,10 @@ module.exports.Quantity = class Quantity extends Expression
 
   exec: (ctx) ->
     @
+  
+  toString: () ->
+    "#{@value}#{@unit}"
+
```
cmoesel: Does this match the spec?  According to page 161, the string representation of a quantity is: `(-)?#0.0# '<unit>'`.

Src/coffeescript/cql-execution/src/elm/quantity.coffee
```diff
@@ -29,6 +33,15 @@ clean_unit = (units) ->
 
 module.exports.createQuantity = (value,unit) ->
   new Quantity({value: value, unit: unit})
+  
+module.exports.parseQuantity = (str) ->
+  components = /(\d+\.?\d*)(.+)/.exec str
```
cmoesel: Again, based on the spec, I think this is not entirely correct.  Irregardless of format, one notable issue is that this would not properly work with negative quantities.

Src/coffeescript/cql-execution/src/elm/quantity.coffee
hadleynet: Fixed in https://github.com/cqframework/clinical_quality_language/pull/35/commits/4732809275ed36d328d657bb7ce014cd7403d37e

Src/coffeescript/cql-execution/src/elm/quantity.coffee
hadleynet: Fixed in https://github.com/cqframework/clinical_quality_language/pull/35/commits/4732809275ed36d328d657bb7ce014cd7403d37e

PR #52, opened 2016-06-14
clarify installation instructions.

ErichBSchulz:
I encounted an error on building due to a missing `tools.jar` because I had JRE but not the JDK


cmoesel:
Thanks for the PR with the clarification.  I'll merge it.

PR #62, opened 2016-08-23
Added profile types to model info to support negation functionality i…

brynrhodes:
…n QDM.

Add the QDM 5.0 draft modelinfo.
Fixed an issue with the model info loader not working properly with multiple versions of a model.


Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/QdmModelInfoProvider.java
```diff
@@ -8,8 +8,30 @@
...
-                ModelInfo.class);
+        switch (version) {
+            case "4.1.2":
+                return JAXB.unmarshal(QdmModelInfoProvider.class.getResourceAsStream("/gov/healthit/qdm/qdm-modelinfo.xml"),
+                        ModelInfo.class);
+            case "4.2":
+                return JAXB.unmarshal(QuickModelInfoProvider.class.getResourceAsStream("/gov/healthit/qdm/qdm-modelinfo-4.2.xml"),
+                        ModelInfo.class);
+            case "5.0":
+            default:
```
edeyoung: what about moving this logic:

```
               return JAXB.unmarshal(QuickModelInfoProvider.class.getResourceAsStream("/gov/healthit/qdm/qdm-modelinfo-5.0.xml"),
                        ModelInfo.class);
```

under `case "5.0":` and then making the logic under `default:` be an error. This way, if you get an unsupported version, you can fail gracefully.

Src/java/elm/src/main/java/org/cqframework/cql/elm/tracking/ProfileType.java
```diff
@@ -0,0 +1,25 @@
+package org.cqframework.cql.elm.tracking;
+
+import java.util.Collection;
+
+/**
+ * Created by Bryn on 8/22/2016.
```
edeyoung: Based on the exchange with Moesel, it would be nice to have some comments here explaining what a `ProfileType` is and when is should be used.

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-5.0.xml
```diff
@@ -1,391 +1,461 @@
...
+        <ns4:element name="negationRationale" type="System.Code"/>
     </ns4:typeInfo>
     <ns4:typeInfo xsi:type="ns4:ClassInfo" name="QDM.PatientCharacteristicSex" retrievable="false" baseType="QDM.QDMBaseType"/>
     <ns4:typeInfo xsi:type="ns4:ClassInfo" name="QDM.EncounterActive" retrievable="false" baseType="QDM.QDMBaseType">
-        <ns4:element name="relevantPeriod" type="DateTimeInterval"/>
-        <ns4:element name="reason" type="Code"/>
-        <ns4:element name="facilityLocation" type="Code"/>
-        <ns4:element name="locationPeriod" type="DateTimeInterval"/>
-        <ns4:element name="lengthOfStay" type="Quantity"/>
+        <ns4:element name="relevantPeriod" type="interval&lt;System.DateTime&gt;"/>
```
edeyoung: is there a different approach that could be used to signify an interval? The escaped characters make it hard to read in this format.

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/QdmModelInfoProvider.java
brynrhodes: The intent was to allow 5.0 to be the default when no version is specified, but you make a good point, this would also change invalid versions to be 5.0 as well. I'll make the change, thanks!

Src/java/elm/src/main/java/org/cqframework/cql/elm/tracking/ProfileType.java
brynrhodes: Agreed, I will add.

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-5.0.xml
brynrhodes: While I agree it's difficult to read, it would be a pretty far-reaching change. Perhaps submit an issue to make the change?

PR #78, opened 2016-10-05
Added libraryLine to all tests

ChrisMarkle:
I was using these files to test CQL Online. CQL Online requires a library line.

### DIFF
--- a/Src/java/cql-to-elm/src/test/resources/org/cqframework/cql/cql2elm/DateTimeLiteralTest.cql
+++ b/Src/java/cql-to-elm/src/test/resources/org/cqframework/cql/cql2elm/DateTimeLiteralTest.cql
@@ -1,3 +1,5 @@
+library DateTimeLiteralTest version '2'
+
 define YearLiteral: @2014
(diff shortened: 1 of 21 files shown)


cmoesel:
What is the purpose of adding the library line to all of the tests?  When the test data is built, one of the functions of the CoffeeScriptTestDataGenerator is to insert a library line automatically, if it is missing.  See:
https://github.com/cqframework/clinical_quality_language/blob/master/Src/java/cql-to-elm/src/test/java/org/cqframework/cql/cql2elm/utilities/CoffeeScriptTestDataGenerator.java#L135-L137

brynrhodes:
CQL doesn't require a library line, so I'd ask the same question as Chris, what's the purpose of adding the library line here?

cmoesel:
Is it OK to close this PR? Unless there is a convincing argument to keep the library lines in, I don't think they're necessary.

ChrisMarkle:
I agree. Close the PR.

Chris Markle
ESAC Inc Analyst
618.979.4421

PR #81, opened 2016-12-05
Cql exec frmwrk examples

dwwinters:
CQL execution framework examples broken since their inception two years ago.  Incorporated minor fixes to get them running again.  Biggest changes to the framework in the past two years appears to be the introduction of the DateTime type and the executor class.  Adjusting the examples to use these went a long way towards getting them running again.


Src/coffeescript/cql-execution/src/elm/clinical.coffee
```diff
@@ -11,7 +11,7 @@ module.exports.ValueSetDef = class ValueSetDef extends Expression
     #todo: code systems and versions
 
   exec: (ctx) ->
-    valueset = ctx.codeService.findValueSet(@id, @version) ? new ValueSet(@id, @version)
+    valueset = ctx._codeService.findValueSet(@id, @version) ? new ValueSet(@id, @version)
```
cmoesel: It's odd that you needed to do that.  The `Context` class should be exposing a `codeService` getter so you don't have to access the private `_codeService` property directly.  See: https://github.com/cqframework/clinical_quality_language/blob/cccc2144b17c55d111631d9291c2e4e3f6b9b273/Src/coffeescript/cql-execution/src/runtime/context.coffee#L16-L18

Src/coffeescript/cql-execution/src/example/CMS146v2_CQM.cql
```diff
@@ -46,7 +46,7 @@ valueset "Ambulatory/ED Visit": '2.16.840.1.113883.3.464.1003.101.12.1061'
 valueset "Antibiotic Medications": '2.16.840.1.113883.3.464.1003.196.12.1001'
 valueset "Group A Streptococcus Test": '2.16.840.1.113883.3.464.1003.198.12.1012'
 
-parameter MeasurementPeriod Interval<DateTime>
+parameter MeasurementPeriod default Interval[DateTime(2013, 1, 1, 0, 0, 0, 0), DateTime(2014, 1, 1, 0, 0, 0, 0))
```
cmoesel: What was the purpose of this change?  In theory, the previous version was completely valid.

Src/coffeescript/cql-execution/src/example/age.coffee
```diff
@@ -26,70 +26,80 @@ module.exports = {
                "highClosed" : false,
                "type" : "Interval",
                "low" : {
-                  "name" : "DateTime",
-                  "type" : "FunctionRef",
-                  "operand" : [ {
+                  "type" : "DateTime",
```
cmoesel: I noticed that you changed the ELM (age.coffee) but not the CQL (age.cql).  Was the ELM just out of date (not recently translated)?

dwwinters: Correct.  The ELM was out of date.

Src/coffeescript/cql-execution/src/example/CMS146v2_CQM.cql
dwwinters: The framework was bombing on line 97 of https://github.com/cqframework/clinical_quality_language/blob/7a5b59d747edee49a266107934cfd129209e56e7/Src/coffeescript/cql-execution/src/elm/clinical.coffee

The problem appeared to be that no measurement period was actually being specified.  Putting a default interval in the CQL seemed to be a reasonable solution (although not the only one).

cmoesel: I expect it was bombing because a parameter value for `MeasurementPeriod` wasn't supplied.  When a parameter is defined just using a type (with no default value), then you must provide a parameter value for it to the executor before executing it.

I suspect that if you revert it and add a `MeasurementPeriod` parameter to the executor (with a valid DateTime interval value), it should run fine.

Src/coffeescript/cql-execution/src/example/exec-cms146v2_CQM.coffee
```diff
@@ -97,5 +98,5 @@ psource = new cql.PatientSource [ {
 
 ctx = new cql.Context(lib, psource, cservice)
```
cmoesel: As discussed on IM, I think this line can be removed.

Src/coffeescript/cql-execution/src/example/exec-age.coffee
```diff
@@ -44,5 +45,5 @@ psource = new cql.PatientSource [ {
 
 ctx = new cql.Context(lib, psource)
```
cmoesel: As discussed on IM, I think this line can be removed.

cmoesel:
@dwwinters: I had some additional time, so I did some extra testing and addressed some of the issues we previously discussed.  Please review my changes and let me know if you see any issues.  Please also test locally since you initially had some trouble with the `ctx.codeService` in _clinical.coffee_.

dwwinters:
@cmoesel: Pulled your changes down and they worked on my installation (Node v4.4.4).  Not sure why c3a57cf4a5acd9eb2ea145b8c7982e19c2819117 now works, when yesterday it did not.

cmoesel:
@dwwinters : OK for me to merge then?  Or do you have further updates you want to get into the PR?

dwwinters:
No.  I have further changes.  Still running into issues with including the FHIRHelpers library.

cmoesel:
Closing this pull request as there's too much going on here.  Will consolidate, remove extraneous changes, and re-issue at a later date.

PR #85, opened 2017-02-27
Add Exp to system functions

cmoesel:
Add Exp to system functions so it is properly translated to the ELM Exp operator rather than a user-function.

v12 branch needs the same fix applied, but wasn't sure how you prefer it done (separate PR, merge from master to v12, etc).


brynrhodes:
I've been rebasing the v12 branch onto master as changes are made, so I'll merge this then do that there. Thanks Chris!

PR #87, opened 2017-03-02
Added simple semantic versioning support to the default library loader.

brynrhodes:


cmoesel:
The code looks OK to me, but there are no unit tests -- and I'm unable to manually test it myself at this time.  Can I assume that this has been tested and works?

brynrhodes:
Yes, I've tested it with a library that references another library without
a version and it retrieves the most recent version.

Regards,
Bryn Rhodes
bryn@databaseconsultinggroup.com

Issue #86, opened 2017-02-28
ELM Annotations Data Duplication


alexanderelliott121:
I'm on the MITRE Bonnie team.  I'm currently trying to create an html representation of the CQL that will have the appropriate data to enable logic highlighting.
My current attempts to derive cql from the XML ELM's annotations run into issues with duplicated information.

Does a strategy currently exist for re-creating CQL from the XML annotations and ignoring the duplication? If not, is this something that can be resolved during CQL-to-ELM generation?

Examples from CMS177:

1. Element references duplicated in full.
This snippet from the MDD Diagnosis annotation fully reproduces the contents of the MDD Encounter element.  See references 58 and 48 below.

    In addition, reference 48 contains the following:
    "MDD Encounter" Encounter     such that Diagnosis.prevalencePeriod overlaps Encounter.relevantPeriod**define "MDD Encounter":**

    See bolded portion.  This line seems incorrectly formatted, with the define statement being appended where it shouldn't be.

```
  <annotation xsi:type="a:Annotation">
    <a:s r="58">define "MDD Encounter": distinct (["Encounter, Performed": "Office Visit"] union ["Encounter, Performed": "Outpatient Consultation"] union ["Encounter, Performed": "Face-to-Face Interaction"] union ["Encounter, Performed": "Psych Visit - Diagnostic Evaluation"] union ["Encounter, Performed": "Psych Visit - Family Psychotherapy"] union ["Encounter, Performed": "Psych Visit - Psychotherapy"] union ["Encounter, Performed": "Psychoanalysis"]) Encounter    where Encounter.relevantPeriod during "Measurement Period"</a:s>
  </annotation>
```

``` 
<a:s r="49">                       
  <a:s r="48">"MDD Encounter" Encounter     such that Diagnosis.prevalencePeriod overlaps Encounter.relevantPerioddefine "MDD Encounter":
    <a:s r="47">distinct
      <a:s r="46">                                
        <a:s r="41">(
          <a:s r="40">                                      
            <a:s r="38">                                         
              <a:s r="36">                                            
                <a:s r="34">                                                
                  <a:s r="32">                                                  
                    <a:s r="30">                                                     
                      <a:s r="28">["Encounter, Performed": "Office Visit"]
                      </a:s> union
                    <a:s r="29">["Encounter, Performed": "Outpatient Consultation"]
                  </a:s>                                                  
                </a:s> union
[...]

Issue #61, opened 2016-08-19
Annotations in JSON


brynrhodes:
Annotations in XML work as expected because the content of the annotation is an html fragment. However, when the ELM is output as JSON, the rendering is very difficult to reconstruct. There needs to be a better way to communicate the content of the annotation when the output is formatted using JSON.

pkmitre:
The current XML annotation is difficult to represent in JSON because the XML uses mixed content, ie

    <annotation xsi:type="a:Annotation">
      <a:s r="27">
        define PatientAge:
        <a:s r="26">
          AgeInYearsAt(
          <a:s r="25">
            start of
            <a:s r="24">MeasurementPeriod</a:s>
          </a:s>
          )
        </a:s>
      </a:s>
    </annotation>

In the JSON, all of the text strings within a tag, regardless of ordering relationship to other neighbor tags, are simply represented as an array of strings, ie

                    "value": [
                      "AgeInYearsAt",
                      "(",
                      ")"
                    ]

The relationship of the strings to the other tags is not preserved. Would it work to simply avoid the use of mixed content? For example, something like

    <annotation xsi:type="a:Annotation">
      <a:s r="27">
        <a:s>define PatientAge:</a:s>
        <a:s r="26">
          <a:s>AgeInYearsAt(</a:s>
          <a:s r="25">
            <a:s>start of</a:s>
            <a:s r="24">MeasurementPeriod</a:s>
          </a:s>
          <a:s>)</a:s>
        </a:s>
      </a:s>
    </annotation>

This could be represented in unambiguous JSON.

brynrhodes:
This issue was corrected by pull request #94 and the fix will be included in the next release.

PR #94, opened 2017-03-21
Fix annotations

brynrhodes:
This change removes the link spanning from the narrative output to address issues #86 and #61. For #86, the problem is that the span-painting algorithm does not account for the fact that the parse tree will move back through the source in the case of in-fix operators, and so duplicates text output in the narrative. For #61, the problem is that the way spanning is currently done involves mixed content, which outputs poorly in the JSON format. This fix addresses both of these in the simplest way possible, by just removing the link output. So all narratives are now output as a single span. I will log another issue to request that the link-spanning functionality be added back in, but delay implementation until we get feedback that it is an important enough feature to warrant effort.

### DIFF
--- a/Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/Cql2ElmVisitor.java
+++ b/Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/Cql2ElmVisitor.java
@@ -149,7 +149,8 @@ public class Cql2ElmVisitor extends cqlBaseVisitor {
 
     private void pushNarrative(@NotNull ParseTree tree) {
         org.antlr.v4.runtime.misc.Interval sourceInterval = tree.getSourceInterval();
-        System.out.println(String.format("Push: %d..%d, Tree: %s", sourceInterval.a, sourceInterval.b, tree.getText()));
+        // Debugging line for Github issues #61 & 86
+        //System.out.println(String.format("Push: %d..%d, Tree: %s", sourceInterval.a, sourceInterval.b, tree.getText()));
 
         // If there is a parent narrative
         // add the text from the current text pointer to the start of the new source context to the narrative
@@ -158,7 +159,8 @@ public class Cql2ElmVisitor extends cqlBaseVisitor {
             org.antlr.v4.runtime.misc.Interval tokenInterval =
                     new org.antlr.v4.runtime.misc.Interval(currentToken, sourceInterval.a - 1);
             String content = tokenStream.getText(tokenInterval);
-            System.out.println(String.format("CurrentToken: %d, Content: %s", currentToken, content));
+            // Debugging line for Github issues #61 & 86
+            //System.out.println(String.format("CurrentToken: %d, Content: %s", currentToken, content));
             parentNarrative.getContent().add(content);
         }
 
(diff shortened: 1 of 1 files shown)


alexanderelliott121:
I've tested this against CMS146v2_CQM.cql, and run into the binary infix operator issue from https://github.com/cqframework/clinical_quality_language/issues/86

Key point is r=14 and its children, which have the leading 'AgeInYearsAt(start of MeasurementPeriod) >=' followed by elements containing the same data.

         <annotation xsi:type="a:Annotation">
            <a:s r="21">define &quot;InDemographic&quot;:
              <a:s r="20">
                  <a:s r="14">
                    AgeInYearsAt(start of MeasurementPeriod) >=
                    <a:s r="12">
                      AgeInYearsAt(
                      <a:s r="11">
                        start of 
                        <a:s r="10">MeasurementPeriod</a:s>
                      </a:s>
                      )
                    </a:s> 
                    >= 
                    <a:s r="13">2</a:s>
                  </a:s> 
                  and 
                  <a:s r="19">
                    AgeInYearsAt(start of MeasurementPeriod) &lt;
                    <a:s r="17">
                      AgeInYearsAt(
                      <a:s r="16">
                        start of 
                        <a:s r="15">MeasurementPeriod</a:s>
                        </a:s>
                        )
                      </a:s> 
                      &lt; 
                      <a:s r="18">18</a:s>
                  </a:s>
               </a:s>
            </a:s>
         </annotation>

brynrhodes:
Hmmm, so when I run this locally it works for me (I see a single span for all expressions). I did push additional changes to this branch today as part of this pull request, and the output you've posted looks like it would have prior to those changes. Can you verify you're running the latest on the fix-annotations branch?

cmoesel:
I can confirm Bryn's comment.  I pulled down the `fix-annotations` branch and ran CQL-to-ELM against CMS146v2_CQM.cql.  For the _InDemographic_ expression, I get the following annotation:

```xml
<annotation xsi:type="a:Annotation">
  <a:s>define &quot;InDemographic&quot;:
    AgeInYearsAt(start of MeasurementPeriod) >= 2 and AgeInYearsAt(start of MeasurementPeriod) &lt; 18</a:s>
</annotation>
```

Bryn, is this what you expect?  It seems to work for me.  That said, I question how useful this will be to Bonnie.  I expect Bonnie wants a higher level of fidelity than just annotations at the expression level.

brynrhodes:
Yes, that's what I expect to be output, it solves both the problems, although with admittedly less useful output. That's why I opened issue #95 to track whether or not we need to go to the extra effort of actually linking the source up to the nodes.

cmoesel:
@alexanderelliott121: Assuming you can get the cql-to-elm working as described above, does this change improve things for you, keep things about the same, or make them worse?  If either of the first two, I'm inclined to accept the PR.  If, however, it makes things worse, we should probably discuss.

edeyoung:
I'll let @alexanderelliott121 respond to if he sees the same output. Based on the output the @cmoesel noted, that won't be helpful for the Bonnie team. The most helpful thing would be to have all terminal nodes from the cql tree listed in the annotations with their associated elm logic components.

cmoesel:
Thanks, @edeyoung.  Then I guess the relevant question really is whether or not this update sets you back, or leaves you in the same place.  If the previous functionality was unusable anyway, I'm inclined to accept this PR and move on with a _plan_ to get the functionality Bonnie needs in place.  On the other hand, if the previous (broken) functionality was still better than this fix, then we should discuss w/ Bryn.

brynrhodes:
Okay, can we use the other issue to track adding the linked nodes functionality back in and let this one go through so we can at least get processable JSON output?

edeyoung:
Would the output here dictate the specification for the JSON output?

brynrhodes:
Only to the extent that we would make sure that the annotations were JSON friendly (i.e. no mixed content spans).

alexanderelliott121:
I've had a chance to run the tests again.  I'm now seeing the output expected for the fix (the local identifiers are stripped out and the duplicate lines have been removed).

Issue #82, opened 2017-02-10
Escaped chars not translated to ELM correctly?


cmoesel:
Given the following CQL:
```
define EscapeExample: 'Hello \'World\''
```
the CQL-to-ELM translator produces:
```json
{
  "name" : "EscapeExample",
  "context" : "Patient",
  "accessLevel" : "Public",
  "expression" : {
    "valueType" : "{urn:hl7-org:elm-types:r1}String",
    "value" : "Hello \\'World\\'",
    "type" : "Literal"
  }
}
```

Is this intentional?  I would have expected the ELM to remove the escape characters since they're no longer necessary (rather than double-escaping them).  In other words, I expected `"value" : "Hello 'World'"`, not `"value" : "Hello \\'World\\'"`.

brynrhodes:
So, this was quite the rabbit-hole :) Yes, the CQL-to-ELM translator was not correctly dealing with escape sequences in strings and identifiers, so I added logic to escape/unescape CQL literals and identifiers, and a unit test to verify it. Along the way, I found an issue with the CQL grammar itself, the way strings and identifiers are defined slightly too greedily, which results in an edge case with escape characters, specifically the backslash escape (`\\`) isn't handled properly. Because that's a specification change, I can't fix that until the grammar changes in the spec, so I've submitted an STU comment to address that, and until then, I've commented out the affected tests in the unit test.

However, having fixed all that and verified that escape sequences are being handled correctly for all possible escape sequences in both strings and identifiers, the output in the XML is still `Hello \'World\'` instead of the simpler `Hello 'World'`. In the JSON, the value is the simpler, but apparently, the escape logic for XML escapes single-quotes.

So I'm committing the fix I have so far, but leaving this issue open until the full fix can be applied based on the spec change.

cmoesel:
That's a reasonable approach.  Thanks for looking into this.  If it works in JSON, I'm happy for now!

brynrhodes:
Submitted CQL STU comment #1317 to address the STRING and QUOTEDIDENTIFIER lexical rules. (A similar fix was applied to the FHIRPath specification).

brynrhodes:
Released in v1.1.1-SNAPSHOT

PR #98, opened 2017-03-22
#82: Fixed escape sequences in strings and identifiers not being hand…

brynrhodes:
…led correctly by the translator.

To address issue #82, this adds a StringEscapeUtils class and the visitor then uses it to unescape strings and identifiers in the parser.

There is still an issue with the escape sequence for a backslash, but this requires a grammar change to address, so an STU comment has been submitted against the specification and issue #82 will be addressed once that comment is handled, ideally as part of CQL 1.2.


cmoesel:
Will get to reviewing this and the others, but first I'm trying to get confirmation from the team that they're good to go with #94 (before moving on to these).  Hopefully I'll receive that confirmation today or tomorrow.

Issue #99, opened 2017-03-22
Add support for FHIR STU3


brynrhodes:
Now that FHIR STU3 is published, update the FHIR model info provider to reference the new version.

brynrhodes:
Released in v1.1.1-SNAPSHOT

Issue #89, opened 2017-03-06
FHIRHelpers source should be included by default as part of the FHIR model


nkarip:
CQL-to-ELM translation fails with error "Could not resolve library name FHIRHelpers" if library is not included in current working directory. Per Bryn, FHIRHelpers source should be included by default as part of the FHIR model so that the libraries don't have to be distributed.

brynrhodes:
Released in v1.1.1-SNAPSHOT

PR #104, opened 2017-03-28
#99: Added support for FHIR STU3.

brynrhodes:
#89: FHIRHelpers libraries are now distributed as resources with the FHIR model info.


Examples/Zika/zika-virus-intervention-logic.xml
```diff
@@ -58,483 +58,625 @@
...
             </operand>
          </expression>
       </def>
-      <def name="Zika Symptoms" context="Patient" accessLevel="Public">
-         <expression xsi:type="Query">
-            <source alias="C">
-               <expression dataType="fhir:Condition" codeProperty="code" xsi:type="Retrieve">
+      <def localId="105" name="Zika Symptoms" context="Patient" accessLevel="Public">
+         <annotation xsi:type="a:Annotation">
+            <a:s>define &quot;Zika Symptoms&quot;:&#xd;
```
cmoesel: Note that these annotations are encoding the carriage return as `&#xd;`.  That's probably not optimal (and doesn't happen on Mac) -- but probably isn't harmful either, so I'll leave it as-is for now.

PR #105, opened 2017-03-31
Add task to generate fatJar with all runtime dependencies.

ngocnguyen:
Hi Bryn,
I added a task to build fatJar with all runtime dependencies, it's useful for us.
Thanks,
Ngoc

### DIFF
--- a/Src/java/build.gradle
+++ b/Src/java/build.gradle
@@ -84,9 +84,15 @@ subprojects {
         classifier = 'sources'
         from sourceSets.main.allSource
     }
+
+    task fatJar(type: Jar) {
+        classifier = 'fat'
+        from { configurations.runtime.collect { it.isDirectory() ? it : zipTree(it) } }
+        with jar
+    }
     
     artifacts {
-        archives javadocJar, sourcesJar
+        archives javadocJar, sourcesJar, fatJar
     }
 }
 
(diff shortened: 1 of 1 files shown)


cmoesel:
I've tried this locally, and it seems to do as described (build a far jar).  The only thing I don't like is that when all the other jars get merged in, their licenses, notices, etc, also get merged in at the root level -- making it difficult to determine the real license of the project (or in some cases, misleading).  That said, this is likely an issue with all fat jars like this...  For example, there is an Eclipse license file in the root of the jar.

@ngocnguyen -- do you have any ideas about how we might be able organize the dependency libraries and notices so there is no confusion about the license of the cql-to-elm lib (Apache 2.0)?

@brynrhodes -- if there isn't a way to affect this layout in the fat jar, do you want to accept it anyway?  Or do you think it's absolutely necessary to have a fix?

cmoesel:
BTW -- I also tried it with the [ShadowJar](https://github.com/johnrengelman/shadow) plugin and it had about the same result (Eclipse license files in the root of the jar).

brynrhodes:
If there's an easy way to address the license issue, I'd like to see that resolved, I think that's pretty important to be clear about what the license is, especially as it relates to included dependencies. That being said, I don't want to hold up a useful task. Thoughts @ngocnguyen ?

One thing I would like to see is documentation on the task in the project readme.

ngocnguyen:
Sorry, I don't have any thoughts on dependency licenses.

brynrhodes:
Okay, well, I suppose let's just merge it then @cmoesel

Issue #95, opened 2017-03-21
Add links to local identifiers in ELM to the annotation narrative


brynrhodes:
Due to issues #86 and #61, links to local identifiers in the ELM were removed from the annotation narrative. This issue is a request to add them back in, pending implementer feedback that it is an important/desired feature.

edeyoung:
This is a desired feature. Thanks!

brynrhodes:
Addressed by pull request #122.

PR #122, opened 2017-06-15
#95: Fixed annotations not including spans that reference ELM local i…

brynrhodes:
…dentifiers. The fix also addresses closed issues #86 and #61.


cmoesel:
@brynrhodes -- It looks like Travis failed when attempting to compile the Java.  It can't find the `Chunk` class.  Is it possible you forgot to add a file to the commit -- or a new dependency in the build file?

brynrhodes:
Sorry, missed that file. Committed now.

alexanderelliott121:
I'm currently attempting to test this against CMS 177.
Currently seeing the following error when I try to run with annotations:



Error:[43:97, 43:111] Child chunk cannot be added because it is not contained within the parent chunk.
Error:[57:69, 57:90] elementType
Error:[59:23, 59:42] Could not validate reference to expression Initial Population because its definition contains errors.


I'm unable to attach the file I'm using for testing, I will email it to you separately.

alexanderelliott121:
I can confirm that the ordering issue has been resolved.

Issue #127, opened 2017-06-16
Sort Expressions Using Query Aliases


cmoesel:
I’m chasing down an issue in the coffeescript execution engine, but it’s led me to question some things regarding sort.

# The Problem

In the R1.1 spec, we introduce sort with this example:
```
[Encounter: “Inpatient”] E sort by start of period
```
I think this is the way sort by is intended to be used – directly referencing a property in each of the returned items.

That said, we also have some examples further on that do things like this:
```
First([Condition: "Acute Pharyngitis"] C sort by C.onsetDateTime desc)
```
Note the use of the `C` alias in the sort expression.  I’m not sure this is actually valid and it raises the question: **Can a query alias be referenced in a sort clause?**

At first glance, this seems OK – but when you really think about it, `C` is an alias used during the query execution.  The sort is defined to operate on the _results_ of the query – _after_ the execution has occurred – at which point `C` doesn’t really have meaning.

Looking at the CQL 1.1 spec, section 5.3.3 states:
> After the iterative clauses are executed for each element of the query source, the sort clause, if present, specifies a sort order for the final output.  This step simply involves sorting the output of the iterative steps…

Since the alias only holds meaning during the iterative clauses (and changes during each iteration), then if sort comes _after_ iteration, the alias has no meaning.  Section 5.3.4 makes this ordering even more clear when it lists out the steps to implementing a query: ForEach, Times, Filter, Distinct, Sort.  It seems that by the time you get to _Sort_, the alias shouldn't be in scope anymore.

Another reason it likely doesn’t make sense to reference an alias in the sort clause is that it introduces a possibility for ambiguity when the thing you’re sorting has a property with the same name as your alias.  Consider this:
```
({Tuple{E: 1}, Tuple{E: 3}, Tuple{E: 2}}) N sort by E
```
Like the first example above, this shows how sort is intended to work -- by referencing a _property_ in the returned items.  This is fine, but what happens when we do this instead:
```
({Tuple{E: 1}, Tuple{E: 3}, Tuple{E: 2}}) E sort by E
```
If we allow aliases in the sort expression, what does the `E` in the sort mean?  Is it the _property_ `E` or is it the item `E`?  It’s unclear.

# Where It's Currently Allowed

Despite the above, there are a few things that hint that using aliases in a sort clause is allowed:
* the spec has a number of examples where an alias is used in a sort clause
* the cql-to-elm translator _allows_ aliases in sort clauses (they become `AliasRef` instances)

[...]

PR #129, opened 2017-06-17
#127: Fixed aliases allowed to be referenced in sorts. Added $this it…

brynrhodes:
…erator to sorts to enable sorting by the current element.


JSRankins:
@brynrhodes and @cmoesel Agree with the fix. However, the use of the alias in a sort is used in examples in the updated CQL-based HQMF. Bryn, not sure what the possibility is of getting the examples updated there. Do we need to make a comment when the IG gets posted?

brynrhodes:
Yes, this is an issue with the CQL specification itself as well, we plan on submitting an STU errata comment once CQL 1.2 is published.

PR #137, opened 2017-06-23
Add logic for handling comparisons of time duration Quantities

mulcahyk:
When doing a compare on Quantities with a duration convert that duration to milliseconds so that different durations can be compared.

This addresses the issue in BONNIE-626.


mulcahyk:
If there are any changes made here then PR#20 in projecttacoma/cql-qdm-patientapi will need to be updated.

cmoesel:
I'll take a look at this today or Monday, depending on when I have time.  Can you please link to BONNIE-626?

mulcahyk:
https://jira.mitre.org/browse/BONNIE-626

Src/coffeescript/cql-execution/src/elm/quantity.coffee
```diff
@@ -28,16 +28,31 @@ module.exports.Quantity = class Quantity extends Expression
...
 
   before: (other) ->
-    if other instanceof Quantity and other.unit == @unit then @value < parseFloat other.value else null
+    if other instanceof Quantity and other.unit == @unit 
+      @value < parseFloat other.value 
+    else if other instanceof Quantity and other.unit of time_units and @unit of time_units
+      thisSmallestDuration = smallestDuration(@)
+      otherSmallestDuration = smallestDuration(other)
+      thisSmallestDuration < otherSmallestDuration
+    else null
```
edeyoung: it looks like this set of conversions should be done for all comparisons: `sameOrBefore`, `sameOrAfter`, `after`, `before`, `equals`. 

It looks like it should also be done for the operations, perhaps (division, subtraction, etc.). If you convert it for operations, it seems like you'll need to do a conversion back to some unit. Which unit? Maybe to the most granular unit? e.g. 5 hours - 30 minutes gives 270 minutes, 4 hours, or 4 hours 30 minutes? How will rounding be handled?

cmoesel: Agreed regarding the other comparison operators.  While we _should_ support other operations as well (addition, etc) -- I think I'd merge in this PR without them, as it does represent an incremental improvement.  But for consistency, I think we definitely need to support _all_ comparison operators -- it's confusing to support only `before`.

Src/coffeescript/cql-execution/src/elm/quantity.coffee
```diff
@@ -28,16 +28,31 @@ module.exports.Quantity = class Quantity extends Expression
...
     
   equals: (other) ->
     if other.unit? && other.value? then @unit == other.unit && @value == parseFloat other.value else null
 
-time_units = {'years':'year',  'months': 'month',  'days' :'day', 'minutes': 'minute', 'seconds':'seconds', 'milliseconds' : 'millisecond' }
+time_units = {'years': 'year', 'yr': 'year', 'y': 'year', 'months': 'month', 'weeks': 'week', 'wk': 'week', 'wks': 'week', 'days': 'day', 'd': 'day', 'minutes': 'minute', 'min': 'minute', 'seconds':'second', 'sec':'second', 'second':'second', 'milliseconds' : 'millisecond' }
```
cmoesel: I'm pretty sure `yr`, `y`, and `wks` aren't valid time units in CQL or UCUM.  Some of the new ones above are UCUM, which is probably fine -- but if you're going to add UCUM, you might as well add them all.  See the "Time Units" section here: http://download.hl7.de/documents/ucum/ucumdata.html

Src/coffeescript/cql-execution/src/elm/quantity.coffee
```diff
@@ -28,16 +28,31 @@ module.exports.Quantity = class Quantity extends Expression
...
   if time_units[units] then time_units[units] else units
 
+smallestDuration = (qty) ->
+  millivalue = switch
+    when clean_unit(qty.unit) == 'minute' then qty.value * 60000
+    when clean_unit(qty.unit) == 'hour' then qty.value * 3600000
+    when clean_unit(qty.unit) == 'day' then qty.value * 86400000
+    when clean_unit(qty.unit) == 'week' then qty.value * 604800000
+    else qty.value
+  millivalue
```
cmoesel: If the unit is in years, months, or seconds, it will fall through to the `else` and be returned as if it is `milliseconds`.  This would result in some bugs.

Src/coffeescript/cql-execution/src/elm/quantity.coffee
mulcahyk: This is done to deal with the fact that at the moment the *units* field of a Result Quantity in Bonnie is a freetext field.  Ideally we would have a dropdown (or some other method) to limit the units that users can use in Bonnie.

cmoesel: OK, but... this is a reference implementation for a standard -- so we really shouldn't be adding non-standard criteria to it.  Would it be possible to do a small translation of these on the Bonnie side before sending to the CQL execution?  e.g., if Bonnie sees a `y` unit, it translates it to `years` before sending to CQL execution?

Src/coffeescript/cql-execution/src/elm/quantity.coffee
```diff
@@ -19,25 +19,72 @@ module.exports.Quantity = class Quantity extends Expression
...
+      thisSmallestDuration = smallestDuration(@)
+      otherSmallestDuration = smallestDuration(other)
+      thisSmallestDuration < otherSmallestDuration
+    else null
     
   equals: (other) ->
     if other.unit? && other.value? then @unit == other.unit && @value == parseFloat other.value else null
 
-time_units = {'years':'year',  'months': 'month',  'days' :'day', 'minutes': 'minute', 'seconds':'seconds', 'milliseconds' : 'millisecond' }
+# Hash of time units and their UCUM equivalents
```
cmoesel: Why did you change this to return UCUM units?  Previously, it returned the CQL duration units (`year`, `month`, etc).  Since `clean_unit` uses this, and `clean_unit` is used by other functions as well (like `doAddition` and `doSubtraction`), this change may affect other things.

I tried to see if it affected any of the unit tests, but the unit tests don't run at all in this branch.  I think it's the branch you're based on `cql4bonnie` that introduced the unit test problem, however.

Src/coffeescript/cql-execution/src/elm/quantity.coffee
```diff
@@ -19,25 +19,72 @@ module.exports.Quantity = class Quantity extends Expression
...
+time_units = {'years': 'a', 'year': 'a'
+  , 'months': 'mo', 'month':'mo'
+  , 'weeks': 'wk', 'week': 'wk'
+  , 'days': 'day', 'day':'d'
```
cmoesel: `days` results in `day`, while `day` results in `d`.  I think you just missed one.  As noted above, however, I'm not sure converting to UCUM is what we want anyway.

Src/coffeescript/cql-execution/src/elm/quantity.coffee
```diff
@@ -19,25 +19,72 @@ module.exports.Quantity = class Quantity extends Expression
...
 
+smallestDuration = (qty) ->
+  if parseFloat qty.value
+    millivalue = switch
+      when clean_unit(qty.unit) == 's' then qty.value * 1000
+      when clean_unit(qty.unit) == 'min' then qty.value * 60000
+      when clean_unit(qty.unit) == 'h' then qty.value * 3600000
+      when clean_unit(qty.unit) == 'd' then qty.value * 86400000
+      when clean_unit(qty.unit) == 'wk' then qty.value * 604800000
+      when clean_unit(qty.unit) == 'mo' then qty.value * 2629746000  # Based on a mean month length of 30.436875 days
```
cmoesel: I understand why you did it this way, and I don't have a better idea, but... it does make me squirm a bit.  We should probably file a comment on the CQL spec to see if it can identify the approach that should be followed when comparing months to other units.

Src/coffeescript/cql-execution/src/elm/quantity.coffee
```diff
@@ -19,25 +19,72 @@ module.exports.Quantity = class Quantity extends Expression
...
+      when clean_unit(qty.unit) == 'a' then qty.value * 31556926000  # Based on a "common" year of 365 days
```
cmoesel: Same comment as with months.

Src/coffeescript/cql-execution/src/elm/quantity.coffee
```diff
@@ -19,25 +19,72 @@ module.exports.Quantity = class Quantity extends Expression
...
+      thisSmallestDuration = smallestDuration(@)
+      otherSmallestDuration = smallestDuration(other)
+      thisSmallestDuration <= otherSmallestDuration
+    else null
 
   sameOrAfter: (other) ->
-    if other instanceof Quantity and other.unit == @unit then @value >= parseFloat other.value else null
+    if other instanceof Quantity and other.unit == @unit
+      @value >= parseFloat other.value
+    else if other instanceof Quantity and other.unit of time_units and @unit of time_units
```
cmoesel: I'm not sure I understand what `other.unit of @unit` and `@unit of time_units` is doing here.  It's the use of the `of` keyword that is confusing me, I think.  Can you explain what's going on?

cmoesel: This same construct is seen in the other functions (`after`, `before`, etc, as well).

Src/coffeescript/cql-execution/src/elm/quantity.coffee
mulcahyk: CQL spec R 1.2 CQL section 2.5.4.1 states: 

> Note also that the UCUM time-period units can be used when expressing duration quantities.

I'll support inputs of UCUM time units but use their more common forms in the logic.

cmoesel: Perfect.

Src/coffeescript/cql-execution/src/elm/quantity.coffee
```diff
@@ -19,24 +19,78 @@ module.exports.Quantity = class Quantity extends Expression
...
+      otherSmallestDuration = smallestDuration(other)
+      thisSmallestDuration < otherSmallestDuration
+    else null
     
   equals: (other) ->
     if other.unit? && other.value? then @unit == other.unit && @value == parseFloat other.value else null
 
-time_units = {'years':'year',  'months': 'month',  'days' :'day', 'minutes': 'minute', 'seconds':'seconds', 'milliseconds' : 'millisecond' }
+# Hash of time units and their UCUM equivalents
+# See http://unitsofmeasure.org/ucum.html#para-31
```
edeyoung: I think it would be good to add a comment that both case sensitive and case insensitive UCUM units are supported.

Src/coffeescript/cql-execution/src/elm/quantity.coffee
```diff
@@ -19,24 +19,78 @@ module.exports.Quantity = class Quantity extends Expression
...
+time_units = {'years': 'year', 'year': 'year', 'a': 'year', 'ANN': 'year', 'ann': 'year'
+  , 'months': 'month', 'month':'month', 'mo': 'month', 'MO': 'month'
+  , 'weeks': 'week', 'week': 'week', 'wk', 'week'
```
edeyoung: it looks like `'WK'` got forgotten

Src/coffeescript/cql-execution/src/elm/quantity.coffee
```diff
@@ -19,24 +19,78 @@ module.exports.Quantity = class Quantity extends Expression
...
+  , 'days': 'day', 'day':'day', 'd': 'day', 'D': 'day'
+  , 'hours': 'hour', 'hour': 'hour', 'h': 'hour', 'H': 'hour'
+  , 'minutes': 'minute', 'minute': 'minute', 'min': 'minute', 'MIN': 'minute'
+  , 'seconds':'second', 'second':'second', 's': 'second'
+  , 'milliseconds' : 'millisecond', 'millisecond' : 'millisecond', 'ms': 'millisecond'
```
edeyoung: we can now safely support megaseconds and kiloseconds. Even though I doubt anyone will ever use these, should we add them back in?

Src/coffeescript/cql-execution/src/elm/quantity.coffee
```diff
@@ -19,24 +19,78 @@ module.exports.Quantity = class Quantity extends Expression
...
+smallestDuration = (qty) ->
+  if parseFloat qty.value
+    millivalue = switch
+      when clean_unit(qty.unit) == 'second' then qty.value * 1000
+      when clean_unit(qty.unit) == 'minute' then qty.value * 60 * 1000
+      when clean_unit(qty.unit) == 'hour' then qty.value * 60 * 60 * 1000
+      when clean_unit(qty.unit) == 'day' then qty.value * 24 * 60 * 60 * 1000
+      when clean_unit(qty.unit) == 'week' then qty.value * 7 * 24 * 60 * 60 * 1000
+      # Only supporting year durations based on a Julian year
+      # See http://unitsofmeasure.org/ucum.html#para-31
```
edeyoung: i think we decided to change this to gregorian based on the emails that just occurred

Src/coffeescript/cql-execution/src/elm/quantity.coffee
```diff
@@ -19,24 +19,78 @@ module.exports.Quantity = class Quantity extends Expression
...
+    if other instanceof Quantity and other.unit == @unit
+      @value < parseFloat other.value
+    else if other instanceof Quantity and time_units[other.unit]? and time_units[@unit]?
+      thisSmallestDuration = smallestDuration(@)
+      otherSmallestDuration = smallestDuration(other)
+      thisSmallestDuration < otherSmallestDuration
+    else null
     
   equals: (other) ->
     if other.unit? && other.value? then @unit == other.unit && @value == parseFloat other.value else null
```
edeyoung: Do we need to do the conversion here as well?

Src/coffeescript/cql-execution/src/elm/quantity.coffee
```diff
@@ -19,24 +19,78 @@ module.exports.Quantity = class Quantity extends Expression
     "#{@value} '#{@unit}'"
 
   sameOrBefore: (other) ->
```
edeyoung: I think there should be a comment somewhere that states that we do the time conversion for comparisons and add a TODO or a note that we do not do time conversions for operations (addition, multiplication, etc.).

Src/coffeescript/cql-execution/src/elm/quantity.coffee
mulcahyk: I think that for now we leave them out.  When we come back and work in full UCUM support we add them.  Even then I think that we might be safe leaving them out until there is an actual need for those particular units.

Src/coffeescript/cql-execution/src/elm/quantity.coffee
mulcahyk: Good catch.

Src/coffeescript/cql-execution/src/elm/quantity.coffee
```diff
@@ -19,24 +21,89 @@ module.exports.Quantity = class Quantity extends Expression
...
+    else null
+
+# Hash of time units and their UCUM equivalents, both case-sensitive and case-insensitive
+# See http://unitsofmeasure.org/ucum.html#para-31
+# The CQL specification says that dates are based on the Gregorian calendar
+time_units = {'years': 'a_g', 'year': 'a_g', 'YEARS': 'a_g', 'YEAR': 'a_g'
+  , 'a': 'a_j', 'ANN': 'a_j', 'ann': 'a_j', 'A': 'a_j'
+  , 'months': 'mo_g', 'month':'mo_g'
+  , 'mo': 'mo_j', 'MO': 'mo_j'
+  , 'weeks': 'week', 'week': 'week', 'wk', 'week', 'WK', 'week'
```
edeyoung: this looks wrong - too many commas not enough colons.

mulcahyk: Another good catch.  When it was converted to JS the colons were added correctly.

mulcahyk:
@cmoesel please review again a number of changes have been made since your last review.

Src/coffeescript/cql-execution/src/elm/quantity.coffee
```diff
@@ -59,47 +59,64 @@ module.exports.Quantity = class Quantity extends Expression
...
     if other instanceof Quantity and @unit == other.unit
       @value == parseFloat other.value
-    else if other instanceof Quantity and time_units[other.unit]? and time_units[@unit]?
+    else if other instanceof Quantity and ucum_time_units[other.unit]? and ucum_time_units[@unit]?
       thisSmallestDuration = smallestDuration(@)
       otherSmallestDuration = smallestDuration(other)
       thisSmallestDuration == otherSmallestDuration
     else null
 
+time_units = {'years':'year',  'months': 'month',  'days' :'day', 'minutes': 'minute', 'seconds':'seconds', 'milliseconds' : 'millisecond' }
```
edeyoung: can this be removed now?

Src/coffeescript/cql-execution/src/elm/quantity.coffee
```diff
@@ -19,24 +21,106 @@ module.exports.Quantity = class Quantity extends Expression
...
+
+# The smallest common duration is the millisecond
+smallestDuration = (qty) ->
+  if parseFloat qty.value
+    millivalue = switch
+      when get_ucum_unit(qty.unit) == 'second' then qty.value * 1000
+      when get_ucum_unit(qty.unit) == 'minute' then qty.value * 60 * 1000
+      when get_ucum_unit(qty.unit) == 'hour' then qty.value * 60 * 60 * 1000
+      when get_ucum_unit(qty.unit) == 'day' then qty.value * 24 * 60 * 60 * 1000
+      when get_ucum_unit(qty.unit) == 'week' then qty.value * 7 * 24 * 60 * 60 * 1000
```
cmoesel: Shouldn't the above 5 lines use the ucum units for comparison (e.g., `s`, `min`, `h`, etc)?

Src/coffeescript/cql-execution/src/elm/quantity.coffee
```diff
@@ -19,24 +21,106 @@ module.exports.Quantity = class Quantity extends Expression
...
+  }
+
+ucum_to_cql_units = {
+    'a_j':  'year'
+  , 'a_g':  'year'
+  , 'mo_j': 'month'
+  , 'mo_g': 'month'
+  , 'wk':   'week'
+  , 'd':    'day'
+  , 'h':    'hour'
```
cmoesel: You're missing `'min': 'minute'`.

Src/coffeescript/cql-execution/src/elm/quantity.coffee
```diff
@@ -19,24 +21,106 @@ module.exports.Quantity = class Quantity extends Expression
...
+  , 's':    'second'
+  , 'min':  'minute'
```
edeyoung: minor - please order in order or time, so have "minute" be ahead of "second"

Src/coffeescript/cql-execution/src/elm/quantity.coffee
```diff
@@ -19,24 +21,106 @@ module.exports.Quantity = class Quantity extends Expression
...
 
-time_units = {'years':'year',  'months': 'month',  'days' :'day', 'minutes': 'minute', 'seconds':'seconds', 'milliseconds' : 'millisecond' }
 
 clean_unit = (units) ->
-  if time_units[units] then time_units[units] else units
+  if ucum_time_units[units] then ucum_to_cql_units[ucum_time_units[units]] else units
+
+# Hash of time units and their UCUM equivalents, both case-sensitive and case-insensitive
+# See http://unitsofmeasure.org/ucum.html#para-31
+# The CQL specification says that dates are based on the Gregorian calendar
```
edeyoung: please add additional explanation to this comment. For example: 

"UCUM says that years should be Julian. As a result, CQL-based year and month identifiers will be matched to the UCUM gregorian units. UCUM-based year and month identifiers will be matched to the UCUM julian units."

Or something along those lines.

Src/coffeescript/cql-execution/src/elm/quantity.coffee
```diff
@@ -19,24 +21,106 @@ module.exports.Quantity = class Quantity extends Expression
...
+  , 'ms':   'millisecond'
+}
+
+get_ucum_units = (units) ->
+  if ucum_time_units[units]
+    ucum_time_units[units]
+  else units
+
+# The smallest common duration is the millisecond
+smallestDuration = (qty) ->
```
edeyoung: consider renaming to "durationInMilliseconds"

jbradl11: I would also just add a brief comment saying the function is converting the given value to milliseconds. I know between the name of the function and the existing comment it should be easy to figure out what is going on, but just to make it easier for future developers, I would explicitly say its returning the given quantity.value in milliseconds.

Src/coffeescript/cql-execution/src/elm/quantity.coffee
```diff
@@ -103,7 +106,7 @@ get_ucum_units = (units) ->
   else units
 
 # The smallest common duration is the millisecond
```
jbradl11: I still think the comment should explain what is actually being done in the function. But if no one else thinks its an issue, ill approve this.

edeyoung:
checked cql4browers.js in bonnie. saw expected improvements, no errors, and no score declines.

Src/coffeescript/cql-execution/src/elm/quantity.coffee
```diff
@@ -19,24 +21,110 @@ module.exports.Quantity = class Quantity extends Expression
...
+# Hash of time units and their UCUM equivalents, both case-sensitive and case-insensitive
+# See http://unitsofmeasure.org/ucum.html#para-31
+# The CQL specification says that dates are based on the Gregorian calendar
+# UCUM says that years should be Julian. As a result, CQL-based year and month identifiers will
+# be matched to the UCUM gregorian units. UCUM-based year and month identifiers will be matched 
+# to the UCUM julian units.
+ucum_time_units = {'years': 'a_g', 'year': 'a_g', 'YEARS': 'a_g', 'YEAR': 'a_g', 'a_g': 'a_g'
+  , 'a': 'a_j', 'ANN': 'a_j', 'ann': 'a_j', 'A': 'a_j', 'a_j': 'a_j'
+  , 'months': 'mo_g', 'month':'mo_g', 'mo_g': 'mo_g'
+  , 'mo': 'mo_j', 'MO': 'mo_j', 'mo_j', 'mo_j'
```
cmoesel: I think that `'MO': 'mo_j', 'mo_j', 'mo_j'` is intended to be: `'MO': 'mo_j', 'mo_j': 'mo_j'`, right?

Issue #131, opened 2017-06-20
Could not resolve call to function operator with signature


jackcmeyer:
When using the CQL file below, the following error occurs
`Could not resolve call to operator Hospitalization with Signature (QDM.EncounterPerformed).` In the previous versions of CQL, this was a valid way of making a call to a function. 

```
library INC0056733 version '0.0.006'

using QDM version '5.0.2'

parameter "Measurement Period" Interval<DateTime>

context Patient

define "Initial Population": ["Encounter, Performed"] Encounter
  where "Hospitalization"(Encounter) in "Measurement Period"

define function "Hospitalization"(Encounter "Encounter, Performed"): (
singleton from (
    ["Encounter, Performed"] EDVisit
      where EDVisit.relevantPeriod ends 1 hour or less before start of Encounter.relevantPeriod
  )
  ) X
    return if X is null then Encounter.relevantPeriod
      else Interval[start of X.relevantPeriod, end of Encounter.relevantPeriod]
```

When modifying the above example to the following, no errors occur. 
```
library INC0056733 version '0.0.006'

using QDM version '5.0.2'

parameter "Measurement Period" Interval<DateTime>

context Patient

define "Initial Population": ["Encounter, Performed"] Encounter
  where "Hospitalization"(Encounter) in "Measurement Period"

define function "Hospitalization"(Encounter "EncounterPerformed"): (
singleton from (
    ["Encounter, Performed"] EDVisit
      where EDVisit.relevantPeriod ends 1 hour or less before start of Encounter.relevantPeriod
  )
  ) X
    return if X is null then Encounter.relevantPeriod
      else Interval[start of X.relevantPeriod, end of Encounter.relevantPeriod]
```
The following ELM is produced 
```xml 
<?xml version="1.0" encoding="UTF-8"?>
<library xmlns="urn:hl7-org:elm:r1" xmlns:t="urn:hl7-org:elm-types:r1" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:fhir="http://hl7.org/fhir" xmlns:a="urn:hl7-org:cql-annotations:r1">
   <identifier id="INC0056733" version="0.0.006"/>
   <schemaIdentifier id="urn:hl7-org:elm" version="r1"/>
   <usings>
      <def localIdentifier="System" uri="urn:hl7-org:elm-types:r1"/>
[...]

PR #141, opened 2017-06-26
#131: Changed retrieve to return profile types instead of base types …

brynrhodes:
…when used with profile types.

### DIFF
--- a/Examples/CMS135_QDM.json
+++ b/Examples/CMS135_QDM.json
@@ -125,7 +125,7 @@
                "source" : [ {
                   "alias" : "E",
                   "expression" : {
-                     "dataType" : "{urn:healthit-gov:qdm:v5_0_1_draft}EncounterPerformed",
+                     "dataType" : "{urn:healthit-gov:qdm:v5_0_1_draft}PositiveEncounterPerformed",
                      "templateId" : "PositiveEncounterPerformed",
                      "codeProperty" : "code",
                      "type" : "Retrieve",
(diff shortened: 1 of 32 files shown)


brynrhodes:
Good point, I just committed updated XML and JSON ELM for all the spec examples.

cmoesel:
Thanks.  That helps a lot.  I noticed there are now some translation errors due to QUICK, but I assume that's totally unrelated.

Seeing this in real life brings up a question for me though: is there still a use for the `templateId` field in retrieves?  Or does this make it irrelevant?

brynrhodes:
I think it's still relevant because with something like FHIR, where a profile isn't necessarily defined as  a 'derived type'. It's also useful as a way to communicate the "label" of the type.

cmoesel:
Maybe I'm not understanding, but is there even a path through the translator now that would result in the templateId being populated but _different_ than the class name?  I understand that for FHIR such a thing might make sense, but didn't this change just make it so that will never happen?

brynrhodes:
Yes, the "templateId" element of the retrieve is set to the "identifier" of the type, which isn't necessarily the same as the name of the type. For FHIR-based models, we use that to identify the profile.

cmoesel:
OK.  I think I was getting confused because the term _profile_ is used in multiple contexts here.

I'll accept the change.

cmoesel:
Ah, shoot.  I just realized as I merged this that we should have run it by the Bonnie team -- or at least given them a heads up, as this might affect their patient source.  I was thinking they were already a part of the original conversation, but looking back at the issue (#131), I don't think they were.

@jbradl11 -- heads up!  Please let Bryn and me know if this is a big problem!

Issue #136, opened 2017-06-22
Enhancement Request to enable/disable output of messages to ELM


JSRankins:
@brynrhodes and @jackcmeyer There is currently no functionality to disable warnings from being put into the ELM. Reference #130. Though we believe that the inclusion of warning messages in the ELM may be useful for debugging, it may cause confusion in a production setting. Therefore, we believe it would be a good feature enhancement to have the ability to turn on/off the output of messages to ELM. A policy decision will then need to be made on whether these messages should be provided in published measure artifacts.

PR #142, opened 2017-06-27
#136: Added error-level option to the translator to control the level of me…

brynrhodes:
…ssages included in the output ELM.

### DIFF
--- a/Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/LibraryBuilder.java
+++ b/Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/LibraryBuilder.java
@@ -96,6 +96,14 @@ public class LibraryBuilder {
         listTraversal = false;
     }
 
+    private CqlTranslatorException.ErrorSeverity errorLevel = CqlTranslatorException.ErrorSeverity.Info;
+    public CqlTranslatorException.ErrorSeverity getErrorLevel() {
+        return errorLevel;
+    }
+    public void setErrorLevel(CqlTranslatorException.ErrorSeverity severity) {
+        errorLevel = severity;
+    }
+
     private Model loadModel(VersionedIdentifier modelIdentifier) {
         Model model = modelManager.resolveModel(modelIdentifier);
         loadConversionMap(model);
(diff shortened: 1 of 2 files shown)


Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/LibraryBuilder.java
```diff
@@ -326,31 +334,51 @@ else if (e.getSeverity() == CqlTranslatorException.ErrorSeverity.Info) {
...
+            case Warning:
+                return
+                        errorSeverity == CqlTranslatorException.ErrorSeverity.Warning
+                                || errorSeverity == CqlTranslatorException.ErrorSeverity.Error;
+            case Error:
+                return errorSeverity == CqlTranslatorException.ErrorSeverity.Error;
+            default:
+                throw new IllegalArgumentException(String.format("Unknown error severity %s", errorSeverity.toString()));
+        }
+    }
```
cmoesel: I know that cql-to-elm doesn't necessarily have to be performant, and hopefully errors don't happen too often, but another approach here would be to do the comparisons based on the enum's ordinal values.

In other words, I think most of the above could be replaced by:
```java
private boolean shouldReport(CqlTranslatorException.ErrorSeverity errorSeverity) {
  return errorSeverity.ordinal() >= errorLevel.ordinal();
}
```

See: https://stackoverflow.com/questions/11550118/how-to-get-enums-numeric-value

brynrhodes: I looked at using the ordinal() when I implemented this, because you're right, it could be replaced by that code, but since I didn't assign the ordinal values to the enums, that felt very much like relying on an implementation detail to describe the behavior. It's also my understanding that best practice is not to use the ordinal values, so I opted against using them. Do you agree?

cmoesel: If by implementation detail, you mean the order in which the enum is defined, then that is true.  And that is likely why some might consider it bad practice to use it.  I suppose if you did use it, it would warrant a comment in the enum definition to indicate that order DOES matter.

It wouldn't be difficult to update the enum to carry a _level_ int.  Something like:
```java
public enum ErrorSeverity {
    Info(0),
    Warning(1),
    Error(2)

    private int level;

    private ErrorSeverity(int level) {
        this.level = level;
    }

    public int getLevel() { return level; }
}
```
Then you could use `getLevel()` instead of `ordinal()` (BTW -- I didn't try to compile that, so may not be perfect).

All that said, I won't fail the pull request on this -- so if you prefer to keep it as-is, that's OK.

brynrhodes: For now I'd prefer to leave as is.

PR #156, opened 2017-07-03
Added logic to findValueSet

jbradl11:
Re added findValueSet fix. Logic will now still return value sets even if no version is passed in.

### DIFF
--- a/Src/coffeescript/cql-execution/src/example/browser/cql4browsers.js
+++ b/Src/coffeescript/cql-execution/src/example/browser/cql4browsers.js
@@ -42,23 +42,25 @@
     };
 
     CodeService.prototype.findValueSet = function(oid, version) {
-      var ref1, results;
+      var ref1, result, results;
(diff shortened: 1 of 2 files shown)


Src/coffeescript/cql-execution/src/example/browser/cql4browsers.js
```diff
@@ -3626,7 +3628,7 @@
       ref = this.element;
       for (i = 0, len = ref.length; i < len; i++) {
         el = ref[i];
-        obj[el.name] = el.execute(ctx);
+        obj[el.name] = el.exec(ctx);
```
mulcahyk: Why did this change?

jbradl11: This was already changed from a previous commit (In the coffeescript file). However the cql4browsers wasn't regenerated. The name of the commit on cql4bonnie is "Removed execute from Instance class".

PR #159, opened 2017-07-06
Replaced execute with exec in sort function. Regenerated cql4browsers.js

jbradl11:
This is a fix for internal ticker BONNIE-742: Cannot read property “Diagnosis of undefined. 

Please view that ticket for more information on how to test.

### DIFF
--- a/Src/coffeescript/cql-execution/src/example/browser/cql4browsers.js
+++ b/Src/coffeescript/cql-execution/src/example/browser/cql4browsers.js
@@ -5521,7 +5521,7 @@
             ref2 = _this.by;
             for (i = 0, len = ref2.length; i < len; i++) {
               item = ref2[i];
-              order = item.execute(a, b);
+              order = item.exec(a, b);
               if (order !== 0) {
                 break;
               }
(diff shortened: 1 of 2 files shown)


Src/coffeescript/cql-execution/src/example/browser/cql4browsers.js
```diff
@@ -3626,7 +3626,7 @@
       ref = this.element;
       for (i = 0, len = ref.length; i < len; i++) {
         el = ref[i];
-        obj[el.name] = el.execute(ctx);
+        obj[el.name] = el.exec(ctx);
```
edeyoung: please add similar comments to these changes for why execute should not be called.

jbradl11: This is on a previous PR that was merged. Want me to add it to the correct coffee script file and add the comment to this PR?

mulcahyk: When you do the browserify it seems to strip out the comments from the coffeescript.  Maybe there is switch we can use you preserve the comments?

PR #161, opened 2017-07-06
Fix 158 cql formatter

brynrhodes:
Added CQL formatter to automatically format CQL source.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,7 @@
 .DS_Store
 
 /Src/java/.nb-gradle/
+
+.idea/
+.gradle
+*.iml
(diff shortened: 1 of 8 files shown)


brynrhodes:
When parentheses are used as an expression grouper, the current CQL conventions dictate spaces that way, as opposed to when they are used in a function invocation, so that inconsistency is intentional to highlight the difference between expression grouping and function invocation.

For the other two suggestions, I agree those are issues and should be corrected, can we consider those issues against the formatter though so they don't hold up publication of this pull request?

c-schuler:
Yes. I will raise the issues against the formatter to expedite publication.

PR #171, opened 2017-07-11
Returns an object containing each library name and the list of localI…

jbradl11:
[bonnienesting01_fixed.zip](https://github.com/cqframework/clinical_quality_language/files/1139951/bonnienesting01_fixed.zip)
…ds with results

In order to test this, copy the cql4browsers.js into a local copy of cql_qdm_patientapi. Have bonnie point to that local copy. Place a debugger in the cql_calculator.js.coffee (in Bonnie repo), under the line: 
 "results = executeSimpleELM(elm, patientSource, @valueSetsForCodeService(), population.collection.parent.get('main_cql_library'), main_library_version, params)"

Go to a measure and either add a patient or edit a patient.  Open developer tools and wait for the debugger to hit. In the console type "results".

"**localIdPatientResultsMap**" should exist. 
- Click the structure open. 
- Open the first patient_id object. 
- A list of library name objects should appear, each containing a map of localIds and result values.

If you can see the library name and that object contains localids and result values, then this is working as expected.

### DIFF
--- a/Src/coffeescript/cql-execution/src/runtime/context.coffee
+++ b/Src/coffeescript/cql-execution/src/runtime/context.coffee
@@ -84,6 +84,33 @@ module.exports.Context = class Context
   set: (identifier, value) ->
     @context_values[identifier] = value
 
+  setLocalIdWithResult: (localId, value) ->
+    @localId_context[localId] = value
+
+  getLocalIdResult: (localId) ->
+    @localId_context[localId]
+
+  # Returns an object of objects containing each library name
+  # with the localIds and result values
+  getAllLocalIds: ->
+    localIdResults = {}
+    # Add the localIds and result values from the main library
+    localIdResults[@parent.source.library.identifier.id] = {}
+    for localId, value of @localId_context
+      localIdResults[@parent.source.library.identifier.id][localId] = value
+    # Iterate over support libraries and store localIds
+    for libName, lib of @library_context
+      @supportLibraryLocalIds lib, localIdResults
+    localIdResults
+
+  # Recursive function that will grab nested support library localId results
+  supportLibraryLocalIds: (lib, localIdResults) ->
+    # Set library identifier name as the key and the object of localIds with their results as the value
+    localIdResults[lib.library.source.library.identifier.id] = lib.localId_context
+    # Iterate over any support libraries in the current support library
+    for supportLibName, supportLib of lib.library_context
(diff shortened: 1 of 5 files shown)


Src/coffeescript/cql-execution/src/runtime/context.coffee
```diff
@@ -84,6 +84,33 @@ module.exports.Context = class Context
...
+    @localId_context[localId]
+
+  # Returns an object of objects containing each library name
+  # with the localIds and result values
+  getAllLocalIds: ->
+    localIdResults = {}
+    # Add the localIds and result values from the main library
+    localIdResults[@parent.source.library.identifier.id] = {}
+    for localId, value of @localId_context
+      localIdResults[@parent.source.library.identifier.id][localId] = value
```
cmoesel: Here you are iterating `@localId_context` and storing the key/value in `localIdResults[@parent.source.library.identifier.id]` one by one.

In `supportLibraryLocalIds`, you are are just assigning `lib.localId_context` to `localIdResults[lib.library.source.library.identifier.id]`.

In both cases you are essentially trying to do the same thing -- is there a reason for the difference?

Src/coffeescript/cql-execution/src/runtime/results.coffee
```diff
@@ -4,12 +4,12 @@ module.exports.Results = class Results
...
 
-  recordPatientResult: (patientId, resultName, localId_hash, result) ->
+  recordPatientResult: (patient_ctx, resultName, result) ->
+    patientId = patient_ctx.patient.id()
     @patientResults[patientId] ?= {}
-    @localIdPatientResultsMap[patientId] ?= {}
     @patientResults[patientId][resultName] = result
-    for localId, value of localId_hash
-      @localIdPatientResultsMap[patientId][localId] = value
+    @localIdPatientResultsMap[patientId] ?= {}
```
cmoesel: I don't understand the purpose of this assignment since you just re-assign `@localIdPatientResultsMap[patientId]` in the next line anyway...

jbradl11: Good catch, this was left over from my old approach.

Src/coffeescript/cql-execution/src/runtime/context.coffee
jbradl11: Good catch, I will be updating this PR soon.

PR #175, opened 2017-07-13
CQL Exec: Fix Distinct on lists of objects

cmoesel:
Prior to this fix, the Distinct operator had a bug that caused lists of objects to always be reduced to only a list of one -- whether or not they were distinct.

cql4browsers also picked up another very small recent change that apparently hadn't been browserified.

### DIFF
--- a/Src/coffeescript/cql-execution/src/elm/list.coffee
+++ b/Src/coffeescript/cql-execution/src/elm/list.coffee
@@ -100,9 +100,11 @@ module.exports.Distinct = class Distinct extends Expression
 
   exec: (ctx) ->
     arg = @execArgs ctx
-    container = {}
-    container[itm] = itm for itm in arg
-    value for key, value of container
+    seen = []
+    arg.filter (item) ->
+      isNew = seen.every (seenItem) -> !equals(item, seenItem)
+      seen.push item if isNew
+      isNew
 
 # ELM-only, not a product of CQL
 module.exports.Current = class Current extends UnimplementedExpression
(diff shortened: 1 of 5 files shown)


cmoesel:
I'm pretty sure that's according to spec.  It's also how we implement things like list `contains`.  I'll see if I can find a reference in the spec.

cmoesel:
Page 53 of the spec says:
> For structured values, equality returns true if the values being compared are the same type (meaning they have the same types of elements) and the values for each element are the same value.

If you really interpret "same type" as "same types of elements", then I think we do it right.  If we interpret it as needing to have the same constructor -- then I think we might run into problems in our implementation when people construct literal types from the data model (e.g., `Medication, Active"{ code: "foo" }`).  I don't think we need to or want to go down that path.

rdingwell:
I think the issue that I am worried about is that object literals are already being used within at least one measure.  Sometimes within a distinct union that contains objects created from a Retrieve and object literals of a specific type.

cmoesel:
OK.  Another part of the spec says:
> For tuple types, this means that equality returns true if and only if the tuples are of the same type, and the values for all elements by name are equal.

So there it's more clear that the tuple should have the same type _and_ equal attribute values (as opposed to the other part of the spec that implies type-equality is determined by the attributes being the same).
 
That said -- I suggest that the current PR is much better than the existing functionality and should be accepted -- and a new issue can be logged about construction of type literals.  I suspect that one will be tricky.

Also, I think the measure you are referencing would in fact still work if the literals and retrieve results get squished anyway (since they would only be squished if they had the same values) and they're only used in queries.

PR #176, opened 2017-07-13
Support for the ToList functionality

mulcahyk:
A change to support the functionality available in the CQL-to-ELM that handles the conversion ToList.

In reference to:  BONNIE-676


jbradl11:
Code looks good, pending CQL to ELM translation server 1.2.7 integration.

edeyoung:
demonstrating a regression

edeyoung:
regression unrelated to "ToList" other than that "ToList" may be exposing it. issue on the ncqa account Test CMS 117. Appears to be related to direct reference codes.

Issue #183, opened 2017-07-16
CQL-to-ELM Translation of "A 2 days or more after B"


cmoesel:
The spec is a little fuzzy on this, but given the phrase `A 2 days or more after B`, I would expect that it would be `true` when `A` is _exactly_ 2 days after `B`.  For example, it would be `true` when `A` is `2012-01-03T00:00:00Z` and B is `2012-01-01T00:00:00Z`.  Processing the ELM that is produced from this phrase, using these values, however, results in `false`.

Given the simpler CQL phrase, `A 2 days after B`, cql-to-elm 1.2.9 translates it as:

```xml
<expression xsi:type="SameAs">
    <operand name="A" xsi:type="ParameterRef"/>
    <operand xsi:type="Add">
        <operand name="B" xsi:type="ParameterRef"/>
        <operand value="2" unit="days" xsi:type="Quantity"/>
    </operand>
</expression>
```

This translation is equivalent to `A same as (B + 2 days)` -- so if `A` is exactly 2 days after `B`, it is `true`.

If we take that same phrase and just add `or more` to allow dates that are also more than 2 days after `B`,  the resulting CQL phrase, `A 2 days or more after B`, is translated as:

```xml
<expression xsi:type="After">
    <operand name="A" xsi:type="ParameterRef"/>
    <operand xsi:type="Add">
        <operand name="B" xsi:type="ParameterRef"/>
        <operand value="2" unit="days" xsi:type="Quantity"/>
    </operand>
</expression>
```

Note that this translation is equivalent to `A after (B + 2 days)` -- so if `A` is exactly 2 days after `B`, it is actually `false`.

This is inconsistent -- in that adding `or more` shouldn't change the semantics of the first qualifying moment.  Although the spec isn't 100% clear on this exact case, I believe that the correct approach should be to allow the exact delta (`A + 2 days` as well).

I'd suggest that the correct translation should be:

```xml
<expression xsi:type="SameOrAfter">
    <operand name="A" xsi:type="ParameterRef"/>
    <operand xsi:type="Add">
        <operand name="B" xsi:type="ParameterRef"/>
        <operand value="2" unit="days" xsi:type="Quantity"/>
    </operand>
</expression>
```

brynrhodes:
I agree, the translation is incorrect, and looking into this, it's just a logic error in the processing of timing phrases. Committing a fix now.

PR #185, opened 2017-07-16
#183: Fixed incorrect translation of timing phrases with offset durat…

brynrhodes:
…ions.


cmoesel:
@brynrhodes - do you need any more reviews before this is merged, or shall I merge it?

brynrhodes:
Yes, merge please.

PR #190, opened 2017-07-18
Fix diff calculations when straddling DST

cmoesel:
When a diff is done between two dates, when one is observing DST and the other is not, we need to properly account for the timezone offset differences.

Also updates duration just a bit to do lower level precision comparisons in a more predictible way.


losborne:
JIRA Test: https://jira.mitre.org/browse/BONNIE-866

losborne:
Regression test passed.

PR #194, opened 2017-07-20
BONNIE-882 Adds type check

alexanderelliott121:
Adds checks to ensure that when an Uncertainty attempts to calculate, it will return null if it attempts to compare mismatched type.


jbradl11:
Please add a way to test this.

alexanderelliott121:
External JIRA
https://oncprojectracking.healthit.gov/support/browse/BONNIE-317

alexanderelliott121:
Test: https://jira.mitre.org/browse/BONNIE-887

PR #200, opened 2017-07-25
CQL Exec: Add Precision to Many Interval Operators

cmoesel:
Adds support and unit tests for comparing at precision using the following Interval operators: After, Before, Contains, In, Includes, Included In, Meets, Meets After, Meets Before, Overlaps, Overlaps After, Overlaps Before, ProperIncludes, ProperIncludedIn.

Support wasn't added to the following interval operators because they're not implemented at all yet: ProperContains, ProperIn, SameOrAfter, SameOrBefore.

See:
* external BONNIE-311: https://oncprojectracking.healthit.gov/support/browse/BONNIE-311
* internal BONNIE-901: https://jira.mitre.org/browse/BONNIE-901
* PR Test: https://jira.mitre.org/browse/BONNIE-902

### DIFF
--- a/Src/coffeescript/cql-execution/src/datatypes/interval.coffee
+++ b/Src/coffeescript/cql-execution/src/datatypes/interval.coffee
@@ -7,35 +7,35 @@ cmp = require '../util/comparison'
 module.exports.Interval = class Interval
   constructor: (@low, @high, @lowClosed = true, @highClosed = true) ->
 
-  contains: (item) ->
+  contains: (item, precision) ->
     if item instanceof Interval then throw new Error("Argument to contains must be a point")
     closed = @toClosed()
     ThreeValuedLogic.and(
-      cmp.lessThanOrEquals(closed.low, item),
-      cmp.greaterThanOrEquals(closed.high, item)
+      cmp.lessThanOrEquals(closed.low, item, precision),
+      cmp.greaterThanOrEquals(closed.high, item, precision)
     )
 
-  properlyIncludes: (other) ->
+  properlyIncludes: (other, precision) ->
     if not (other instanceof Interval) then throw new Error("Argument to properlyIncludes must be an interval")
     ThreeValuedLogic.and(
-      @includes(other),
-      ThreeValuedLogic.not(cmp.equals(@, other))
+      @includes(other, precision),
+      ThreeValuedLogic.not(other.includes(@, precision))
     )
 
-  includes: (other) ->
+  includes: (other, precision) ->
     if not (other instanceof Interval) then throw new Error("Argument to includes must be an interval")
     a = @toClosed()
     b = other.toClosed()
     ThreeValuedLogic.and(
-      cmp.lessThanOrEquals(a.low, b.low),
-      cmp.greaterThanOrEquals(a.high, b.high)
+      cmp.lessThanOrEquals(a.low, b.low, precision),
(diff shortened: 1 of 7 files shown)


Src/coffeescript/cql-execution/src/datatypes/interval.coffee
```diff
@@ -7,60 +7,60 @@ cmp = require '../util/comparison'
...
+      cmp.lessThanOrEquals(closed.low, item, precision),
+      cmp.greaterThanOrEquals(closed.high, item, precision)
     )
 
-  properlyIncludes: (other) ->
+  properlyIncludes: (other, precision) ->
     if not (other instanceof Interval) then throw new Error("Argument to properlyIncludes must be an interval")
     ThreeValuedLogic.and(
-      @includes(other),
-      ThreeValuedLogic.not(cmp.equals(@, other))
```
c-monkey: I don't understand the unit tests well enough, have we characterized the changes from changing from equals to includes?

Do we know why this hasn't caused issues previously?

Src/coffeescript/cql-execution/src/datatypes/interval.coffee
```diff
@@ -155,39 +155,45 @@ module.exports.Interval = class Interval
...
-      @meetsAfter(other)
+      @meetsBefore(other, precision),
+      @meetsAfter(other, precision)
     )
 
-  meetsAfter: (other) ->
+  meetsAfter: (other, precision) ->
     try
-      cmp.equals @toClosed().low, successor(other.toClosed().high)
+      if precision? and @low instanceof DateTime
```
c-monkey: Probably a naive question. Why is precision ignored unless the interval is a DateTime?

Src/coffeescript/cql-execution/src/datatypes/interval.coffee
cmoesel: Believe it or not, it's essentially logically equivalent.  The spec describes "properly includes" as:
> The properly includes operator for intervals returns true if the first interval completely includes the second and the first interval is strictly larger than the second.

Before, we tested `a properly includes b` by checking if `a includes b` first.  If true, then to ensure that `a` is a larger interval than `b`, we checked to be sure they weren't equal.  If `a includes b` but `a != b`, then `a` must be larger than `b`.

For this next part, keep in mind that normal `includes` allows for intervals that are the same.  In other words, `Interval[1,5] includes Interval[1,5]` is `true`...

In the new logic, we still check that `a includes b`.  If it does, rather than check equality, we check to ensure that `b includes a` is `false`.  Because if `b includes a` is `true`, and `a includes b` is also true, then they are effectively equal (at the requested precision, anyway).  This is why the new code is essentially logically equivalent.  If `a includes b` but `not b includes a`, then we know `a` is larger than `b`.

The reason I didn't keep the `cmp.equals` logic is because that `equals` function doesn't support taking precision into account.

Src/coffeescript/cql-execution/src/datatypes/interval.coffee
cmoesel: The spec only defines behavior for precision when the interval is a DateTime interval.  The only precision values that the CQL grammar allows are the date-related precisions (e.g., `Year`, `Month`, `Day`, etc) -- so precision will only be populated on DateTime intervals anyway.

Issue #191, opened 2017-07-18
CQL Formatter strips comments


brynrhodes:
The CQL formatter currently strips comments from the output. These should be preserved as part of the formatted output.

brynrhodes:
This feature is included in the 1.2.11-SNAPSHOT release.

Issue #181, opened 2017-07-14
Unit testing in the CQL Formatter


brynrhodes:
Unit tests should be added to the CQL Formatter to verify:

1) Formatting functionality for all language constructs (coverage testing)
2) Formatting functionality does not impact ELM output (formatting transparency)

The unit test should include a scaffold that would, for any input CQL, verify that the unannotated ELM output from the original CQL is the same as the unannotated ELM output from the formatted CQL.

This unit test scaffold should then be invoked on all the regression testing scripts in the CQL-to-ELM translator.

brynrhodes:
This feature is included in the 1.2.11-SNAPSHOT release.

Issue #174, opened 2017-07-12
CQL being deleted when going through formatter


jackcmeyer:
When running a CQL file through the formatter, CQL code is being deleted when there are errors in the file.

For example, when running this CQL file: 
 
```
library test version '0.0.000'

using QDM

notvalueset "Female Administrative Sex": '2.16.840.1.113883.3.560.100.2'
valueset "TestValueset": '1.1.111'

context Patient

define "TestDefinition":
	["Encounter, Performed": "TestValueset "]
```

The output of the formatter is

```
library test version '0.0.000'

using QDM

valueset "TestValueset": '1.1.111'

context Patient

define "TestDefinition":
	["Encounter, Performed": "TestValueset "]
```

Another example is, when running this CQL file: 
```
library test version '0.0.000'

using QDM

valueset "TestValueset": '1.1.111'

context Patient

define "TestDefinition":
	['"Encounter, Performed": "TestValueset "]
```

The output of the formatter is: 

```
library test version '0.0.000'

using QDM

valueset "TestValueset": '1.1.111'

context Patient

define "TestDefinition":
```

Thoughts on this?

brynrhodes:
Yes, when there is a syntax error in the input, the parser will not be able to construct a full parse tree, and since the formatter works by visiting the parse tree, any part of the parse tree that couldn't be constructed will be missed in the output.

We discussed this today and will add functionality to the formatter to detect the case that there are syntax errors resulting in partial parse trees and revert to just returning the input source, unformatted.

brynrhodes:
This fix is included in the 1.2.11-SNAPSHOT release.

Issue #163, opened 2017-07-07
Cql Formatter recommendations


c-schuler:
There are a few changes to the formatter that I would recommend:
I noticed that spaces after brackets are sometimes removed, which hinders readability, but not functionality. For example:
`AgeInYearsAt(start of MeasurementPeriod) < 64` -> `AgeInYearsAt(start of MeasurementPeriod)< 64`
Another example:
`Interval[A."start datetime", A."stop datetime"] overlaps MeasurementPeriod` -> `Interval[A."start datetime", A."stop datetime"]overlaps MeasurementPeriod`

I would also suggest formatting structured values differently. 
Here is example of how they are formatted:
`Tuple { firstEl: Numerator, secondEl: 88+96*89-9.0 / 2, thirdEl: {1,2, 3} union {8,0, 9, 10} }` ->
```
Tuple { firstEl: Numerator, secondEl: 88 + 96 * 89 - 9.0 / 2, thirdEl: { 1, 2, 3 }
		union { 8, 0, 9, 10 } }
```
I would suggest putting each element on a new line:
```
Tuple { 
    firstEl: Numerator, 
    secondEl: 88 + 96 * 89 - 9.0 / 2, 
    thirdEl: { 1, 2, 3 } union { 8, 0, 9, 10 } 
}
```
I think that is cleaner.

brynrhodes:
This feature is included in the 1.2.11-SNAPSHOT release.

PR #201, opened 2017-07-27
Fix 181 formatter testing

c-schuler:
This pull request addresses the following issues with the CqlFormatter:
[#191](https://github.com/cqframework/clinical_quality_language/issues/191) - Stripping comments
[#163](https://github.com/cqframework/clinical_quality_language/issues/163) - Spacing and cleaner Tuple output
[#174](https://github.com/cqframework/clinical_quality_language/issues/174) - Deletion of code with syntax errors
[#181](https://github.com/cqframework/clinical_quality_language/issues/181) - Unit testing

### DIFF
--- a/Src/java/tools/cql-formatter/build.gradle
+++ b/Src/java/tools/cql-formatter/build.gradle
@@ -3,5 +3,11 @@ apply plugin: 'application'
 mainClassName = 'org.cqframework.cql.tools.formatter.Main'
 
 dependencies {
-    compile project(':cql')
+    compile project(":cql-to-elm")
+    compile project(":cql")
+}
+
+
+test {
+    classpath = project.sourceSets.test.runtimeClasspath + files("${rootDir}/cql-to-elm/src/test/resources")
 }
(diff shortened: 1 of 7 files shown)


Src/java/tools/cql-formatter/build.gradle
```diff
@@ -3,5 +3,11 @@ apply plugin: 'application'
 mainClassName = 'org.cqframework.cql.tools.formatter.Main'
 
 dependencies {
-    compile project(':cql')
+    compile project(":cql-to-elm")
```
brynrhodes: Can we make sure this is only a test dependency? I want to make sure the cql formatter can be used with the cql-to-elm translator.

Src/java/tools/cql-formatter/src/main/java/org/cqframework/cql/tools/formatter/Main.java
```diff
@@ -27,11 +27,13 @@ public static void main(String[] args) throws IOException {
         cqlLexer lexer = new cqlLexer(input);
```
brynrhodes: Can we refactor this, and the corresponding code block from the test cases, to a single static function that takes an InputStream and returns a String that is the formatted output?

PR #203, opened 2017-07-28
Added interval collapse implementation

c-monkey:
This fixes a bug discovered (Interval never implemented Collapse) resolving https://oncprojectracking.healthit.gov/support/browse/BONNIE-331

### DIFF
--- a/Src/coffeescript/cql-execution/src/elm/interval.coffee
+++ b/Src/coffeescript/cql-execution/src/elm/interval.coffee
@@ -133,4 +133,31 @@ module.exports.Starts = class Starts extends UnimplementedExpression
 
 module.exports.Ends = class Ends extends UnimplementedExpression
 
-module.exports.Collapse = class Collapse extends UnimplementedExpression
+module.exports.Collapse = class Collapse extends Expression
+  constructor: (json) ->
+    super
+
+  exec: (ctx) ->
+    result = @execArgs ctx
+    if result?.length > 1
+      # sort intervals by start
+      result.sort (a,b)->
+        return -1 if a.low < b.low
+        return 1 if a.low > b.low
+        0
+
+      # collapse intervals as necessary
+      intervals = []
+      a = result.shift()
+      b = result.shift()
+      while b
+        if b.low <= a.high
+          a.high = b.high if b.high > a.high
+        else
+          intervals.push a
+          a = b
+        b = result.shift()
+      intervals.push a
+      result = intervals
+
+    result
--- a/Src/coffeescript/cql-execution/src/elm/interval.coffee
+++ b/Src/coffeescript/cql-execution/src/elm/interval.coffee
@@ -147,8 +147,12 @@ module.exports.Collapse = class Collapse extends Expression
 
       # sort intervals by start
       result.sort (a,b)->
-        return -1 if a.low < b.low
-        return 1 if a.low > b.low
+        if typeof a.low.before == 'function'
+          return -1 if a.low.before b.low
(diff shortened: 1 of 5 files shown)


Src/coffeescript/cql-execution/src/elm/interval.coffee
```diff
@@ -133,4 +133,31 @@ module.exports.Starts = class Starts extends UnimplementedExpression
...
+        if b.low <= a.high
+          a.high = b.high if b.high > a.high
+        else
+          intervals.push a
+          a = b
+        b = result.shift()
+      intervals.push a
+      result = intervals
+
+    result
```
edeyoung: consider changing the name of `result` as the first variable. you call the arguments `result`, implying they will be returned. Then, at the end, you assign `intervals` to result, which is actually what's returned. This is confusing. Perhaps start with `intervals` and then add things to `collapsed_intervals` and then return `collapsed_intervals`.

c-monkey: Nice suggestion. I've used camel case which seems like the norm for this file.

Src/coffeescript/cql-execution/src/elm/interval.coffee
```diff
@@ -133,4 +133,36 @@ module.exports.Starts = class Starts extends UnimplementedExpression
...
+    if result?.length > 1
+      # we don't handle imprecise intervals at this time
+      for a in result
+        if a.low.isImprecise?() || a.high.isImprecise?()
+          throw new Error("Collapse does not support imprecise dates at this time.")
+
+      # sort intervals by start
+      result.sort (a,b)->
+        return -1 if a.low < b.low
+        return 1 if a.low > b.low
```
cmoesel: If `a.low` is an instance of `DateTime`, I don't think you can use native `<` and `>` here.  You probably have to do something like:
```
if typeof a.low.before == 'function'
  return -1 if a.low.before b.low
  return 1 if a.low.after b.low
else
  return -1 if a.low < b.low
  return 1 if a.low > b.low
0
```

Src/coffeescript/cql-execution/src/elm/interval.coffee
```diff
@@ -133,4 +133,36 @@ module.exports.Starts = class Starts extends UnimplementedExpression
...
+        0
+
+      # collapse intervals as necessary
+      collapsedIntervals = result
+      result = []
+      a = collapsedIntervals.shift()
+      b = collapsedIntervals.shift()
+      while b
+        if b.low <= a.high
```
cmoesel: Again, those native `<=` operators shouldn't work if this is a `DateTime`.  For `<=` you'd have to use `sameOrBefore`.  Not sure why your tests are passing -- but it's probably just luck (or you're passing in the intervals in the right order already).

Src/coffeescript/cql-execution/src/elm/interval.coffee
```diff
@@ -133,4 +133,36 @@ module.exports.Starts = class Starts extends UnimplementedExpression
...
+          a.high = b.high if b.high > a.high
```
cmoesel: For `DateTime`, use `after` instead of `>`.

cmoesel:
Code looks good, unit tests pass.  Did not test in Bonnie.

Will probably need to be rebased since another PR was just merged.

c-monkey:
This is in Bonnie, we should merge the PR now. @cmoesel approves merging this into cql4bonnie branch.

cmoesel:
I'll do it.

PR #205, opened 2017-08-01
Ucum4bonnie

rdingwell:
Unit conversion, comparison and calculation for the CQL masses


Src/coffeescript/cql-execution/test/elm/comparison/test.coffee
```diff
@@ -75,6 +75,25 @@ describe 'Less', ->
   it 'should be true for 5 < 6', ->
     @aLtB_Int.exec(@ctx).should.be.true()
 
+  it 'should be true for 5 m < 4 m', ->
```
cmoesel: I think you mean: `it 'should be false for 5 m < 4 m', ->`

cmoesel: And it looks like similar mistakes are repeated below.  The tests are actually right, but the string descriptions don't always match what the test is asserting.

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/CqlTranslator.java
```diff
@@ -230,7 +230,7 @@ public static String convertToJSON(Library library) throws JAXBException {
         return writer.getBuffer().toString();
     }
 
-    private static void loadModelInfo(File modelInfoXML) {
+    public static void loadModelInfo(File modelInfoXML) {
```
cmoesel: Why?

rdingwell: I think this is a leftover from a couple of years ago from another project.  I was creating javascript models on the fly in a UI and needed access to the method for things to work correctly, at least thats as best as I can remember.  I can take this out if needed.

Src/coffeescript/cql-execution/test/elm/comparison/test.coffee
rdingwell: Whooraay for cut and paste, will update.

Src/coffeescript/cql-execution/test/elm/arithmetic/test.coffee
```diff
@@ -1,6 +1,39 @@
...
+  q = Q.createQuantity(expectedValue,expectedUnit)
+  q.equals(object).should.be.true("Expected "+ object + " to equal " + q)
+  # if object.value
+  #   object.value.should.equal expectedValue
+  # else
+  #   (object.value == expectedValue).should.equal true
+  # if object.unit
+  #   object.unit.should.equal expectedUnit
+  # else
+  #   (object.unit == expectedUnit).should.equal true, "Expected "+ (object.unit) + " to equal " + (expectedUnit)
```
rdingwell: nope will remove

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/CqlTranslator.java
cmoesel: Eh.  It's probably fine as is.  I just didn't understand how it fit in with the rest of this PR.

Src/coffeescript/cql-execution/src/elm/aggregate.coffee
```diff
@@ -1,24 +1,32 @@
...
+  if allQs
     unit = arr[0].unit
     values = []
     for i in arr
-      if i.constructor.name == "Quantity" && i.unit == unit
-        values.push i.value
-      else
-        return []
+      values.push i.convertUnits(unit)
     return compact(values) # need to make sure that there are not any null values from the quntities
```
rdingwell: That is what I was assuming, just because the initial array has been compacted I wasn't sure that there would be values wouldn't contain a null.  Though, this is also from 2 years ago before I started throwing exceptions, so I'm not sure the assumption is or isnt valid.  I can take it out if needed.

Src/coffeescript/cql-execution/src/elm/instance.coffee
```diff
@@ -15,7 +15,9 @@ module.exports.Instance = class Instance extends Expression
     super
     @classType = json.classType
     @element = ( new Element(child) for child in json.element)
-
+ 
+  # TODO:  Implement switch statement for other system types such as Code and Concept
+  #
```
rdingwell: will remove, zombied from a couple of years ago I imagine

Src/coffeescript/cql-execution/src/elm/aggregate.coffee
cmoesel: Yeah, now that I've looked through the code more thoroughly, I don't think `i.convertUnits` can return `null` -- so if you're editing the code anyway, I say remove the extra `compact`.  Especially since this function gets called a lot.

Src/coffeescript/cql-execution/src/elm/quantity.coffee
```diff
@@ -21,49 +28,64 @@ module.exports.Quantity = class Quantity extends Expression
...
+    @multiplyDivied(other,"/")
+
+  multiplyBy: (other) ->
+    @multiplyDivied(other,".") # in ucum . represents multiplication
+
+  multiplyDivied: (other, operator) ->
+    if other instanceof Quantity
+      if @unit and other.unit
+        can_val = @to_ucum()
+        other_can_value = other.to_ucum()
```
rdingwell: I had issues when attempting to use canonicalize and fell back to this approach.  I didn't deg to deeply into what that method was doing just new that it was not working as I thought it would.

Src/coffeescript/cql-execution/src/elm/quantity.coffee
```diff
@@ -21,49 +28,64 @@ module.exports.Quantity = class Quantity extends Expression
...
+        ucum_value = ucum_multiply(can_val,[[operator,other_can_value]])
+        createQuantity(ucum_value.value, units_to_string(ucum_value.units))
+      else
+        value = if operator == "/" then @value / other.value  else @value * other.value
+        unit = @unit || other.unit
```
rdingwell: is there a test like that that is working?  I thought that the parser would throw and exception in cases like those.  I can certainly guard against it.

Src/coffeescript/cql-execution/src/elm/quantity.coffee
```diff
@@ -99,34 +121,62 @@ ucum_to_cql_units = {
...
-      # Support for the UCUM units based on the Julian calendar
-      when get_ucum_unit(qty.unit) == 'mo_j' then qty.value * 30.4375 * 24 * 60 * 60 * 1000  # Based on a Julian mean month length of 30.4375 days
-      when get_ucum_unit(qty.unit) == 'a_j' then qty.value * 365.25 * 24 * 60 * 60 * 1000  # Based on a Julian year of 365.25 days
-      else qty.value
-    millivalue
-  else
-    null
-
-module.exports.createQuantity = (value,unit) ->
+# this is used to perform any convertions of CQL date time fileds to their ucum equivalents
```
rdingwell: boo, typos are my legayc

Src/coffeescript/cql-execution/src/elm/quantity.coffee
```diff
@@ -140,30 +190,27 @@ module.exports.parseQuantity = (str) ->
...
 
 module.exports.doMultiplication = (a,b) ->
   if a instanceof Quantity and b instanceof Quantity
-    # TODO: proper conversion of units (e.g., 5 m * 5 m = 5 m^2)
-    null
+    a.multiplyBy(b)
   else
     [q,d]  = if a instanceof Quantity then [a,b] else [b,a]
-    new Quantity({unit: q.unit, value: q.value * d})
+    createQuantity(q.value * d, q.unit)
```
rdingwell: I think that should work

Src/coffeescript/cql-execution/src/elm/quantity.coffee
cmoesel: If you're sure that it will always result in the same base units, regardless of the original units passed in, then it's fine.  My concern is if it parses to different units for similar unit dimensions (e.g., if `kg` parsed down to `g` and `[lb_av]` parsed down to `[oz_av]`, then comparing just the values wouldn't be right).

Perhaps a test or two that compared metric vs non-metric would help confirm one way or the other.  Alternately, we could look at the source code to confirm -- or in the code above do a check to verify the units are the same between `can_val` and `other_can_value`.

Src/coffeescript/cql-execution/src/elm/quantity.coffee
cmoesel: Ah, yeah -- you're right.  The spec doesn't allow it.  Could you put in a comment noting that decimal / quantity should never happen, according to the spec?

Src/coffeescript/cql-execution/src/elm/quantity.coffee
```diff
@@ -99,34 +121,62 @@ ucum_to_cql_units = {
...
+# this is used to perform any convertions of CQL date time fileds to their ucum equivalents
+ucum_unit = (unit) ->
+  ucum_time_units[unit] || unit || ''
+
+#just a wrapper function to deal with possible exceptions being thrown
+convert_value = (value, from ,to ) ->
+  try
+    decimalAdjust("round", ucum.convert(value,ucum_unit(from),ucum_unit(to)), -8)
+  catch e
+    throw new IncompatibleTypesException(from, to, e)
```
rdingwell: I think they should and I'll fix it.  The ucum library will throw the exception but it's just a string so I felt the need to wrap it a more object oriented fashion.  It also provides a central location for rounding the values.

Src/coffeescript/cql-execution/src/elm/quantity.coffee
rdingwell: Returning different base units shouldn't be a problem because that is handled by the ucum_multiply method.   It's how we can end up with  10 'm' / 1 's' = 10 'm/s' .   The purpose of the parsing the units is to take something like kg and have it represented as g.   It's also how we end up with the example that bonnie provided for  < 10 'mg/[lb_av]'

Src/coffeescript/cql-execution/src/elm/quantity.coffee
rdingwell: Actually looking at the section you highlighted it is dealing with 2 Quantity objects where one or both may not have a unit associated with them.  This is going back a couple of years so I'm not sure that is a valid assumption.  Should this just treat these as regular quantities and throw an exception if one has a unit and the other does not?

Src/coffeescript/cql-execution/src/elm/quantity.coffee
```diff
@@ -1,7 +1,14 @@
...
+{ FunctionRef } = require './reusable'
+{ decimalAdjust } = require '../util/math'
 { ValueSet, Code } = require '../datatypes/datatypes'
+{ Exception } = require '../datatypes/exception'
 { build } = require './builder'
+ucum = require  'ucum'
 
+module.exports.IncompatibleTypesException = IncompatibleTypesException = class IncompatibleTypesException extends Exception
+  constructor: (@a , @b , e) ->
+    super("Incompatible Types " + @a + " " + @b, e)
```
edeyoung: This error appears to be thrown when a unit is not a UCUM unit. See screen shot.

image

This results in a confusing error message to the user: the types are compatible (they're both `wks`), they're just not correct.

Can there be a "bad ucum unit" exception that gets thrown instead in these situations?

edeyoung: I also suggest changing the error message to:

```
super("Incompatible types '" + @a + "' and  '" + @b + "'", e)
```

So it will look like: "Incompatible types 'wks' and 'wks'"

rdingwell: I can make this change

edeyoung:
The measure developers may want to push this until 9/12 release. They will get back to us by 8/14.

edeyoung:
Bryn brought up the issue that by implementing this in Bonnie, we are implying that vendors need to implement this. This is currently not a requirement for vendors and adds substantial complexity. This will be discussed on 8/22.

rdingwell:
I do not agree with the statement that implementing this in Bonnie is
implying that vendors need to implement it as well.  Measure developers
developing measures that use Quantities with units is where the implication
lies.  If MD's are going to use those features of the specification then
Bonnie needs to be able to deal with them.  If Bonnie can handle those
types of expressions and the MD's do not use them then there is no
implication.  I do not understand the logic that says that the underlying
calculation engine being feature complete to the specification is a problem.

brynrhodes:
@rdingwell , you're absolutely right to clarify that the requirement would be imposed by the content of the program measures, not the supported feature set of the engine. The reason I raise it in this context though is that the program measure developers were asked to prioritize the feature, implying they would use it on program measures.

Src/coffeescript/cql-execution/src/elm/quantity.coffee
```diff
@@ -1,7 +1,14 @@
...
+{ FunctionRef } = require './reusable'
+{ decimalAdjust } = require '../util/math'
 { ValueSet, Code } = require '../datatypes/datatypes'
+{ Exception } = require '../datatypes/exception'
 { build } = require './builder'
+ucum = require  'ucum'
 
+module.exports.IncompatibleTypesException = IncompatibleTypesException = class IncompatibleTypesException extends Exception
+  constructor: (@a , @b , e) ->
+    super("Incompatible Types '" + @a + "' and '" + @b +"'", e)
```
edeyoung: From Luke:

It looks like `@b + “’”, e` in CoffeeScript somehow gets turned into` this.b(+"'", e)` in JavaScript (very strange), which causes the error “this.b is not a function” to be thrown.
 
Changing the line to this seems to resolve the issue:
```super("Incompatible Types ‘#{@a}’ and ‘#{@b}’”, e)```

edeyoung:
After discussion today, it sounds like we should proceed with this integration.

rdingwell:
Change made and pushed

Issue #212, opened 2017-08-14
Update QDM 5.3 model info based on QDM 5.3 Annotated


brynrhodes:
Clarifications and errata have been added to QDM 5.3, the model info file for QDM 5.3 needs to be updated to reflect these changes.

brynrhodes:
Addressed in v1.2.12-SNAPSHOT.

PR #213, opened 2017-08-14
#212: Changes based on clarification and errata for QDM 5.3 Annotated…

brynrhodes:
…: Added relevantPeriod to MedicationOrder and MedicationDispensed; Added relatedTo to AssessmentPerformed and the Communication types; Changed diagnosis to diagnoses and facilityLocation to facilityLocations for EncounterPerformed; Changed component to components throughout.

### DIFF
--- a/Src/java/qdm/schema/qdm.5.3.xsd
+++ b/Src/java/qdm/schema/qdm.5.3.xsd
@@ -272,6 +272,7 @@
 					<xs:element name="method" type="dt:Code" minOccurs="0" maxOccurs="1"/>
 					<xs:element name="result" type="dt:Any" minOccurs="0" maxOccurs="1"/>
 					<xs:element name="component" type="Component" minOccurs="0" maxOccurs="unbounded"/>
+					<xs:element name="relatedTo" type="dt:Code" minOccurs="0" maxOccurs="unbounded"/>
 				</xs:sequence>
 			</xs:extension>
 		</xs:complexContent>
@@ -332,6 +333,7 @@
 				<xs:sequence>
 					<xs:element name="authorDatetime" type="dt:DateTime" minOccurs="0" maxOccurs="1"/>
 					<xs:element name="negationRationale" type="dt:Code" minOccurs="0" maxOccurs="1"/>
+					<xs:element name="relatedTo" type="dt:Code" minOccurs="0" maxOccurs="unbounded"/>
 				</xs:sequence>
 			</xs:extension>
 		</xs:complexContent>
@@ -350,6 +352,7 @@
 				<xs:sequence>
 					<xs:element name="authorDatetime" type="dt:DateTime" minOccurs="0" maxOccurs="1"/>
 					<xs:element name="negationRationale" type="dt:Code" minOccurs="0" maxOccurs="1"/>
+					<xs:element name="relatedTo" type="dt:Code" minOccurs="0" maxOccurs="unbounded"/>
 				</xs:sequence>
 			</xs:extension>
 		</xs:complexContent>
@@ -368,6 +371,7 @@
 				<xs:sequence>
 					<xs:element name="authorDatetime" type="dt:DateTime" minOccurs="0" maxOccurs="1"/>
 					<xs:element name="negationRationale" type="dt:Code" minOccurs="0" maxOccurs="1"/>
+					<xs:element name="relatedTo" type="dt:Code" minOccurs="0" maxOccurs="unbounded"/>
 				</xs:sequence>
 			</xs:extension>
 		</xs:complexContent>
@@ -947,6 +951,7 @@
 			<xs:extension base="QDMBaseType">
 				<xs:sequence>
 					<xs:element name="authorDatetime" type="dt:DateTime" minOccurs="0" maxOccurs="1"/>
+					<xs:element name="relevantPeriod" type="dt:DateTimeInterval" minOccurs="0" maxOccurs="1"/>
 					<xs:element name="refills" type="xs:integer" minOccurs="0"/>
 					<xs:element name="dosage" type="dt:Quantity" minOccurs="0"/>
 					<xs:element name="supply" type="dt:Quantity" minOccurs="0"/>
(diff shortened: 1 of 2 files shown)


Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-5.3.xml
```diff
@@ -85,8 +85,8 @@
         <ns4:element name="admissionSource" type="System.Code"/>
         <ns4:element name="relevantPeriod" type="interval&lt;System.DateTime&gt;"/>
         <ns4:element name="dischargeDisposition" type="System.Code"/>
-        <ns4:element name="facilityLocation" type="QDM.FacilityLocation"/>
-        <ns4:element name="diagnosis" type="list&lt;System.Code&gt;"/>
+        <ns4:element name="facilityLocations" type="QDM.FacilityLocation"/>
```
edeyoung: The type should be changed to a list type: `list&lt;QDM.FacilityLocation&gt;`

Src/java/qdm/schema/qdm.5.3.xsd
```diff
@@ -272,6 +272,7 @@
 					<xs:element name="method" type="dt:Code" minOccurs="0" maxOccurs="1"/>
 					<xs:element name="result" type="dt:Any" minOccurs="0" maxOccurs="1"/>
 					<xs:element name="component" type="Component" minOccurs="0" maxOccurs="unbounded"/>
+					<xs:element name="relatedTo" type="dt:Code" minOccurs="0" maxOccurs="unbounded"/>
```
edeyoung: A question that we've been having with `relatedTo`. This has a type `dt:Code`, but from all of the email exchanges, it sounds like it should be an id. How would this be represented in this file?

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-5.3.xml
```diff
@@ -85,8 +85,8 @@
...
+        <ns4:element name="diagnoses" type="list&lt;System.Code&gt;"/>
```
edeyoung: do the pluralized names need to be reflected in the schema file?

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-5.3.xml
brynrhodes: Agreed and applied.

Src/java/qdm/schema/qdm.5.3.xsd
brynrhodes: We don't really have an "ID" type in the base CQL system types, so I kind of overloaded Code for that purpose. It's probably not entirely appropriate, but in the absence of a suitable ID type (specifically one that had a System), I think it's better to be able to use a Code, with the understanding that the only thing required is the "code" element that would contain the identifier as a string. The "system" element can then be used to provide a namespace if necessary.

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-5.3.xml
brynrhodes: I intentionally left that change out of the schema because from an XML perspective, it's still an "element", it's just that the element may appear more than once, so that's more consistent with conventions in an XSD file.

Src/java/qdm/schema/qdm.5.3.xsd
edeyoung: double checking with the development team that this makes sense for us.

Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-5.3.xml
edeyoung: Makes sense, thanks!

Src/java/qdm/schema/qdm.5.3.xsd
edeyoung: This should work, thanks!

edeyoung: Should a comment be added somewhere to reflect the overloaded function of "Code" or does this exist already?

brynrhodes: There should probably be documentation on how QDM is realized in CQL, yes. We don't have anywhere that talks about this specifically right now, but the Formatting and Usage Guide is probably the appropriate place for that. I'll bring it up with the QDM stewards and see what we can do about that type of documentation.

edeyoung: ok, thanks!

p9g: xsd file has xmlns="urn:healthit-gov:qdm:v5_0_2_draft" and targetNamespace="urn:healthit-gov:qdm:v5_0_2_draft"

Should these be updated to v5_3?

brynrhodes: Yep, I verified the one in the model info but missed it in the xsd. Since this has pull request has already been merged, I'll consider this a typo and make the change directly.

p9g: I think you forgot to fix the typo in the targetNamespace also

brynrhodes: Yep, fixed.

cmoesel:
I realize I am late to the discuss (darned vacations), but an approach I would have considered is creating a QDM `ID` that extends `System.Code`, and using `QDM.ID` as the type for things like `relatedTo`.  It ends up being pretty much the same thing, but I think it makes the distinction more clear when trying to determine if something is a real code value or an id value.
```xml
<ns4:typeInfo xsi:type="ns4:ClassInfo" name="QDM.ID" retrievable="false" baseType="System.Code" />
```

If it's too late (or if it's a bad idea), then fine -- but I thought I should just throw it out there.

JSRankins:
@brynrhodes and @cmoesel: if a change is going to be made, the MAT team can accommodate it if the solution is provided within the next couple of days. We would need something by COB Friday, since we are getting ready to go into our last sprint of work prior to the MAT 5.4 release. Or it can stay as is. Thoughts?

PR #219, opened 2017-08-23
Cql4bonnie param fix

jbradl11:
Chris added code to ParameterRef in support of libraries. Added fix for parameters in children contexts. Reverted findValueSet code back to original state. All but one CalculateAge unit test pass now. CalculateAge unit test passes sometimes (Moesel can explain in more detail why that happens).


losborne:
Regression Test run, only change is expected as user has to re-upload measure.  All others are equal.
```
hmdmtjc
-: CMS ID: CMS113v6 FIRST BRANCH HAS MORE PASSES (57 > 50)
```

Issue #221, opened 2017-08-25
Create an ID type in QDM for the ID attributes


brynrhodes:
Create an ID type with NamingSystem and Value elements, and use it as the type for all ID attributes in the QDM model info.

This enhancement is in response to a comment on PR #213

brynrhodes:
Addressed in v.1.2.13-SNAPSHOT

PR #222, opened 2017-08-25
#221: Added an Id type to QDM to support more accurate modeling of Ids.

brynrhodes:


brynrhodes:
Fixed, thanks!

brynrhodes:
@cmoesel diagnoses and principalDiagnosis are intended to be codes, not references to a Diagnosis instance.

For patientId, reporter, and recorder, good point; agreed, fixed, and committed.

JSRankins:
@brynrhodes and @cmoesel: Bryn, I thought that reporter and recorder attributes were not instances but would be code based since they map over to some of the data flow attributes in QDM.

brynrhodes:
@JSRankins that's a good point, but my read of the QDM specification is that those are intended to point to instances, the language used is "individual or device". Have they been considered type-level in the past?

p9g:
What value should be used for namingSystem?

brynrhodes:
@p9g namingSystem would be optional, but if provided, it should be a globally understood namespace such as a uri or an oid.

JSRankins:
@brynrhodes: Good question. Since they have never been used in measures and they don't really show up in the current CQL-based HQMF IG, the only thing I think we can rely on is how they are defined in QDM and how they were templated in the past. 

In QDM, _source_ (I assume this is what is meant as reporter) is defined as "the originator of the quality data element. The source may be an individual or a device." _Recorder_ is defined as "the individual or device that enters the data element into a health record field. The desired recorder also may be, but is not necessarily, the source of the data."

In the past, their associated templates took actual codes. I'm not sure they were modeled correctly. On the reporting side, they would have been handled as participants that would have required codes and  information related to an actual person, organization or device that participated as the reporter or recorder.

Since these are not used in measures, I'm not sure it is worth holding up the release. If we find that they are needed at some point, we may have to revisit this. Thoughts?

JSRankins:
@p9g and @brynrhodes: Does this change how id would be used with related to?

brynrhodes:
@JSRankins agreed, we may need to revisit, but since the risk is minimal given the attributes are not used in any current program measures, we should go ahead and commit. Barring any objections from the MITRE team, we should merge the PR.

brynrhodes:
@p9g no, the logic for fulfillment would be the same, for example:

    define "Fulfills With Assessment":
      "Assessment" A
        where exists ("Referral" R where R.id in A.relatedTo)

cmoesel:
As noted above, you've got my approval.  @edeyoung, how about you?  Your team will have to implement this.

edeyoung:
Looks good to me.

cmoesel:
Ha.  @edeyoung beat me to the punch with that approval.

brynrhodes:
@cmoesel can you merge, since I submitted the PR?

PR #224, opened 2017-08-29
cql4bonnie babelify ucum.js

hossenlopp:
- Adds cake task for building cql4browsers.js.
- Babelifies ucum.js dependency to be ES5 compliant.
- Adds necessary dependencies to package.json.


losborne:
Regression test shows expected errors only related to invalid ucum units.  All other results are equal.

Expected Errors were with:
CMS744 (bpm), CMS65 (mmHg), CMS113 (wks), and CMS3 (wks)

Issue #225, opened 2017-08-30
component in QDM 5.3 model info file missing reference high/low for laboratory test performed


edeyoung:
The QDM documentation specifies that reference high and low should be optional attributes of a "component" when used for a laboratory test performed.

brynrhodes:
Addressed in v.1.2.14-SNAPSHOT

PR #226, opened 2017-08-30
#225: Added referenceRange as an attribute of the component of a Labo…

brynrhodes:
…ratory Test, Performed.

The change was implemented by defining a new ResultComponent type that derives from Component and adds the referenceRange element. The LaboratoryTestPerformed.components element was then changed to use the new ResultComponent type. Hopefully, using inheritance minimizes the impact on implementers, but seeking feedback on the approach.


Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-5.3.xml
```diff
@@ -177,6 +177,9 @@
             </ns4:typeSpecifier>
         </ns4:element>
     </ns4:typeInfo>
+    <ns4:typeInfo xsi:type="ns4:ClassInfo" name="QDM.ResultComponent" retrievable="false" baseType="QDM.Component">
```
edeyoung: Looks good, although I wonder if `ResultComponent` is the best name. Aren't most uses of component like a result even out side of lab tests? I do not have a good suggestion for another name. If you're comfortable with this name, then this PR looks good to me.

PR #228, opened 2017-08-31
UCUM Unit Performance Enhancements

hossenlopp:
- Added caching of ucum unit validity.
- Checks units on Quantity creation.
- Skip ucum.convert if comparing quantities and they are the same unit.
- Unit tests for Quantity creation.


losborne:
Manual Regression test passes.

Differences (investigated and are OK):
```
epecqmama:
* CMS736v0 Pending
epecqmncqa:
* CMS134v5 Pending
* CMS65v6 Pending
* CMSv0 "Test165" Pending
```

Issue #234, opened 2017-09-07
Issue with comparing Quantities 


rdingwell:
The following statement is taken from a measure that is currently being developed and highlights what I believe is an issue with the way the CQL spec currently handles comparisons of Quantities with incompatible units.   
````
define "Pap Test with Results":
    ["Laboratory Test, Performed": "Pap Test"] LabTest
         where ( LabTest.result as Code in "Positive Result" )
	 or ( ( LabTest.result as Quantity ) < 100 '%' )
	 or ( ( LabTest.result as Quantity ) > 45 'mmol/L' )
	 or ( LabTest.result >= 30 )
````
According to the spec in a number of locations. 

> For comparisons involving quantities, the dimensions of each quantity must be the same, but not necessarily the unit. For example, units of 'cm' and 'm' are comparable, but units of 'cm2' and 'cm' are not. Attempting to operate on quantities with invalid units will result in a run-time error.

In the case above the LabTest result can be coded either as a percentage or as millimoles per liter.  This presents a problem, if the result is a Quantity coded as mmol/L the initial comparison to a Quantity as a percentage is expected to throw a runtime error.   

For instances where an expected Quantity can be coded in multiple, incompatible units there  does not appear to be a way to handle the situation without throwing a runtime error at some point.  I can see throwing exceptions for arithmetic operations on Quantities that are incompatible but think that something else needs to be considered for comparison operations.   Perhaps this could be handled by some sort of Uncertainty or possibly a qualifier of some sort that would allow for the comparison to be performed only if the units where compatible with each other.

cmoesel:
Another approach might be to leverage the three-valued logic that CQL supports and return `null` when comparisons are made using incompatible units.  Since `null` is basically equivalent to _maybe_ or _I don't know_, then this approach would essentially say _"Is 50 'mmol/L' < 100 '%'?  I don't know!"_

brynrhodes:
I agree that an error isn't the most friendly. I'm a little hesitant to return null, although I think it's probably the best answer. It's worth a comment to the STU, so I'll submit one related to this issue.

As a potential workaround though, you could test the unit:

    define "Pap Test with Results 2":
      ["Laboratory Test, Performed": "Pap Test"] LabTest
        where
          case
            when LabTest.result as Code in "Positive Result" then true
            when LabTest.result.unit = '%' then LabTest.result < 100 '%'
            when LabTest.result.unit = 'mmol/L' then LabTest.result > 45 'mmol/L'
[...]

PR #235, opened 2017-09-07
Return null for comparisons between incompatible units (BONNIE-1046)

hossenlopp:
For comparisons that involve units if they are incompatible units, the comparison will return false instead of throwing an error.

JIRA Test: https://jira.mitre.org/browse/BONNIE-1058


cmoesel:
Sorry for coming late to the party, and I don't want to throw a wrench in things, but... it may be more appropriate to return `null` in these cases rather than `false`.  See my comment on #234.

brynrhodes:
I agree with @cmoesel , null is more appropriate than false.

hossenlopp:
I agree, `null` makes more sense. Change has been made and relevant bonnie branches have been updated for testing.

Src/coffeescript/cql-execution/test/elm/comparison/test.coffee
```diff
@@ -93,6 +93,14 @@ describe 'Less', ->
   it 'should be false for 5 m < 5 km', ->
     @aLtB_Quantity_diff.exec(@ctx).should.be.true()
 
+  it 'should be false for 5 Cel < 4 m', ->
```
cmoesel: The description of this test, and the others, needs to be updated to say "should be **null**..." rather than "should be **false**..."

Src/coffeescript/cql-execution/src/elm/comparison.coffee
```diff
@@ -1,5 +1,6 @@
 { Expression } = require './expression'
 { Uncertainty } = require '../datatypes/datatypes'
+{ IncompatibleTypesException } = require './quantity'
 
 # Equal is completely handled by overloaded#Equal
```
cmoesel: Should we also have similar logic in `Equal` (e.g., should `85 'mmol/L' = 65 '%'` return null)?

Src/coffeescript/cql-execution/test/elm/comparison/test.coffee
hossenlopp: Whoops. I totally screwed that up. I'll fix that.

Src/coffeescript/cql-execution/src/elm/comparison.coffee
hossenlopp: Good catch. I'll go implement that too.

hossenlopp:
Requested updates have been made to include checks with Equal and NotEqual and relevant bonnie branches have been updated.

losborne:
JIRA Test Passes.

Issue #237, opened 2017-09-12
Upgrade gradle build to 4.1


brynrhodes:
Current stable gradle version is 4.1. The build should be updated to use this latest version.

PR #238, opened 2017-09-12
#237: Upgraded gradle build to 4.1.

brynrhodes:


c-schuler:
I was able to successfully build the project with the upgraded Gradle.
Tried the following without issue:
gradle build
gradle idea
./gradlew build
./gradlew idea

PR #241, opened 2017-09-28
Performance Enhancement - Reuising the JAXBContext object

c-schuler:


c-schuler:
I left those static methods as a convenience (wasn't sure if anyone was using them). I can remove them if they serve no purpose.

c-schuler:
Okay, those have been removed.

Issue #240, opened 2017-09-18
Breaking changes in FHIR 1.0.2 modelinfo


cmoesel:
Previously, CQL could reference an Observation's value as a quantity via the field `valueQuantity` (e.g., `Observation O where O.valueQuantity < 40 'mg/dL'`).

Since that time, the `fhir-modelinfo-1.0.2.xml` definition has changed to support the new choice options, and the previous way of referencing a value quantity is no longer valid.  Now it seems that you just reference `value` and the CQL-to-ELM will cast appropriately if needed.

While this is easier for authoring (in some ways), it does reflect a departure from the way that FHIR is actually serialized on the wire -- you would never see a `value` field in FHIR data -- it would always be `valueQuantity` or `valueCodeableConcept`, etc.  From an implementation point of view, I found the more specific field names easier to deal with.  (Furthermore, I'm not sure you even see `value` in the FHIR spec, isn't it always `value[x]`?).  Is this representation up for discussion or is set in stone?

This also brings up an important point regarding versioning.  Our CDS artifacts on CDS Connect were compiled against the _old_ version of the `fhir-modelinfo-1.0.2.xml` file -- and thus use `valueQuantity`, `effectiveDateTime`, etc.  They will not compile against the _new_ version of that file.  The problem is, the modelinfo version number is tied to the FHIR version number (1.0.2), so even though significant _breaking_ changes were made to the modelinfo file, the version number was not incremented -- and as a result there is no way for the CDS Connect artifacts to indicate they used the _previous_ version of that modelinfo.  Worse yet, there are now two very different versions of the FHIR modelinfo version 1.0.2 floating around.

I think that the FHIR modelinfos (and perhaps others as well) need to somehow support versioning independent of the version of what they're modeling.  This means putting the FHIR version number in the model _name_ and starting at version _1.0_ of the modelinfo, or needing to introduce a 4th component to the version number, where components 1-3 match the FHIR version, but component 4 is incremented w/ changes to the modelinfo file (but at that point we are breaking semantic versioning).

brynrhodes:
As part of the functionality introduced in CQL 1.2, we added the ability to represent choice types in CQL, and with the release of that version of the translator, models that could take advantage of that functionality began to. You're right that it's a breaking change though, in that we only offer the choice type, not the explicitly-typed attributes. 

[...]

PR #242, opened 2017-09-28
Revert fhir-modelinfo-1.0.2.xml

cmoesel:
Reverting fhir-modelinfo-1.0.2.xml to previous commit (ef0eef9dfa8b182a6ab4167b16aaa2e32934ae1f).

For more information, see: https://github.com/cqframework/clinical_quality_language/issues/240

**NOTE**: It appears that the most recent version changed _more_ than just the Choice types.  It also:
* Removed `FHIR.` prefix from typeInfo names (e.g., `FHIR.Observation` --> `Observation`)
* Changed the way lists are represented from:
   ```xml
   <ns4:element name="useContext" type="list&lt;FHIR.CodeableConcept&gt;"/>
   ```
   to
   ```xml
   <ns4:element name="useContext">
       <ns4:typeSpecifier xsi:type="ns4:ListTypeSpecifier" elementType="FHIR.CodeableConcept"/>
   </ns4:element>
   ```

Since this PR reverts to the previous version, it also _drops_ the changes listed above.  Is that OK, or is there something fundamentally wrong with the previous representations?


brynrhodes:
Yes, those changes are fine, a model info loader should be able to deserialize with or without the namespace, and with either the inline representation or the structured representation for types.

cmoesel:
Thanks, @brynrhodes!  When is the next scheduled release when this change will be available via the Maven repository?

PR #244, opened 2017-09-29
Remove version check for equivalency operator.

losborne:
Affects Code, Value Set, and Concept equivalency.
Equivalency should no longer return false if the only difference is in version.

Modified Equivalent class to use the equivalent function in comparisn.js rather
than a different but similar expression.

The only tests that failed due to this change were expected (tests that were
looking for different versions). I updated those tests to expect equivalence if
the only difference is in version.

JIRA Ticket: https://jira.mitre.org/browse/BONNIE-1102
Corresponding cql_qdm_patientapi PR: https://github.com/projecttacoma/cql_qdm_patientapi/pull/84


losborne:
Regression script was run and passed -- all results equal.

cmoesel:
Assuming we need to wait for @jbradl11 review before merging, but let me know if otherwise.

Src/coffeescript/cql-execution/test/util/comparison.coffee
```diff
@@ -113,3 +113,6 @@ describe 'equivalent', ->
     equivalent('123', '123').should.be.true()
     equivalent(123, 123).should.be.true()
     equivalent('123', new Code('123', 'test', '2016')).should.be.false()
+
+  it 'should consider codes with different versions to be equivalent', ->
```
hossenlopp: I think a test should be added to check equivalence between code and concept/code lists.

losborne: Added some below.

losborne:
Should be fixed now--missed one name change when I went to make the lines a bit shorter.

PR #249, opened 2017-10-06
fix null check for equivalency

losborne:
This is a fix to the null check for equivalency.

Corresponding cql_qdm_patientapi PR: https://github.com/projecttacoma/cql_qdm_patientapi/pull/85


hossenlopp:
second PR projecttacoma/cql_qdm_patientapi#86

PR #253, opened 2017-10-18
Removed eval usage, now support spaces in names  of cql defined funct…

jbradl11:
Removed old code and refactored the builder file to no longer use `eval`.


Src/coffeescript/cql-execution/src/elm/builder.coffee
```diff
@@ -7,15 +7,11 @@ module.exports.build = build = (json) ->
   if (typeIsArray json)
     return (build child for child in json)
 
-  if json.type is "FunctionRef" then new buildFunctionRef(json)
+  if json.type is "FunctionRef" then new E.FunctionRef(json)
   else if json.type is "Literal" then E.Literal.from(json)
-  else if functionExists("E.#{json.type}") then constructByName("E.#{json.type}", json)
+  else if functionExists(json.type) then constructByName(json.type, json)
```
cmoesel: I realize we wrote this code together, but it occurs to me that we could probably streamline it by removing the custom functions here.  So it just becomes:
```
else if typeof E[json.type] is "function" then new E[json.type](json)
```
and those other functions go away.  What do you think?

cmoesel: If you've already done a bunch of testing and integration work with it as-is, I am totally fine with just leaving it too.  It's up to you.

jbradl11: I thought about refactoring this before creating the PR, but I thought this was easier to understand (using the function names). I am open to changing it.

cmoesel: Nope, that's fine then.  Shall I merge it or do you need a second review from the Bonnie team?

jbradl11: I am going to have a second person review it, first. Then I can reach out to you to merge it.

PR #254, opened 2017-10-20
Regenerated cql4browsers using the newer cake build-cql4browsers command

jbradl11:
Latest cql4browsers was not generated properly. This PR is to update cql4browsers.js


hossenlopp:
Let me run regression tests then we'll call it good.

cmoesel:
So... shall I merge?

losborne:
@cmoesel Yes, go ahead and merge.

Issue #256, opened 2017-11-05
offsetPolarity in visitTimeLiteral causing negative timezone offsets to always result in 0.0


c-schuler:
I believe this [line](https://github.com/cqframework/clinical_quality_language/blob/0237ba278ec40d69f04aeda4e2c7a45012da9558/Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/Cql2ElmVisitor.java#L905):

`int offsetPolarity = matcher.group(12).equals("+") ? 1 : 0;`

 should be 

`int offsetPolarity = matcher.group(12).equals("+") ? 1 : -1;`

When the offsetPolarity is zero, then the hour part of the timezone offset will always result in 0.0 because of the following [line](https://github.com/cqframework/clinical_quality_language/blob/0237ba278ec40d69f04aeda4e2c7a45012da9558/Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/Cql2ElmVisitor.java#L925):

`result.setTimezoneOffset(libraryBuilder.createLiteral((double)(hourOffset * offsetPolarity)));`

brynrhodes:
Included in release 1.2.18

PR #260, opened 2017-11-22
#256: Fixed an issue with negative time zones offsets in DateTime and…

brynrhodes:
… Time literals being translated as a 0.


brynrhodes:
Agreed, positive polarity tests added.

Issue #270, opened 2017-12-21
Add support for the Jan2018 Ballot version of FHIR


brynrhodes:
The Jan2018 ballot version of FHIR has been published [here](http://hl7.org/fhir/2018Jan/). Support for this version of the specification should be added (FHIR version 3.2.0).

brynrhodes:
Included in release 1.2.18

PR #276, opened 2017-12-29
#270: Added support for FHIR version 3.2.0.

brynrhodes:


brynrhodes:
Agreed on the type specifiers, I've updated the helpers library to be consistent with previous versions of the library.

On the missing conversions, spot-checking the list they are all legitimate differences between R3 and R4. For example, ActivityDefinition.status is now a PublicationStatus.

Issue #275, opened 2017-12-29
Errors translating a referenced library are not correctly reported to the referencing translation


brynrhodes:
When errors occur during the translation of an included library, they are not correctly reported to the base translator. The translation still fails appropriately, but the errors are not output to the console, so the translation appears to fail for no reason.

brynrhodes:
Included in release 1.2.18

PR #277, opened 2017-12-29
#275: Fixed errors translating included libraries not being correctly…

brynrhodes:
… reported to the including translator.


brynrhodes:
I've added a test to verify the behavior, and doing so uncovered a potential issue with the library manager caching the results of a translation of a library with errors.

Issue #282, opened 2018-01-17
DefaultLibrarySourceProvider can load incorrect library version


brynrhodes:
If an include declaration includes a version but the source for a different version of the library is available, the DefaultLibrarySourceProvider will incorrectly load the most recent version of the library, even if that version is incompatible with the requested version. This results in an error once the translated library is compared against the requested version, rather than the correct behavior of only loading library sources that are compatible with the requested version.

brynrhodes:
This feature is included in release [1.2.19](https://github.com/cqframework/clinical_quality_language/releases/tag/v1.2.19)

PR #283, opened 2018-01-17
#282: Fixed the DefaultLibrarySourceProvider loading incorrect source…

brynrhodes:
… version when requested version is not available.

### DIFF
--- a/Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/model/Version.java
+++ b/Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/model/Version.java
@@ -37,6 +37,40 @@ public class Version implements Comparable<Version> {
         return 0;
     }
 
+    public boolean compatibleWith(Version that) {
+        if (that == null)
+            return false;
+
+        // this version is compatible with that version if:
+            // this.major = that.major
+            // this.minor >= that.minor
+        String[] thisParts = this.version.split("\\.");
+        String[] thatParts = that.version.split("\\.");
+        int length = Math.max(thisParts.length, thatParts.length);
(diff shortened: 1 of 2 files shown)


Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/model/Version.java
```diff
@@ -37,6 +37,40 @@ public int compareTo(Version that) {
...
+        int length = Math.max(thisParts.length, thatParts.length);
+        for (int i = 0; i < length; i++) {
+            int thisPart = i < thisParts.length ? Integer.parseInt(thisParts[i]) : 0;
+            int thatPart = i < thatParts.length ? Integer.parseInt(thatParts[i]) : 0;
+            switch (i) {
+                case 0: {
+                    // Major version
+                    if (thisPart != thatPart)
+                        return false;
+                }
```
cmoesel: Note that Java switch statements fall through until they hit a `return` or a `break`.  So, if the major case (`0`) is hit, but the parts are equal, it will fall through and execute the code in the minor case (`1`).  Now... since it's impossible for a part to be `==` and `>` at the same time, and since the patch case (`2`) is empty, this logic actually works as-written, but...  it _looks_ like a bug, and I think that in order to show the intent, there really should be a break at the end of `case 0` and `case 1`.

brynrhodes: Totally agree, great catch. I'm too used to C# where the compiler will tell me when I've missed those.

brynrhodes:
Fantastic, thank you. When you get a chance, can you merge it?

Issue #286, opened 2018-01-23
hasValue() is not resolving in FHIRPath expressions


brynrhodes:
The following constraint evaluation from the FHIR specification fails to translate:

    library "ele-1"
    using FHIR version '3.0.0'
    parameter Context Patient
    define Constraint: Context.select(hasValue() | (children().count() > id.count()))

The error given is "could not resolve method hasValue()"

brynrhodes:
This feature is included in release [1.2.19](https://github.com/cqframework/clinical_quality_language/releases/tag/v1.2.19)

PR #289, opened 2018-01-28
#286: Added hasValue() resolution for FHIRPath.

brynrhodes:
Fixed hasValue() not resolving


olbrich:
Would it make sense to generalize this method a bit to something like `hasChild('value')` ?

PR #291, opened 2018-01-29
Cql4bonnie add coverage

losborne:
This PR is to add tests to improve coverage in preparation with merging branch `cql4bonnie` into master.  After this PR has been merged, a new PR into master from cql4bonnie will occur.

To determine which changes lacked coverage, I first did a merge of `master` into `cql4bonnie`, then used `nyc` to show the coverage.  I've included `nyc` as a dependency in `package.json`, so now coverage is viewable after running `cake test` by opening the file `Src/coffeescript/cql-execution/coverage/index.html`.

From those findings, I have generated a list of which lines of new code in `cql4bonnie` did not have coverage.  I've either added coverage or commented as to how this is not possible:

File | Coverage | Notes
-- | -- | --
Src/coffeescript/cql-execution/.vscode/launch.json | no tests needed |  
Src/coffeescript/cql-execution/Cakefile | no tests needed |  
Src/coffeescript/cql-execution/package.json | no tests needed |  
Src/coffeescript/cql-execution/src/cql.coffee | 100%   covered |  
Src/coffeescript/cql-execution/src/datatypes/clinical.coffee | 100%   covered |  
Src/coffeescript/cql-execution/src/datatypes/datetime.coffee | 178: else Iif b.year? and b.month? and b.day? and b.hour? then b   = b.add(tzDiff/60, DateTime.Unit.HOUR)     215 : if not(other instanceof DateTime) then return null | Added   some DateTime Tests, but this block of code was never hit.  It seems the timezone offset of b is   changed to match a before this block, so I'm not sure how to reach this   statement.
Src/coffeescript/cql-execution/src/datatypes/interval.coffee | 100%   of new code is covered |  
Src/coffeescript/cql-execution/src/datatypes/uncertainty.coffee | 9, 18, 23:     if   typeof a !=   typeof   b             return null | coverage   added and bug fixed, null -> false
Src/coffeescript/cql-execution/src/elm/aggregate.coffee | 100%   covered |  
Src/coffeescript/cql-execution/src/elm/arithmetic.coffee | 100%   of new code is covered |  
Src/coffeescript/cql-execution/src/elm/builder.coffee | 100%   covered |  
Src/coffeescript/cql-execution/src/elm/clinical.coffee | 38: return false unless @code? and   @valueset? | coverage   added
Src/coffeescript/cql-execution/src/elm/comparison.coffee | 21,35,49,63 :         throw error | Don't   really think it's necessary to have unit test for these errors
Src/coffeescript/cql-execution/src/elm/conditional.coffee | 100%   covered |  
Src/coffeescript/cql-execution/src/elm/datetime.coffee | 128, 141: return null   (TODO need to address DifferenceBetween crossing   DST) | implemented   DST changes and un-skipped associated unit tests.
Src/coffeescript/cql-execution/src/elm/expression.coffee | (low coverage for this file) 10: @localId = json.localId | turned   on annotations and updated tests accordingly.    There is one test in particular that now has a hard-coded localId in   the Expect that should be looked at carefully.
Src/coffeescript/cql-execution/src/elm/external.coffee | 100%   covered |  
Src/coffeescript/cql-execution/src/elm/instance.coffee | 100%   covered |  
Src/coffeescript/cql-execution/src/elm/interval.coffee | 100%   covered |  
Src/coffeescript/cql-execution/src/elm/list.coffee | 100%   of new code is covered |  
Src/coffeescript/cql-execution/src/elm/nullological.coffee | (low   coverage for this file) 21-24:     exec: (ctx) ->         for   arg in @args           result = arg.execute(ctx)           if result? then return result | un-skipped   tests and fixed code to properly handle lists
Src/coffeescript/cql-execution/src/elm/overloaded.coffee | 25:         throw   error     44-51:     exec: (ctx) ->         try           ThreeValuedLogic.not equals @execArgs(ctx)...         catch   error           if   error instanceof IncompatibleTypesException             return null           else             throw   error | NotEqual   is actually implemented as Not followed by Equal, so this code is never   hit.  There are comments in the code   explaining so.
Src/coffeescript/cql-execution/src/elm/parameters.coffee | 100%   covered |  
Src/coffeescript/cql-execution/src/elm/quantity.coffee | 82-83:        value = if   operator == "/" then @value / other.value  else @value   * other.value             unit = @unit   \|\| other.unit   169: units_to_string =   (units = {})   190-191: ucum_multiply =   (t, ms=[]) ->       return   t if ms.length   == 0 | It   seems that this block is intended to deal with the case when one has a unit   and the other doesn't.  There are   already tests for that case, and they don't hit this block of code.
Src/coffeescript/cql-execution/src/elm/query.coffee | 100%   covered |  
Src/coffeescript/cql-execution/src/elm/reusable.coffee | 100%   of new code is covered |  
Src/coffeescript/cql-execution/src/elm/string.coffee | 100%   covered |  
Src/coffeescript/cql-execution/src/elm/structured.coffee | 100%   of new code is covered |  
Src/coffeescript/cql-execution/src/elm/type.coffee | 100%   of new code is covered |  
Src/coffeescript/cql-execution/src/example/browser/README.md | ( no   tests needed ) |  
Src/coffeescript/cql-execution/src/example/browser/cql4browsers.js | ( not   tested with unit tests ) |  
Src/coffeescript/cql-execution/src/example/browser/simple-browser-support.coffee | ( not   tested with unit tests ) |  
Src/coffeescript/cql-execution/src/runtime/context.coffee | 89-90: # Temporary fix. Real fix will be to return a list of all result values for a given localId.         ctx = @localId_context[localId] |  
Src/coffeescript/cql-execution/src/runtime/executor.coffee | 22: r.recordPatientResult(patient_ctx, expression, expr.execute(patient_ctx)) | this   is only an api syntax change, the original call is not covered in master
Src/coffeescript/cql-execution/src/runtime/results.coffee | 100%   covered |  
Src/coffeescript/cql-execution/src/util/comparison.coffee | 100%   covered |  
Src/coffeescript/cql-execution/src/util/math.coffee | 84-85:   if typeof exp == 'undefined' \|\| +exp == 0         return Math[type](value)   89-90:   if isNaN(value) \|\| !(typeof exp == 'number' && exp % 1 == 0)         return NaN | decimalAdjust   is only ever called with exp of -8 in the codebase, so these lines are   untestable in the context of the repo.


Src/coffeescript/cql-execution/src/datatypes/datetime.coffee
```diff
@@ -201,7 +204,17 @@ module.exports.DateTime = class DateTime
       a = new DateTime(a.year, a.month, a.day, a.hour, a.minute, a.second, 0, a.timezoneOffset)
       b = new DateTime(b.year, b.month, b.day, b.hour, b.minute, b.second, 0, b.timezoneOffset)
 
-    a.durationBetween(b, unitField)
+    if unitField == DateTime.Unit.YEAR || unitField == DateTime.Unit.MONTH
+      a.durationBetween(b, unitField)
+    else
+      a = a.toUncertainty()
```
jbradl11: I think the new "a" and "b" variables should be named something different. Thoughts?

Src/coffeescript/cql-execution/src/datatypes/datetime.coffee
```diff
@@ -201,7 +204,17 @@ module.exports.DateTime = class DateTime
       a = new DateTime(a.year, a.month, a.day, a.hour, a.minute, a.second, 0, a.timezoneOffset)
       b = new DateTime(b.year, b.month, b.day, b.hour, b.minute, b.second, 0, b.timezoneOffset)
 
-    a.durationBetween(b, unitField)
+    if unitField == DateTime.Unit.YEAR || unitField == DateTime.Unit.MONTH
```
jbradl11: I would add a comment for this if / else block.

Src/coffeescript/cql-execution/test/elm/clinical/test.coffee
```diff
@@ -171,7 +174,12 @@ describe 'CalculateAge', ->
   it 'should execute age in months', ->
     # what is returned will depend on whether the day in the current month has
     # made it to the 17th day of the month as declared in the birthday
-    [@full_months, @full_months+1].indexOf(@months.exec(@ctx)).should.not.equal -1
+    dayOfMonth = @today
+    for i in [1 .. 28]
+      dayOfMonth.setDate(i)
```
jbradl11: Please add an additional comment here. The logic is confusing to me.

jbradl11:
You should re-generate cql4browsers.js in this PR.

losborne:
Added cql4browsers and opened cql_qdm_patientapi PR for it here: https://github.com/projecttacoma/cql_qdm_patientapi/pull/97/files

Src/coffeescript/cql-execution/package.json
```diff
@@ -36,9 +36,12 @@
     "coffee-script": "^1.12.0",
     "coveralls": "^2.13.1",
     "mocha": "^3.4.2",
+    "nyc": "^11.4.1",
     "should": "^11.2.1"
   },
   "dependencies": {
+    "istanbul": "^0.4.5",
```
cmoesel: Why is `istanbul` a normal dependency rather than a `devDependency`? Isn't it only needed for collecting coverage on test runs?

Src/coffeescript/cql-execution/src/elm/nullological.coffee
```diff
@@ -20,6 +20,12 @@ module.exports.Coalesce = class Coalesce extends Expression
 
   exec: (ctx) ->
     for arg in @args
-      result = arg.execute(ctx)
-      if result? then return result
+      # Lists are treated differently
```
cmoesel: I assume this is to cover the last signature in the documented cases:
```
Coalesce<T>(argument1 T, argument2 T) T
Coalesce<T>(argument1 T, argument2 T, argument3 T) T
Coalesce<T>(argument1 T, argument2 T, argument3 T, argument4 T) T Coalesce<T>(argument1 T, argument2 T, argument3 T, argument4 T, argument5 T) T
Coalesce<T>(arguments List<T>) T
```
In this case, shouldn't we also be ensuring that length of `@args` is `1`?  If the length of `@args` is more than one, we should probably be treating it like the multi-variable case instead.

cmoesel: Additionally, if there is only a single argument, but it is an `ExpressionRef` pointing to a list (rather than a List literal), then I think it wouldn't have `arg.elements` -- you'd have to execute the arg to find out it's a list.  I think.

Src/coffeescript/cql-execution/test/datatypes/datetime.coffee
```diff
@@ -443,9 +448,7 @@ describe 'DateTime.differenceBetween', ->
     a.differenceBetween(b, DateTime.Unit.SECOND).should.eql new Uncertainty(0)
     a.differenceBetween(b, DateTime.Unit.MILLISECOND).should.eql new Uncertainty(0)
 
-  # NOTE: Skipped for the purposes of Bonnie 11/1 release. Needs to be addressed
-  # afterwards.
-  it.skip 'should handle crossing DST in the spring', ->
+  it 'should handle crossing DST in the spring', ->
```
cmoesel: You only had to take out the `.skip` and it worked.  Nice!

Src/coffeescript/cql-execution/test/datatypes/uncertainty.coffee
```diff
@@ -24,6 +24,10 @@ describe 'Uncertainty', ->
     should(everything.low).be.null
     should(everything.high).be.null
 
+    differentTypes = new Uncertainty(1, "1")
+    differentTypes.low.should.eql 1
+    differentTypes.high.should.eql "1"
```
cmoesel: This is fine in that it documents the current behavior, but _really_, I don't think it should be allowed to have the low and high as different types.  Is there a use case?  That should probably _throw_...

If you are making changes anyway, can you just comment that this probably _shouldn't_ be supported?

Src/coffeescript/cql-execution/test/elm/interval/test.coffee
```diff
@@ -33,7 +33,7 @@ describe 'Interval', ->
     @closed.high.exec(@ctx).should.eql new DateTime(2013, 1, 1)
 
   it 'should exec to native Interval datatype', ->
-    ivl = @open.exec(@cql)
+    ivl = @open.exec(@ctx)
```
cmoesel: Ha.  Oops.  That bug must have been written on a late night with too much CQL on the mind.

Src/coffeescript/cql-execution/src/elm/nullological.coffee
```diff
@@ -20,11 +20,12 @@ module.exports.Coalesce = class Coalesce extends Expression
...
-      if arg.elements?
-        for item in arg.elements
-          result = item.execute(ctx)
-          if result? then return result
+      # A single List (ExpressionRef->List) as the argument is treated differently
+      if @args.length == 1 && @args[0].constructor.name in ['List', 'ExpressionRef']
+        list = arg.execute(ctx)
+        if list?
+          for item in list
+            if item? then return item
```
cmoesel: So...  I probably wasn't specific enough in my comment.  I was using `ExpressionRef` as an example of something that might _execute_ to a List (as opposed to a List literal).  There are other things that could execute to a list as well.  For example, a retrieve:
```
Coalesce([Encounter]
```
or a union:
```
define ListA = List{null, null, 3}
define ListB = List{4, null, 6}
define Three = Coalesce(ListA union ListB)
```

So, I think the code should maybe be something more like:
```coffee
    for arg in @args
      result = arg.execute(ctx)
      # if a single arg, and it's a list, then it's expected to coalesce over that list
      if @args.length == 1 && Array.isArray(result)
        for item in result
          if item? then return item
      else
        if result? then return result
    null
```
I didn't try compiling, but hopefully that conveys the idea well enough.

cmoesel:
So...  can I merge this into cql4bonnie?  Was the cql4browsers tested in Bonnie to ensure there were no regressions?

losborne:
@cmoesel All clear to merge.  Just checked for regressions on bonnie and there were none.

PR #299, opened 2018-02-07
Execution Fixes to Align w/ Spec

cmoesel:
This contains several fixes to align better w/ the CQL specification, as well as a few house-keeping items.  Since this changes the result of some logic (and the structure of the `Concept` class), this should be tested in Bonnie before merging.

**1. Concept should have "display" property (not "text")**

According to CQL documentation, the `Concept` type has `codes` and `display`.  Our representation incorrectly had `text` instead of `display`.  If any Bonnie code is currently referencing `.text` on a `Concept`, it will have to be changed to `.display`.

**2. Support creating instances of Concept**

Now CQL that constructs a `Concept` (see below for exmaple) will properly return a `Concept` class:
```
Concept {
  codes: { CodeA, CodeB },
  display: 'Test Concept'
}
```

**3. Fix instance comparison logic in equals**

There was an edge case where a subclass instance might be considered `equal` to a superclass instance -- now equality will be `false` for instances that aren't the same exact type.

**4. Fixes related to `null` handling**

Several fixes to match the spec regarding when `null` should be returned:
- `InValueSet` should return `null` when the `code` is `null`
- `Equal` should return `null` when either value is `null`
- Path accessors should return `null` (instead of undefined) when the path doesn't exist or has no value
- `SingletonFrom` should return `null` when the input is `null` (previously, it crashed)

**5. Upgrade coffeescript dependency to 1.12.7**

This matches the version that _was_ used to generate cql4browsers prior to the last PR.  The last PR reverted cql4browsers back to 1.12.4.  This brings it back and ensures that if the packaged coffeescript is used, it will be the right version.  That said, the generated code doesn't appear to be different (except for the changes explicitly introduced in this PR).

**6. Fix warning text in generated data.coffee files**

The generated data.coffee files have a warning about modifying them and suggest how to add tests.  They incorrectly suggested editing `data.coffee`.  Now they correctly suggest editing `data.cql`.


Src/coffeescript/cql-execution/test/util/comparison.coffee
```diff
@@ -51,7 +51,8 @@ describe 'equals', ->
 
     equals(new Foo('abc', [1,2,3]), new Foo('abc', [1,2,3])).should.be.true()
     equals(new Foo('abc', [1,2,3]), new Foo('abcd', [1,2,3])).should.be.false()
-    equals(new Foo('abc', new Bar('xyz')), new Foo('abc', new Bar('xyz'))).should.be.true()
+    equals(new Foo('abc', new Bar('xyz', [1,2,3])), new Foo('abc', new Bar('xyz', [1,2,3]))).should.be.true()
+    equals(new Foo('abc', new Bar('xyz')), new Foo('abc', new Bar('xyz'))).should.be.false() # because Bar is missing components
```
losborne: This behavior seems strange to me.  What I found in the spec said:

> For structured values, equality returns true if the values being compared are the same type (meaning they have the same types of elements) and the values for each element are the same value.

Maybe the wording "the values for each element are the same value" is the confusing part, because the spec is clear that `equals(null, null)` should return `null`, but I would interpret "are the same value" to be true for `null` and `null`.

Are you confident this is the right behavior in this case?

cmoesel: That's a fair question.  I'll try to channel the "December Chris Moesel" to remember how I arrived at that conclusion.  Since there is a test for it, it must have been intentional -- but I agree that it's a bit odd.

If it's wrong, that means we'll have to special-case the equality comparison when assessing equality of components within a structure.

cmoesel: In addition to the quotes from the spec that Luke mentions above, I found the following...  and must admit there is still some ambiguity on the matter:

> **9.5.2 Equal**
> **Signature:**
> `=<t>(left T, right T) Boolean`
> **Description:**
> _snipped some text..._
> For tuple types, this means that equality returns true if and only if the tuples are of the same type, and the values for all elements by name are equal.

The phrase "all elements by name are equal" would seem to imply that the semantics of "equal" should be used on every named element (justifying the current implementation).

> **9.12.5 Equal**
> **Signature:**
> `=(left Code, right Code) Boolean`
> `=(left Concept, right Concept) Boolean`
> **Description:**
> The equal (=) operator for Codes and Concepts uses tuple equality semantics. This means that the operator will return true if and only if the values for each element by name are equal.
> If either argument is null, or contains any null components, the result is null.

Although the above is specific to `Code` and `Concept`, it indicates that it uses "tuple equality semantics", and then further qualifies "If either argument is null, _or contains any null components_, the result is null."  This seems to confirm that in tuple equality, `null` components result in `null`.

**BUT...**

Buried in section 2.5.6.3, we find a parenthetical statement that seems to suggest that null components _should_ be considered the same for tuples:

> (i.e. tuples are equal if they have the same type and the same values (or no value) for each element of the same name).

So...  I think that the specification is not as clear as I'd like it to be, but it _mostly_ suggests that the current implementation (in this PR) is correct.  Does this implementation cause issues or regressions for Bonnie?  If _no_, then I suggest we move forward as-is, but we _also_ submit an issue against the spec asking for clarification and justification as to the intended approach.

losborne: There were no regressions reported by the regression test in Bonnie, so I agree with you that while the spec seems to be a little unclear on this that the current implementation makes sense.

Everything else in the code looks good to me, and the tests pass.  Approved

p9g:
Note that the CQL-Based HQMF IG, Vol 2, section 2.6 has Conformance Requirement 9 (Concepts): The CQL construct, **concept**, SHALL NOT be used.

cmoesel:
@p9g -- Thanks for pointing that out.  I actually added support for constructing `Concept` because it was needed by another project that executes CQL for CDS logic.  It sounds like it will be inconsequential to Bonnie though...  That's good -- less things to break. ;-)

Src/coffeescript/cql-execution/src/elm/instance.coffee
```diff
@@ -24,4 +24,5 @@ module.exports.Instance = class Instance extends Expression
     switch @classType
```
jbradl11: We can remove this #TODO now correct?

cmoesel: Why yes, I believe we can!

cmoesel:
OK, I've removed the out-of-date comment.  @losborne or @jbradl11 -- may I merge this now?

PR #308, opened 2018-03-12
Fix 195 ucum validation

brynrhodes:


cmoesel:
Haven't tried it, but in general it looks sound.  I'd consider adding some tests for the combinatorial or other special representations, such as fractional units (e.g., `mg/dL`), squared units (e.g., `mm2`), "per" units (e.g., `/mg`), multiplicative units (e.g., `10*3`), etc.

Also, I think we've indicated that _no units_ should be treated like `1`.  What happens if you try to validate a unit that is `null` or empty string?  I don't see any special case code for that.

brynrhodes:
Thanks for the feedback @cmoesel. As far as the more in-depth tests, we're using a UCUM validation package that has its own unit testing, so the intent of these tests is just to verify that the code path for UCUM validation is working.

As far as the behavior of quantities with no units, that's more of a run-time consideration. We should add some unit tests to check for that behavior as part of the engine test suite, but I don't think it makes sense here.

cmoesel:
@brynrhodes -- thanks for the clarification regarding the purpose of the tests.  I agree, then, that we don't need to cover all the possible amalgamations.

Regarding behavior of quantities without units, I was talking about the validation during the CQL-to-ELM process.  Are you saying there isn't any way to create a unit-less quantity except at run-time?  Is this validation only happening on quantity literals (e.g. `Quantity { value: 2 }` won't be validated)?

brynrhodes:
@cmoesel, that's correct, the validation is only happening on the literal selector. Anywhere else would require evaluation because it's a generic expression.

brynrhodes:
Actually, both `12 'years'` and `12 'YEARS'` will fail individually because `years` and `YEARS` are not UCUM units. `years` is a CQL keyword that is interpreted as the UCUM unit `a`.

The question stands though, should equivalent ignore case-sensitivity of UCUM units, and the answer is no, equivalence still needs to respect UCUM units, which are case-sensitive.

brynrhodes:
And I just verified, the inconsistency is necessary, without the class get resourceAsStream, the tests fail.

brynrhodes:
@cmoesel, @c-schuler, do either of you have any objections to the PR or can we merge it?

c-schuler:
I created a test and `12 'years'` does translate. Should it not? `12 'YEARS'` does throw an error.

brynrhodes:
Looks like I misspoke, the units are actually 'years' there. I think that may be an issue, but it would need community feedback, so should be logged as a separate issue, it's independent of this PR.

PR #318, opened 2018-04-12
Remove CoffeeScript cql-execution engine (moved)

cmoesel:
Removing all source code related to the cql-execution engine since it
now exists in its own repository.  The README.md and OVERVIEW.md
have been retained to inform people where the new code it and since
these pages may be linked from external sources (like the spec).

See: https://github.com/cqframework/cql-execution/


cmoesel:
@edeyoung -- are you fine with the cql-execution code being removed from the clinical_quality_language repo?

PR #334, opened 2018-05-03
fix qdm 4.3 bug

### LINKED ISSUE #292: QDM 4.3 cannot resolve "birth datetime"
I have the following cql

```
using QDM version '4.3'

context Patient

define "test1":
        AgeInYearsAt(@2017-01-01T00:00:00.000)

```

I attempted to compile it and got the following:

```
================================================================================
TRANSLATE ./Test.cql
Translation failed due to errors:
Error:[6:2, 6:39] Member birth datetime not found for type QDM.Patient.
```

I then fixed this by modifying line 5 of Src/java/qdm/src/main/resources/gov/healthit/qdm/qdm-modelinfo-4.3.xml

from
`        <ns4:element name="birthDatetime" type="System.DateTime"/>`
to
`        <ns4:element name="birth datetime" type="System.DateTime"/>`

then rebuilt the compiler and ran it again and it was successful.

I'm not sure if the solution is just to submit this as a PR so I posted here just to make sure it was brought to your attention.

athenaecascketta:
Fixes #292


cmoesel:
I don't know if there have been conversations that I've not been a part of, but... I see that the bug was that the property identified in line 2 as the `patientBirthDatePropertyName` (`"birth datetime"`) didn't match the element name defined in line 5 (`"birthDatetime"`).  This would obviously cause problems.

That said, it seems to me that the more consistent solution would be to fix line 2 to indicate `"birthDatetime"`.  All of the other element definition names use camelCase without spaces, so it seems odd to have _one_ that breaks that rule.

Again, perhaps there were discussions in which this was deemed the best fix -- and if so, you can ignore me -- but from my perspective, I think it's better to fix the inconsistent `birth datetime` to be `birthDatetime` than the other way around.

brynrhodes:
@cmoesel you're right, the correct fix is to update the `patientBirthDatePropertyName`.

athenaecascketta:
Makes sense, will do that instead

PR #404, opened 2019-03-26
Support for generic types

cnanjo:
Hi Bryn,

I have added support for Generic types along the lines of what we discussed a while back. Let's discuss when you get a chance as some of this code is a bit complex. It might be easier if I walk you through some of this code. I ran all the tests and they passed. I modified the structure of model-info to support explicit generic parameters. I did not update all the model-info files. I can do this using an XSLT but let's discuss first.


cmoesel:
I was not a part of the conversation regarding the introduction of generics, why it's needed, how it will be used, etc -- so I won't comment on that.  But... I am concerned about the breaking change to the modelinfo schema that requires pretty much every class in every modelinfo to be updated.  We don't own all the modelinfos -- others have written their own and will be affected.  It's also possible that others have written tools that process modelinfos for run-time support of models (for example, I've done that to support FHIR in our execution engine).  

It looks like the schema could be updated to support generic types without wrapping an `elements` tag around the existing `element` tag.  If we were starting from scratch, I'd agree that having the `elements` wrapper is probably a better approach -- but with existing implementations I think it's better to favor non-breaking changes even if they're stylistically imperfect.  If I'm wrong, and the new `elements` tag is absolutely necessary, I'd prefer to align this change with a ballot.  I understand that technically modelinfo is not a normative part of the spec, but if we're going to introduce a breaking change, I'd rather do it when people are _expecting_ changes -- and have the opportunity to alert them to it.

cnanjo:
Hi Chris, actually there is a notion of generics already in the ModelInfo. It is just hard coded - List and Intervals. The idea is to make generics fully available to support models that use generics.
Regarding ModelInfo breaking changes, yes, I agree with you and was reluctant with that change because it breaks backward compatibility (though I used a simple XSLT to bring older files to compliance). If we need to simply add parameters and elements in the same sequence to avoid a breaking change, so be it. (I don't really like it but I am pragmatic). While we may discuss the merits of either approach (and the cons obviously for the proposed change), I strongly recommend against a 'ballot' of modelinfo. It is not a standard metamodeling language. It is one of the strength of ModelInfo actually in that it can evolve more freely that way based on the needs we see in the field. If we are to standardize it, then I would push back since I don't understand why we would reinvent the wheel. I have worked closely now with this metamodel and realize that we duplicated essentially (again) what OpenEHR does quite well which is unfortunate. Hence, discussing with the community how we wish to proceed - a good idea. Balloting this proposal - bad idea :)

cmoesel:
Thanks, @cnanjo.  I'm sorry -- I didn't mean to suggest that modelinfo should actually be balloted, but that we should align the change with a ballot timeframe -- or at least with an official STU update.  I realize we don't have to, but modelinfo has changed very little since R1, and CQL-to-ELM is used by many, so I'm just afraid of catching people off-guard if we release a significant breaking change to modelinfo at an arbitrary point in time.  Once it's released, CQL-to-ELM won't work with old modelinfos anymore.

If it _doesn't_ have a breaking change, however, I'm fine with applying the update at any point in time.  So I guess first we should decide if we want the more elegant, but backwards-incompatible approach, or if we want the non-breaking approach that is less elegant.  @brynrhodes?

cnanjo:
I defer to both of you on the decision. If we decide against the wrapper element, please reject/delete the pull request and I will make the fix and resubmit the next version once I verify that all tests pass.

brynrhodes:
Hi @cnanjo, thanks so much for the submission, and @cmoesel , thanks so much for the review. I completely agree that we should not introduce breaking changes in the model info, and I actually added some import logic to "derive" the Patient context in 1.4 from the explicit attributes to maintain backwards compatibility.

So if it's alright, I'd suggest we go with the non-breaking change. Sorry if that causes rework Claude, but can you withdraw and resubmit the PR with the changes Chris suggested?

cmoesel:
FYI -- I did not do a careful review of the code, so I only identified the one obvious breaking change.  If there are others, I guess those should be addressed too -- but I *think* most of Claude's work is _additive_ so I'm guessing it's mostly fine.

cnanjo:
The work is additive I believe apart from the ModelInfo changes. There are some improvements, additional logic, etc... that need to be made and which I will make but I am thinking we check in this first version (with the fix to ModelInfo) and I will follow it with another pull request with some additional clean up, etc... just in case given that this is fairly time-intensive.

cnanjo:
Closing pull request.

cnanjo:
I updated the code to remove the backward compatibility issues that existed. I ran all the tests during the rebuild. It should be good at this time.

cnanjo:
Folks, I have pushed some updates based on a quick review of the code with Bryn.

brynrhodes:
@cmoesel , given this is backwards compatible now, any concerns with me merging it in?

cmoesel:
@brynrhodes -- as long as it is backwards compatible and you've confirmed it doesn't break any existing capability (e.g., no regressions), I have no issues with it in general.

BUT... there are two things that I noticed:

* The branch has conflicts (in modelinfo.xsd) that need to be resolved.  Ideally we should test against the _resolved_ code before merging.
* This PR appears to include your new `ContextInfo` and `RelationshipInfo` types -- which are _not_ yet merged into master.  I'd prefer to see those land in master via a PR specific to that capability (instead of trojan horsing this PR).

cnanjo:
I think the issue with the ModelInfo are whitespaces introduced by the editor I was using. I will fix that. ContextIfo and RelationshipInfo should be reverted to what is in the branch. Not sure why it grabbed those. I did not modify these.

brynrhodes:
@cnanjo , @cmoesel , I've got a local branch where I'm resolving conflicts and merging. The context info was committed directly to v14, so it's already present there.

cmoesel:
Can this PR be closed since it was replaced by #412?

PR #412, opened 2019-04-15
V14 generics

brynrhodes:
This replaces PR #404


cmoesel:
@brynrhodes -- while issuing the PR against v14 removed most of the `ContextInfo` and `RelationshipInfo` from the diff, I noticed there are still some changes in that area.  For example:

image

image

image

Is this intentional or did you accidentally merge in some local changes not yet in v14?

brynrhodes:
That was intentional, those were two issues that I found and addressed while regression testing the generics changes. In theory, I could push just those in separate commit to the base v14 branch, but it felt like more overhead than it's worth since this is a draft branch already.

cmoesel:
OK.  I don't want to cause too much trouble -- but since those changes are only bundled into the "merge" commit, it's a little confusing -- as it turns out that commit is now more than a merge.

If it's easy to somehow pull those out into a separate commit (even on this branch as part of this PR), that's preferable -- but if it's not easy, that's fine.  I'm probably too much of a Git idealist. ;-)

brynrhodes:
@cmoesel , okay, that's fixed so that this PR is all and only the generics work, and passes regression.

cmoesel:
@brynrhodes -- I apologize.  I am extremely annoying.  But... Now the commit history has a bunch of duplicates.  I think something went awry with the merging/rebasing.  Did you notice that?

If we use the "Squash and Merge" option it's a moot point.  If we do a traditional merge, it's a bit messy...  what do you think?

Also, are you looking for my emotional approval on this one or are you asking me to download, test, etc.  So far I'm just taking yours and Claude's word that it works and doesn't cause any regressions.

brynrhodes:
@cmoesel, yes, I think that happens because I rebase locally but then it automatically merges in the remote when I push. So I end up rebasing twice and that happens in the history. I'm happy to use the squash and merge option.

And yes, just emotional support. Want to make sure you're good with the direction, no need to do anything further than that, it's a draft branch at this point anyway.

cmoesel:
> @cmoesel , okay, that's fixed so that this PR is all and only the generics work, and passes regression.

So... are you sure that went right?  I'm still seeing the same non-generics-related changes that I screenshotted in my first comment.  See: https://github.com/cqframework/clinical_quality_language/pull/412/files

That said, given its a draft branch, and you're going to squash-merge, I'm not too concerned.  If you don't want to fix it just add something about those changes to the commit message when you squash it.

You have my emotional approval. ;-)

brynrhodes:
Yeah, I have no idea why it thinks that's a difference between the v14, if you look at the v14, that change has been made:
https://github.com/cqframework/clinical_quality_language/blob/v14/Src/cql-lm/schema/model/modelinfo.xsd#L44

So... I think it went right, yes, but git often confuses me :)

At any rate, I'm going to go ahead and merge, thanks @cmoesel !

PR #427, opened 2019-06-07
Made multi-version libraries and models possible

JPercival:


cmoesel:
It's hard to tell based on the description, but will this PR fix #438 as well?  Or is it a different thing?

JPercival:
Bryn and I have been discussing this a bit offline and we don't feel this is the correct approach.

Issue #420, opened 2019-05-03
Incorrect resolution to in(Any, Interval<Any>)


brynrhodes:
With the 1.3.15 release, the following library:

    library Scratchpad2
    using FHIR version '3.0.0'
    include FHIRHelpers version '3.0.0' called FHIRHelpers
    context Patient
    
    define function "Qualifying Encounters"(MeasurementPeriod Interval<DateTime>):
      [Encounter] ValidEncounter
        where ValidEncounter.period during day of MeasurementPeriod

Resolves the "during" as an "in" (choosing the DateTime as opposed to the Period). This is the result of the fact that during generic instantiation, the translator will look for the maximal common supertype, which ends up being System.Any, and that parameterizes any generic signature.

brynrhodes:
Included in 1.3.16-SNAPSHOT and 1.4.4-SNAPSHOT

PR #429, opened 2019-06-07
#420: Disabled Any,Any resolution

brynrhodes:


brynrhodes:
This is failing because it has a test that requires the other PR to pass, unrelated.

Issue #451, opened 2019-09-10
CQL-to-ELM Translator - Signature annotations missing from included libraries


sanumula:
When Signatures are enabled, only the referencing library when translated to
ELM has the signature annotations whereas the included libraries don't. I see that the translator is not passing the signature information while processing the included library at [1] and hence it is defaulting to SignatureLevel.None

[1] https://github.com/cqframework/clinical_quality_language/blob/master/Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/LibraryManager.java#L95

**Example:**
You can reproduce the issue by executing the translator on the top-level cql library i.e., ReferencingLib.cql shown below.

**RootLibrary.cql:**

```
library RootLibrary

define RootLibrarySum: Sum({1,2,3})
```

**BaseLib.cql:**

```
library BaseLib

include RootLibrary called Root

define BaseLibSum: Sum({10,20,30})
```

**ReferencingLib.cql:**

```
library ReferencingLib

include BaseLib called Base

define ReferencingLibSum: Sum({100,200,300})
```

The ELM generated from the above CQL is attached to this ticket.
[cql_elm_without_signature.txt](https://github.com/cqframework/clinical_quality_language/files/3593084/cql_elm_without_signature.txt)

Is this an issue with the translator or am I missing something?

sanumula:
I submitted a [PR](https://github.com/cqframework/clinical_quality_language/pull/452) that addresses this issue.

sanumula:
@brynrhodes Thanks for the review and merging the PR. Could you please let us know when will be the next CQFramework release?

brynrhodes:
Addressed in 1.3.19-SNAPSHOT and 1.4.8-SNAPSHOT

PR #452, opened 2019-09-17
#451 CQL-to-ELM Translator - Signature annotations missing from included libraries

sanumula:
When Signatures are enabled, only the referencing library when translated to
ELM has the signature annotations whereas the included libraries don't. This PR addresses the issue by passing the signature information to the translator while translating the included library to elm.

Example to reproduce the issue:

**ReferencingLibrary.cql**
```
library ReferencingLib

include BaseLibrary called Base
```

**BaseLibrary.cql**
```
library BaseLibrary

define BaseLibSum: Sum({1,2,3})
```

When the ReferencingLibrary.cql is translated to elm programmatically using the CqlTranslator, the elm corresponding to BaseLibrary does not have the signatures.


sanumula:
Thanks for the review!

PR #456, opened 2019-10-08
Update Travis requirements

JPercival:


cmoesel:
Two comments:

1. Is it possible to run this using OpenJDK instead of OracleJDK?  Since Oracle started charging for OracleJDK, we should probably prefer OpenJDK if at all possible.
2. I noticed the addition of a lot of `com.sun.*` packages as dependencies.  Were these things that used to be in OracleJDK8 but were removed in OracleJDK11?  Do these dependencies work w/ OpenJDK?

JPercival:
> Two comments:
> 
>     1. Is it possible to run this using OpenJDK instead of OracleJDK?  Since Oracle started charging for OracleJDK, we should probably prefer OpenJDK if at all possible.

I''m using OpenJDK and OpenJ9 on my local development environment so yes, it's technically possible. I'll check to see if TravisCI offers a OpenJDK environment for builds.

>     2. I noticed the addition of a lot of `com.sun.*` packages as dependencies.  Were these things that used to be in OracleJDK8 but were removed in OracleJDK11?  Do these dependencies work w/ OpenJDK?

The jaxb components were deprecated in Java 9 as part of the overall JDK modularization efforts and removed in Java 11. Those dependencies do work with OpenJDK.

PR #463, opened 2019-12-05
Add ability to translate directly from VersionedIdentifier

JPercival:
* Allow translation from VersionedIdentifier


JPercival:
Closing because out of date. Also, one of the issues fixed by this is addressed via the translation options push down.

PR #466, opened 2020-01-06
Fix issue #436 - Translator Options flow down - CQL 1.3

### LINKED ISSUE #436: Translator options should flow down
When the library manager creates a translator, it uses default options. Library translation options should flow down.

JPercival:
* Resolves #436 
* Includes the same fixes as #465 , but for CQL 1.3. I also had to pull in a few other changes from the 1.4 branch in order to get it to work.
* Upgrade to gradle 5.6.2
* Upgrade to jacoco 0.8.4
* Upgrade javax dependencies

### DIFF
--- a/Src/java/build.gradle
+++ b/Src/java/build.gradle
@@ -30,7 +30,7 @@ subprojects {
     }
 
     jacoco {
-        toolVersion = "0.7.4.201502262128"
+        toolVersion = "0.8.4"
     }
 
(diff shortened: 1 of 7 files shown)


JPercival:
The task pinned the gradle wrapper to version 4.1 which doesn't support Java 11+. We migrated to a new version of Java. That change had already been made in the 1.4 branch which is where I pulled it from.

PR #467, opened 2020-01-07
Fix 278 quick 20 support

brynrhodes:
Adds support for QUICK 3.3.0 and updates the FHIR 4.0.0 model info with the result of StructureDefinitionToModelInfo tooling in the cqf-tooling project.


Src/java/cql-to-elm/src/test/resources/org/cqframework/cql/cql2elm/fhir/v18/fhir-modelinfo-1.8.xml
```diff
@@ -7734,7 +7734,7 @@
         <ns4:element name="value" type="FHIR.string"/>
     </ns4:typeInfo>
     <ns4:conversionInfo fromType="dateTime" toType="System.DateTime" functionName="FHIRHelpers.ToDateTime"/>
-    <ns4:conversionInfo fromType="date" toType="System.DateTime" functionName="FHIRHelpers.ToDateTime"/>
+    <ns4:conversionInfo fromType="date" toType="System.DateTime" functionName="FHIRHelpers.ToDate"/>
```
cmoesel: It seems this change must be intentional, but I'm not sure I understand it.

FHIRHelpers (for FHIR 4.0.0) shows that `FHIRHelpers.ToDate(value date)` returns `value.value` (i.e., the `date`'s `value`:
```
 define function ToDate(value date): value.value 
```
And the model-info (for FHIR 4.0.0) shows that a `date`'s `value` is a `System.Date`:
```xml
    <ns4:typeInfo xsi:type="ns4:ClassInfo" namespace="FHIR" name="date" identifier="http://hl7.org/fhir/StructureDefinition/date" label="date" retrievable="false" baseType="FHIR.Element">
        <ns4:element name="value" elementType="System.Date"/>
    </ns4:typeInfo>
```

So, in the `conversionInfo` above, the `toType` is intended to be a `System.DateTime`, but the specified function returns a `System.Date`.  I checked the system types model-info, and `Date` does not have any formal relationship to `DateTime` in the type hierarchy, so it's not a matter of returning a sub-type.

brynrhodes:
I went back to the 4.0.0 FHIR model here exactly as they are in the current branch (there's one fix that I added based on a new test case), so this PR only includes QUICK 3.3.0 support. To get to FHIR 4.0.1 and QUICK 4.0.0, we're reving the cqf-tooling, so I'd like to consider this as a separate PR and I'll submit another one to get QUICK 4.0.0.

Src/java/cql-to-elm/src/test/resources/org/cqframework/cql/cql2elm/fhir/v18/fhir-modelinfo-1.8.xml
cmoesel: Nevermind.  It appears that this change is on the FHIR `1.8` model, which I must have somehow missed (as my comment seems to indicate I thought it was R4).  Sorry!

PR #470, opened 2020-01-22
Add logic for choice super and sub types

JPercival:
* Added logic and tests for testing the compatibility of Choice types

Same as #469, but for 1.4


JPercival:
Updated to include #469

Issue #450, opened 2019-08-28
CQL-to-ELM Translator 1.4.6 - Internal Translator Error


JSRankins:
When provided the attached libraries, the CQL-to-ELM Translator results in, among other errors, an internal translator error that has no locator:

Error:[n/a] Internal translator error.

This makes it difficult for a tool using the Translator to provide users feedback on where the applicable error occurs. 

Is this an error in the translator, or was it intended by design?

Note: We had to attach the files as .txt files, since github does not support the attachment of .cql files.

[NCQA_Common-5.0.000.txt](https://github.com/cqframework/clinical_quality_language/files/3552721/NCQA_Common-5.0.000.txt)
[DMS_HEDIS_2020-3.0.002.txt](https://github.com/cqframework/clinical_quality_language/files/3552722/DMS_HEDIS_2020-3.0.002.txt)

brynrhodes:
Thank you for the reproduction, I was able to easily reproduce the problem.

The root cause is that the library being included is declared to use QDM version 5.4, but the main library is declared to use QDM version 5.5. When the included library is translated, it errors (correctly) when trying to load QDM 5.4 because 5.5 is already in use. However, because the error is encountered while loading the data model, the Patient context creation logic doesn't complete, leaving a partially constructed Patient expression. This partial Patient context results in an internal NullPointerException when the Patient context is referenced later in the included library. An internal error like that would normally be tagged with the location in the source code causing the error, but because the Patient expression isn't complete, the translator doesn't have a link to the parse tree to obtain that information.

I have a fix ready to apply, and will discuss with the impacted teams the best approach to getting the fix applied to the repository.

brynrhodes:
Addressed in 1.3.19-SNAPSHOT and 1.4.8-SNAPSHOT

PR #471, opened 2020-01-23
#450: Fixes an internal translator error that occurs when model info …

brynrhodes:
…issues prevent loading the context.


JPercival:
Looks good. Builds locally

PR #485, opened 2020-03-11
Fix 461 support fhir 401

brynrhodes:
Added support for the FHIR 4.0.1 technical correction.
Added support for USCore version 3.1.0
Added support for QICore version 4.0.0


brynrhodes:
Hi @cmoesel , thank you for the review, I didn't notice the discrepancy between race and ethnicity, but yes, these are fully generated ModelInfo files from the source StructureDefinitions, so there must be some difference in the way those extensions for represented in the SDs, even though they are structurally the same. I'll take a look at that.

And yes, the intent is for the ELM produced by the translator using these ModelInfos would run against a straight FHIR data access layer, so there needs to be enough information in the "target" element of the extension and slice representations to allow the engine to get back to that. I'll look in to why those targets aren't coming out properly. (They're all coming out as `value` right now, which only works for the primitives).

I do wonder whether we should introduce an option in the translator to allow it to ignore target paths in elements? So that if a data access layer did understand the ModelInfo directly it wouldn't need the target paths?

And for reference, there is a branch of the [cqf-tooling](https://github.com/cqframework/cqf-tooling/tree/fix-45-update-sd-to-modelinfo) repository where I'm making the changes that support these, so using that branch, with the published definitions from FHIR and the USCore/QICore IGs, these ModelInfos are directly generated.

Which is the answer to your first question, the difference between the FHIR 4.0.0 and the FHIR 4.0.1 model info. The 4.0.0 ModelInfo used the old tooling chain (going through XSD-to-ModelInfo), where the 4.0.1 ModelInfo is generated with the new tooling chain, straight from the FHIR StructureDefinitions. (And I started with the technical correction 4.0.1 SDs).

cmoesel:
> I do wonder whether we should introduce an option in the translator to allow it to ignore target paths in elements? So that if a data access layer did understand the ModelInfo directly it wouldn't need the target paths?

I personally wouldn't use it, but if it doesn't add much complexity, I suppose it is possibly that another system might prefer it that way (maybe if they are actually mapping the model to an underlying non-fhir database).

> The 4.0.0 ModelInfo used the old tooling chain (going through XSD-to-ModelInfo), where the 4.0.1 ModelInfo is generated with the new tooling chain, straight from the FHIR StructureDefinitions.

That _partially_ answers my question, but I guess what I really want to know is what is actually different in the _model_ (not what is different in _how the model was created_).  E.g, did it result in any resources being represented in a different way, datatypes being changed, etc?  If I built a data source provider that works against the 4.0.0 model, should I expect it to work with CQL built against the 4.0.1 model?  Etc.

brynrhodes:
I don't have a good answer for this right now. In theory, they should be identical (modulo changes between the 4.0.0 and 4.0.1 structures on which they are based, but from the structural perspective there shouldn't be any of those either). Know of any semantic comparison tools for instances of XSDs?

cmoesel:
> Know of any semantic comparison tools for instances of XSDs?

Ha.  Nope.

brynrhodes:
Note that going forward, the generated ModelInfos output with a stable sort, so this should be a much easier question to answer. We're looking at getting the 4.0.0 sorted in that same way so we can answer this question.

cmoesel:
FYI -- I'm waiting on a more definitive answer re: the `race` vs `ethnicity` extensions before I remove the *Changes requested*.  If you think this is something better addressed (or explained) in a follow-on, let me know.

brynrhodes:
Hi @cmoesel , the difference between the Race and Ethnicity extensions turned out to be pretty subtle and led to a whole host of other fixes and enhancements to the tooling. Even though they have the same elements, they are represented completely differently in the slicing. 

But the model info generation tooling now handles both, as well as provides full mapping information for each slice/extension/primitive in terms of the underlying FHIR model. The translator uses that mapping to transform the ELM. That last bit isn't complete, but it works for primitives and choices, and passes the map through for slices and extensions.

It's next on the list to have the translator deal with the slice and extension mapping as well, but I'm out of time right now, so I'm planning to do that as a followup item rather than tie up this PR.

brynrhodes:
@cmoesel , any chance you can approve this one? I'd like to get it merged before I start the branch for STU5 support.

cmoesel:
Sorry for the delay, @brynrhodes.  I didn't go through the Java code changes in detail (although I skimmed them) -- but here is what I did do:

* I verified that the questions/concerns I raised earlier were addressed in the latest commit (and they were, so thanks!).
* I ran regression on a bunch of artifacts I have locally and while there were changes to the outputted ELM, none of them were substantial or incorrect.  So that's good!
* I changed one of my existing FHIR 4.0.0 artifacts to use the FHIR 4.0.1 model and then compared the resulting ELM files.  A few things I noticed:
    * The `templateId` field was added into the retrieves, and it reflects the FHIR canonical URL.  Fine by me!
    * The `codeProperty` for `MedicationRequest` and `MedicationStatement` was changed from `medication` to `medicationCodeableConcept`.  I have some concerns about this one.
    * As a result of the previous change (I think), I now have a slew of these warnings in my ELM: `Could not resolve code path medicationCodeableConcept for the type of the retrieve FHIR.MedicationRequest.`

Was the change from `medication` to `medicationCodeableConcept` intentional?  I think it's a bit strange to put in a default that causes warnings for anyone who uses it and that references a property that's not even formally in the model.

brynrhodes:
@cmoesel , thank you!

The reason the primaryCodePath is set to medicationCodeableConcept is that the medication element is actually a choice type, and the actual search we want to select is the CodeableConcept variant there. But you're right, it doesn't actually correspond to any element of the resource, nor does it actually correspond to the name of a search parameter.

So I've switched it back to "medication", but note that the `medication` search parameter for all those resource types is actually the `reference` search parameter; `code` is the name of the search parameter that searches by `medication as CodeableConcept`. However, until we have a decent way of saying this in the model info (coming soon), this is something that has to be dealt with in the data access layer.

Issue #345, opened 2018-08-17
Add XML comment in ELM XML with version of CQL-to-ELM translator used


p9g:
The ELM XML generated by the CQL-to-ELM translator should include the version of the CQL-to-ELM translator which was used to generate it.

brynrhodes:
Addressed in 1.4.9-SNAPSHOT

brynrhodes:
Addressed in 1.4.9 release

PR #489, opened 2020-03-21
#345: Added version and translator option information to the ELM output.

brynrhodes:


cmoesel:
I don't have time to run the code (need to go grab lunch), but a few thoughts...
* The test doesn't actually have any options. To prove this works as intended, it seems we should have at least one test that includes several options.
* It looks like the options are just appended in a string.  Did you consider serializing them as a structured object so it's easier to parse and find specific option values? Or were you trying to avoid needing to update the schema?

brynrhodes:
Yeah, it doesn't actually include the version either, which I'm not sure I understand why that's happening, I set the version, it just isn't set in the package info that comes back from the class, so I'm sort of hoping (fingers crossed) that it'll just work once the actual release and snapshot builds are in place.

And yes, I was hoping to avoid doing anything fully structural, it's really just informational at this point. Not because I don't think it would be useful, but I think this minimal would be more useful than nothing, and it's all I have time for right now.

Totally get if we should hold off on this one for a bit, was just trying to get something quick in place since the CqlToElmInfo was already defined, just not being used.

cmoesel:
If you can resolve the conflicts (caused by recent merge) and create at least one test that shows options, I'm comfortable merging it.

brynrhodes:
Done, thanks Chris!

cmoesel:
Thanks, @brynrhodes.  It looks like there is a failing test now.  Travis doesn't show you the details, but I ran it locally and here's the failed test: 
image

brynrhodes:
Woops, that's what I get for not running the regression :(

Fixed.

Issue #505, opened 2020-04-14
FHIRHelpers doesn't show up in the correct namespace


brynrhodes:
The FHIRHelpers library should be implicitly defined in the FHIR namespace, not the current namespace: FHIR: http://hl7.org/fhir

brynrhodes:
Addressed in 1.4.9 release

PR #506, opened 2020-04-14
#505: Added support for namespace aware providers to participate in t…

brynrhodes:
…he namespace selection process for a referenced library

### DIFF
--- a/Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/FhirLibrarySourceProvider.java
+++ b/Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/FhirLibrarySourceProvider.java
@@ -3,14 +3,33 @@ package org.cqframework.cql.cql2elm;
 import org.hl7.elm.r1.VersionedIdentifier;
 
 import java.io.InputStream;
+import java.util.Collections;
 
 /**
  * Created by Bryn on 3/28/2017.
  */
-public class FhirLibrarySourceProvider implements LibrarySourceProvider {
+public class FhirLibrarySourceProvider implements LibrarySourceProvider, NamespaceAware {
+
+    private final String namespaceName = "FHIR";
+    private final String namespaceUri = "http://hl7.org/fhir";
+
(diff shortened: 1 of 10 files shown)


Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/FhirLibrarySourceProvider.java
```diff
@@ -3,14 +3,33 @@
...
 import java.io.InputStream;
+import java.util.Collections;
 
 /**
  * Created by Bryn on 3/28/2017.
  */
-public class FhirLibrarySourceProvider implements LibrarySourceProvider {
+public class FhirLibrarySourceProvider implements LibrarySourceProvider, NamespaceAware {
+
+    private final String namespaceName = "FHIR";
```
JPercival, line 13: So, is the intent that all the LibrarySouceProviders will assign the namespaces to the Libraries? What does this look like in the context of loading a library from a FHIR server?

brynrhodes: The FHIRLibrarySourceProvider only knows how to load the FHIRHelpers library from the translator package resources, so it always has the namespace FHIR{http://hl7.org/fhir}. So the intent is to ensure that whenever this library is loaded from the translator it presents a consistent namespace, while at the same time support "namespace-less" usage as it works right now. So if you don't provide namespaces, it works like it always did, but if you use the translator in a namespace-aware context, then this provider will explicitly set the namespace of the FHIRHelpers library (otherwise it gets put in the default namespace for the context).

brynrhodes: So, no, in general, the LibrarySourceProviders won't assign namespaces (though they might use namespacing information to support finding the appropriate source). Only the FHIRLibrarySourceProvider needs this ability to "override" the default namespace.

Issue #513, opened 2020-04-18
Json Serialization with Jaxb produces inconsistent output


brynrhodes:
The JAXB Json output is inconsistent in it's use of the `type` element, which causes issues attempting to deserialize the Json using a Jackson deserializer.

brynrhodes:
Addressed in 1.4.9 release

PR #518, opened 2020-04-21
#513: Added a JXSON format option to support using the Jackson serial…

brynrhodes:
…izer to produce JSON output.

We've recently tried deserializing JSON ELM in the Java library and the output from the JAXB-JSON serialization is inconsistent (i.e. it produces type elements only in the case that the type cannot be inferred from the element definition). Deserializing using JAXB is not possible for certain environments (specifically Android phones just don't work with it). We have been able to successfully deserialize JSON in Java on Android, but only with the Jackson deserializer, but the Jackson deserializer doesn't work with the JAXB-JSON serialized output. So we're proposing a new JXSON format option that instructs the translator to serialize to JSON with the Jackson serializer, resulting in consistent, and we believe backwards-compatible, JSON output.

However, we're not 100% confident in that backwards-compatible output, and that's the reason for introducing it as a separate option. So the JSON option works the same as it always did, using the JAXB-JSON serializer, where the JXSON option produces JSON, but using the Jackson serializer. We've added a suite of regression tests to the Java engine project, but it only verifies that the content can be deserialized without errors, not that the resulting content is semantically equivalent.

So, this PR is a request to add JXSON as an option to support efforts to use CQL in an Android device, but without impacting the current capabilities, and providing a trajectory to align on JXSON support moving forward (since JAXB support seems to be a dying thing).

### DIFF
--- a/Src/java/build.gradle
+++ b/Src/java/build.gradle
@@ -116,6 +116,10 @@ configure(subprojects.findAll {it.name in ['model', 'elm', 'quick', 'qdm']}) {
         runtime group: 'com.sun.xml.bind', name: 'jaxb-impl', version: '2.4.0-b180830.0438'
         runtime group: 'javax.xml.bind', name: 'jaxb-api', version: '2.3.1'
         runtime group: 'com.sun.activation', name: 'javax.activation', version: '1.2.0'
+        runtime group: 'com.fasterxml.jackson.core', name: 'jackson-databind', version: '2.10.1'
+        runtime group: 'com.fasterxml.jackson.core', name: 'jackson-annotations', version: '2.10.1'
+        runtime group: 'com.fasterxml.jackson.core', name: 'jackson-core', version: '2.10.1'
+        runtime group: 'com.fasterxml.jackson.module', name: 'jackson-module-jaxb-annotations', version: '2.10.1'
     }
 
     ext.xjc = [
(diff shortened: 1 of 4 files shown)


cmoesel:
Thanks for the heads up.  I'm pleasantly surprised to see how little needed to change in the codebase in order to support Jackson.

I tested this out using the legacy JSON mode on a bunch of MITRE-developed artifacts and confirmed that there was no change at all in the output when using `JSON` format.  So that's good!

I also tested it with `JXSON` format and saw that there are definitely some differences from `JSON` format that would need to be investigated further.  Two that popped out immediately:
* JXSON serialization seems to include the `type` property in a lot more of the objects than JSON serialization does
* JXSON serialization has top-level library property `"annotation" : [ { } ]`, whereas JSON serialization has `"annotation" : [ { "translatorOptions" : "", "type" : "CqlToElmInfo" } ]`

So I think I'm fine w/ this PR as-is (as long as that annotation issue noted above isn't a show-stopper), but agree that more work is needed to determine if the two outputs are really semantically equivalent and compatible.

brynrhodes:
I've created an issue to track regression of JXSON: #522 

Yes, the JXSON serializer consistently produces `type` elements, and that's actually the key to getting it to work. In the Jackson deserializer, if the type element isn't consistently present, deserialization will error out.

And yes, there is some more work to be done on the annotation serialization, I've noted #523 for that, but the target for JXSON at this point is a minimal library resource that wouldn't be expected to include annotations, so the use case isn't impacted by that issue.

PR #528, opened 2020-05-27
Deduplicate Jaxb depedencies, move to smaller antlr dependency, fix a…

JPercival:
* Update to use smaller antlr-runtime dependency
* Update to latest jaxb dependencies and consolidate those
  * NOTE: The translator is currently using eclipsemoxy specific functionality, so it's the jaxb implementation I selected
* Fix a few warnings


cmoesel:
Does this affect the output of CQL-to-ELM or is it a side-effect-free refactoring?

JPercival:
> Does this affect the output of CQL-to-ELM or is it a side-effect-free refactoring?

There's not any change intended to any output format (XML, JXSON, JAXB). The main issue was that multiple implementations of JAXB we're being picked up on the classpath. That causes warnings on later versions of the JDK.

The ANTLR change also means that the total size of the translator and associated dependencies is a bit smaller.

Issue #526, opened 2020-05-15
Translator reports syntax errors with library "unknown"


brynrhodes:
The translator is reporting syntax errors with a library of Unknown:
https://github.com/cqframework/clinical_quality_language/blob/25e12ed6d60b40d0b4c772a05219dfd562ef5253/Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/CqlTranslator.java#L475

This makes it impossible to correctly associate the syntax error with a library.

brynrhodes:
Addressed in 1.4.9 release

PR #533, opened 2020-06-11
#526: Fixed syntax errors reporting with library "unknown", rather th…

brynrhodes:
…an the source library context. Syntax errors will now be reported with "Anonymous" if there is no source library, or with the name of the source file if it can be inferred from the translator construction. Users of the translator can also provide sourceinfo if they know it.

### DIFF
--- a/Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/CqlTranslator.java
+++ b/Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/CqlTranslator.java
@@ -523,21 +523,14 @@ public class CqlTranslator {
 
         private VersionedIdentifier extractLibraryIdentifier(cqlParser parser) {
             RuleContext context = parser.getContext();
-            while (!(context instanceof cqlParser.LibraryContext)) {
+            while (context != null && !(context instanceof cqlParser.LibraryContext)) {
                 context = context.parent;
-                if (context == null) {
-                    break;
-                }
             }
 
             if (context instanceof cqlParser.LibraryContext) {
-                for (ParseTree pt : ((cqlParser.LibraryContext)context).children) {
-                    if (pt instanceof cqlParser.LibraryDefinitionContext) {
-                        cqlParser.LibraryDefinitionContext ldc = (cqlParser.LibraryDefinitionContext)pt;
-                        if (ldc.qualifiedIdentifier() != null && ldc.qualifiedIdentifier().identifier() != null) {
-                            return new VersionedIdentifier().withId(StringEscapeUtils.unescapeCql(ldc.qualifiedIdentifier().identifier().getText()));
-                        }
-                    }
+                cqlParser.LibraryDefinitionContext ldc = ((cqlParser.LibraryContext)context).libraryDefinition();
+                if (ldc != null && ldc.qualifiedIdentifier() != null && ldc.qualifiedIdentifier().identifier() != null) {
+                    return new VersionedIdentifier().withId(StringEscapeUtils.unescapeCql(ldc.qualifiedIdentifier().identifier().getText()));
                 }
             }
 
--- a/Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/CqlTranslator.java
+++ b/Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/CqlTranslator.java
@@ -61,6 +61,7 @@ public class CqlTranslator {
     private List<CqlTranslatorException> errors = null;
     private List<CqlTranslatorException> warnings = null;
     private List<CqlTranslatorException> messages = null;
+    private VersionedIdentifier sourceInfo = null;
     private NamespaceInfo namespaceInfo = null;
     private ModelManager modelManager = null;
     private LibraryManager libraryManager = null;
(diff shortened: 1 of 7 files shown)


Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/CqlTranslator.java
```diff
@@ -328,12 +349,22 @@ private CqlTranslator(NamespaceInfo namespaceInfo, ANTLRInputStream is, ModelMan
...
+                          LibraryManager libraryManager, UcumService ucumService, CqlTranslatorOptions options) {
+        this.sourceInfo = sourceInfo;
         this.namespaceInfo = namespaceInfo;
         this.modelManager = modelManager;
         this.libraryManager = libraryManager;
         this.ucumService = ucumService;
         this.errorLevel = options.getErrorLevel();
 
+        if (this.sourceInfo == null) {
+            sourceInfo = new VersionedIdentifier().withId("Anonymous").withSystem("text/cql");
```
JPercival, line 365: If you were to translate a library from an input stream defined as:

library ExpressionLibrary
define Test: 2+2
define -syntax error!

You'd end up with a source info of "Anonymous" which is counter-intuitive. Currently, the code looks up the filename of the CQL source if one is available. It's not generally possible make sure there's always a library identifier present (as the SyntaxErrorWithNoLibrary shows). But would it be more user friendly to attempt to parse the library name from the InputStream for cases where that's possible?

brynrhodes: Good point, and I thought I had that covered, but it doesn't work like I expected, so I added some tests and some code to pull it from what's parsed so far if it's available.

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/CqlTranslator.java
```diff
@@ -470,9 +521,42 @@ public CqlErrorListener(LibraryBuilder builder, boolean detailedErrors) {
...
         }
 
+        private VersionedIdentifier extractLibraryIdentifier(cqlParser parser) {
+            RuleContext context = parser.getContext();
+            while (!(context instanceof cqlParser.LibraryContext)) {
+                context = context.parent;
+                if (context == null) {
+                    break;
+                }
+            }
```
cmoesel, line 531: This suggestion is so trivial that I hesitate to make it, so feel free to ignore it if you wish, but... I personally prefer putting nullchecks into the condition versus a dedicated `break`.  E.g.,
```java
            while (context != null && !(context instanceof cqlParser.LibraryContext)) {
                context = context.parent;
            }
```

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/CqlTranslator.java
```diff
@@ -470,9 +521,42 @@ public CqlErrorListener(LibraryBuilder builder, boolean detailedErrors) {
...
+
+            if (context instanceof cqlParser.LibraryContext) {
```
cmoesel, line 533: Due to the while loop above, at this point `context` is either `null` or it is a LibraryContext, so this could just be `if (context != null)`.  Again, maybe not necessary...  but I don't remember how Java handles `null instanceof`.

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/CqlTranslator.java
```diff
@@ -470,9 +521,42 @@ public CqlErrorListener(LibraryBuilder builder, boolean detailedErrors) {
...
+                for (ParseTree pt : ((cqlParser.LibraryContext)context).children) {
+                    if (pt instanceof cqlParser.LibraryDefinitionContext) {
+                        cqlParser.LibraryDefinitionContext ldc = (cqlParser.LibraryDefinitionContext)pt;
+                        if (ldc.qualifiedIdentifier() != null && ldc.qualifiedIdentifier().identifier() != null) {
+                            return new VersionedIdentifier().withId(StringEscapeUtils.unescapeCql(ldc.qualifiedIdentifier().identifier().getText()));
+                        }
+                    }
+                }
```
cmoesel, line 541: I don't know why I'm being so nit-picky about this section of code, but doesn't Java have a way to find an item in a list based on criteria (as opposed to a big loop)?  I haven't ever used it, but based on Google, I *think* you could do something like:
```java
cqlParser.LibraryDefinitionContext ldc = ((cqlParser.LibraryContext)context).children
    .stream().filter(pt -> pt instanceof cqlParser.LibraryDefinitionContext).findAny();
if (ldc.qualifiedIdentifier() != null && ldc.qualifiedIdentifier().identifier() != null) {
    return new VersionedIdentifier().withId(StringEscapeUtils.unescapeCql(ldc.qualifiedIdentifier().identifier().getText()));
}
```
I don't know.  You might need to add a cast.  Maybe it's not any better.  I just don't like using `for` loops to search for things. ;-)

cmoesel: I'm also assuming a `LibraryContext` can only have one `LibraryDefinitionContext`, right?  If so, and if you keep the `for` loop, then you could add a `return null` or a `break` after line 538 above because there's no sense in continuing to loop over the child contexts if you already found the library definition.

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/CqlTranslator.java
brynrhodes: Agreed and applied

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/CqlTranslator.java
brynrhodes: While this is true, I prefer to be explicit about the fact that I expect context to be a LibraryContext.

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/CqlTranslator.java
brynrhodes: Agreed, and you're right, there is only one libraryDefinitionContext, so I don't need the looping at all. Fixed, thank you for the feedback!

cmoesel: Ha.  Yes, even better!

Issue #570, opened 2020-09-25
ModelInfo targetQualifier should be string, not QName


brynrhodes:
The targetQualifier attribute of the ModelInfo type is currently declared as a QName, but from the perspective of the ModelInfo, it's actually just a string. The JAXB implementation that was used initially didn't have any issues with writing an unresolveable QName as part of an XML document, but as part of harmonizing dependencies with other projects, the JAXB dependency was changed to use the eclipse Moxy implementation. That implementation is fine with reading unresolveable QName references (apparently, because the ModelInfo reading worked just fine) but it is not able to write an unresolveable QName, which didn't become apparent until a release of the translator that used the Moxy implementation was incorporated into CQF Tooling, where the ModelInfo writing occurs. In that code, an obscure NullReferenceException occurs when trying to write the ModelInfo, and it turns out to be because of the unresolveable QName. 

The simplest fix is to correct the type of the targetQualifier attribute, which addresses the issue.

brynrhodes:
Released in 1.5.0

PR #571, opened 2020-09-25
#570: Fixed targetQualifier incorrectly declared as a QName

brynrhodes:
Although this is not, strictly speaking, a backwards compatible change, at this point the only actual usages of ModelInfo I am aware of are in the translator itself. Looking for confirmation of this as part of this review.


JPercival (approved):

cmoesel:
The [cql-exec-fhir](https://github.com/cqframework/cql-exec-fhir) data model provider for cql-execution uses modelinfo.  It reads modelinfo to dynamically determine how to access and return FHIR data in the context of CQL expressions.

That said, we don't use `targetQualifier`, so this has no effect on us.

cmoesel (approved):

Issue #557, opened 2020-08-18
FHIRHelpers ToQuantity incorrectly using unit


brynrhodes:
The FHIRHelpers ToQuantity conversion code uses the "unit" element of the Quantity type in FHIR, it should be using the "code" element, and should only work if system is either unspecified or equal to UCUM (http://unitsofmeasure.org)

brynrhodes:
Here's what it really should be doing:

```
define function ToQuantity(quantity FHIR.SimpleQuantity):
  case
    when quantity is null then null
    when quantity.value is null then null
    when quantity.system is null or quantity.system = 'http://unitsofmeasure.org' then
      System.Quantity {
        value: quantity.value.value,
        unit: quantity.code.value
      }
    else
      Message(null, true, 'MMECalculator.ToQuantity.InvalidFHIRQuantity', ErrorLevel, 'Invalid FHIR Quantity code: ' & quantity.code.value)
  end
```

brynrhodes:
Included in release 1.5.1

PR #573, opened 2020-10-10
#557: Fixed FHIRHelpers logic for ToQuantity and ToRatio

brynrhodes:
Note there is now a ModelInfoComparer and I've added a test for it that compares the previous 4.0.1 model info with this new one to make it easier to see what has changed (or rather, what hasn't changed in this case).


JPercival (approved):

Src/java/quick/src/main/resources/org/hl7/fhir/FHIRHelpers-4.0.1.cql
```diff
@@ -11,16 +11,70 @@ define function ToInterval(period FHIR.Period):
...
+            Message(null, true, 'FHIRHelpers.ToQuantity.ComparatorQuantityNotSupported', 'Error', 'FHIR Quantity value has a comparator and cannot be converted to a System.Quantity value.')
+        when quantity.system is null or quantity.system.value = 'http://unitsofmeasure.org'
+              or quantity.system.value = 'http://hl7.org/fhirpath/CodeSystem/calendar-units' then
+            System.Quantity { value: quantity.value.value, unit: ToCalendarUnit(Coalesce(quantity.code.value, quantity.unit.value, '1')) }
         else
-            Message(null, true, 'FHIRHelpers.ToQuantity.InvalidFHIRQuantity', 'Error', 'Invalid FHIR Quantity code: ' & quantity.code.value)
+            Message(null, true, 'FHIRHelpers.ToQuantity.InvalidFHIRQuantity', 'Error', 'Invalid FHIR Quantity code: ' & quantity.unit.value & ' (' & quantity.system.value & '|' & quantity.code.value & ')')
     end
 
+define function ToQuantityIgnoringComparator(quantity FHIR.Quantity):
```
cmoesel, line 40: Given this function, can't `ToQuantity` (above) be simplified further to:
```
define function ToQuantity(quantity FHIR.Quantity):
  if quantity.comparator is not null then
    Message(null, true, 'FHIRHelpers.ToQuantity.ComparatorQuantityNotSupported', 'Error', 'FHIR Quantity value has a comparator and cannot be converted to a System.Quantity value.')
  else
    ToQuantityIgnoringComparator(quantity)
```

brynrhodes: It definitely could be, but I was worried about introducing another call so I opted against it, given this is low-level conversion code.

brynrhodes:
@cmoesel , I rolled the ToQuantity changes through all the FHIR model infos.

cmoesel (approved):
This all looks good to me.  I temporarily updated a bunch of my CQL artifacts to use the new FHIRHelpers libraries and did not run into any issues.  I also recompiled them using this branch of cql-to-elm and that went smoothly as well.  Last I temporarily updated my FHIR 4.0.0 artifacts to use FHIR 4.0.1 and those seemed to go fine as well.  So... no concerns.  Looks good @brynrhodes and thanks for bringing this to my attention!

Issue #578, opened 2020-10-22
Support model info targets with function and a non-%value expression


msargentibm:
I was attempting to translate cql to elm using USCore. CQL Definition:
```
library "Test" version '1.0.0'
using "USCore" version '3.1.0'
parameter MeasurementPeriod default Interval[
  @1900-01-01T00:00:00.0,
  @2020-12-31T00:00:00.0
]

context Patient

define "getPedBMI":
  [PediatricBMIforAgeObservationProfile] P
    where exists(P.VSCat)
```

The translation succeeds, but the elm that is output appears to be incomplete. In uscore-modelinfo-3.1.0.xml, VSCat makes use of FHIRHelpers with an argument that contain some complex logic:
```
  <typeInfo baseType="USCore.DomainResource" namespace="USCore" name="PediatricBMIforAgeObservationProfile" identifier="http://hl7.org/fhir/us/core/StructureDefinition/pediatric-bmi-for-age" label="US Core Pediatric BMI for Age Observation Profile" retrievable="true" primaryCodePath="code" xsi:type="ClassInfo">
 ... SNIPPED ...
      <element name="VSCat" elementType="System.Concept" target="FHIRHelpers.ToConcept(%parent.category[coding.system='http://terminology.hl7.org/CodeSystem/observation-category',coding.code='vital-signs'].value)"/>
```

However, all information on how to retrieve the VSCat is dropped in the output elm. Only the FHIRHelpers reference is maintained. Output snippet:
```
      <def name="getPedBMI" context="Patient" accessLevel="Public">
         <expression xsi:type="Query">
            <source alias="P">
               <expression xmlns:ns2="http://hl7.org/fhir/us/core" dataType="ns2:PediatricBMIforAgeObservationProfile" templateId="http://hl7.org/fhir/us/core/StructureDefinition/pediatric-bmi-for-age" xsi:type="Retrieve"/>
            </source>
            <where xsi:type="Exists">
               <operand xsi:type="ToList">
                  <operand name="ToConcept" libraryName="FHIRHelpers" xsi:type="FunctionRef">
                     <operand path="VSCat" scope="P" xsi:type="Property"/>
                  </operand>
               </operand>
            </where>
         </expression>
      </def>
```

[...]

PR #580, opened 2020-10-23
#578 Support mapping expansion for arugments to a function

msargentibm:
Proposed change for https://github.com/cqframework/clinical_quality_language/issues/578.

I would also like to also add some tests if the code change seems ok. I would appreciate any suggestions folks might have as far as some of the test cases I should try to cover since I'm still getting used to the syntax used in a Model Info file.

### DIFF
--- a/Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/LibraryBuilder.java
+++ b/Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/LibraryBuilder.java
@@ -2233,7 +2233,7 @@ public class LibraryBuilder {
                 ensureLibraryIncluded(libraryName, source);
             }
             
-            String functionArgument = targetMap.substring(invocationStart + 1, targetMap.indexOf(")"));
+            String functionArgument = targetMap.substring(invocationStart + 1, targetMap.lastIndexOf(')'));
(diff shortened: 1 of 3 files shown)


brynrhodes (changes requested):
Yes, the code change makes sense, a test case would be good to validate.

msargentibm:
Thanks a lot! I'll work on adding some validation later today.

msargentibm:
I took a shot at adding in a test case to the USCore BaseTest. I tried my best to follow the general structure of the cases before it. I tried including assertions for the key components from complex expression in the model info,  but was doubting whether I had too many or too few assertions for this particular case. Please let me know if you would like to see anything altered.

Thanks again for your review.

msargentibm:
I wanted to follow up and see if there are changes I need to make in the validation case I added. Please let me know if there are any changes requested. Thanks!

msargentibm:
Wanted to check in now that the holidays have passed. Are there further updates needed for this PR?

brynrhodes (approved):
Test looks great, thank you so much!

brynrhodes:
So sorry for the delays here, the holidays really got the better of me :( Thank you so much for the contribution!

msargentibm:
Not a problem! I know end of year is an extra busy time. Thanks for the change request and reviews!

Issue #590, opened 2021-01-07
Support QDM 5.6 model info


brynrhodes:
Add support for the QDM 5.6 model info

brynrhodes:
Addressed in the 1.5.2 release

PR #591, opened 2021-01-07
#590: Added support for the QDM 5.6 model info.

brynrhodes:


JPercival:
The build failed due to Maven flakiness. I kicked it off again and it worked.

JPercival (approved):
LGTM. I spot checked a few things in the xsd against the model info and it looked good to me. How are the QDM model infos generated? I know for the FHIR ones there's an operation in the cqf-tooling.

Issue #601, opened 2021-02-18
Update Visitor Framework to latest ELM


brynrhodes:
The translator has a visitor framework for ELM that is incomplete, update this with the latest ELM definition and complete the framework to support multiple visitor-based use cases downstream (ELM-to-CQL, DataRequirements Analysis, etc).

brynrhodes:
Addressed in the 1.5.2 release

PR #603, opened 2021-02-23
Issue #601 visitor update

bryantaustin13:
#601 updated Visitor framework with the latest ELM definition. Added visit code for:
Long
AggregateClause
ContextDef


brynrhodes (changes requested):
This looks great, with the exception of the .json files for test cases. We typically don't commit those json files (they usually don't even exist, so I'm wondering why they are included, and why they are committed?). Is there a specific reason to include those files or can we remove them from the commit?

brynrhodes:
There's still one JSON committed here, ReferencingLibrary.json

brynrhodes (approved):

JPercival (approved):
LGTM.

PR #612, opened 2021-03-10
Feature shared serialization functions

JPercival:
* Made serialization functions of the translator public

The rationale for the change is that this code is duplicated frequently throughout the cql-related codebases (cql-engine, cql-evaluator, cql-ruler) where ever an ELM library needs to be serialized. Most of the other projects are not configured the same way (a specific example is handing the XML 1.0->1.1 replacement) leading to some cases of differences between the serialized ELM being generated by the cql-translator itself and the other libraries.


brynrhodes (approved):
Looks good, builds locally, merging into the feature-v15-support branch

PR #614, opened 2021-03-17
update to include more functionality to the ElmBaseVisitors

jreyno77:


brynrhodes (changes requested):
What is the internalValidation flag doing in the visitQuery method?

Src/java/elm/src/main/java/org/cqframework/cql/elm/visiting/ElmBaseLibraryVisitor.java
```diff
@@ -71,4 +99,19 @@ public T visitIncludeDef(IncludeDef elm, C context) {
     public T visitContextDef(ContextDef elm, C context) {
         return null;
     }
+
+    public T visitStatements(Statements elm, C context) {
+        int n = elm.getDef().size();
```
JPercival, line 104: List<Element> elements = elm.getDef();
for (Element e : elements)
{
 ....
}

Src/java/elm/src/main/java/org/cqframework/cql/elm/visiting/ElmBaseLibraryVisitor.java
```diff
@@ -71,4 +99,19 @@ public T visitIncludeDef(IncludeDef elm, C context) {
...
+        for (int i=0; i<n; i++) {
+            Element c = elm.getDef().get(i);
+            if (c instanceof ExpressionDef) {
+                visitElement((ExpressionDef)c, context);
+            } else if (c instanceof ContextDef) {
+                visitElement((ContextDef)c, context);
+            } else if (c instanceof FunctionDef) {
+                visitElement((FunctionDef)c, context);
```
JPercival, line 112: FunctionDef is a subclass of of ExpressionDef, so it needs to be first in the list of ifs. Otherwise this will never get hit.

JPercival (changes requested):

JPercival (approved):
Approved with the caveat that Bryn is going to fix the issues noted post-merge since he can't do so on the fork.

brynrhodes (approved):
Since this does not impact the translator at all, approving with the caveat that I will fix the issue post merge since I cannot fix it in the fork

PR #661, opened 2021-08-25
Cqf 892/match identifier approach

echicoine-esac:


Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/LibraryBuilder.java
```diff
@@ -15,14 +16,41 @@
...
 public class LibraryBuilder {
+    public static final String SYSTEM = "System";
+    public static final String INTEGER = "Integer";
+    public static final String DECIMAL = "Decimal";
+    public static final String $_THIS = "$this";
+    public static final String HIGH = "high";
+    public static final String HIGH_CLOSED = "highClosed";
+    public static final String COMPATIBILITY_LEVEL_1_5 = "1.5";
+    public static final String $_INDEX = "$index";
+    public static final String $_TOTAL = "$total";
```
brynrhodes, line 34: Although I'm not opposed to the change here, it's not part of implementing the change for match resolution and I would prefer to see it in a different PR. Can we back this change out so we're not mixing changes?

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/LibraryBuilder.java
```diff
@@ -715,6 +743,10 @@ public Element resolve(String identifier) {
         return translatedLibrary.resolve(identifier);
     }
 
+    public List<Pair<String, Object>> resolveCaseIgnored(String identifier) {
```
brynrhodes, line 746: It's not clear what List<Pair<String, Object>> is getting us here? Shouldn't we be declaring a class rather than using the generic Pair<String, Object>?

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/LibraryBuilder.java
```diff
@@ -2064,6 +2096,139 @@ else if (currentType instanceof ListType && listTraversal) {
         return null;
     }
 
+    // TODO: Support case-insensitive models
+    public List<MatchIdentifier> resolveProperties(DataType sourceType, String identifier, boolean mustResolve) {
```
brynrhodes, line 2100: As I suggested in my last review, I really think IdentifierResolution is a better name for the concept here, and more consistent with other classes used in the translator and builder. Can we rename this class please?

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/model/Match.java
```diff
@@ -0,0 +1,27 @@
+package org.cqframework.cql.cql2elm.model;
+
+public enum Match {
+    CASE, CASE_IGNORED, SOUNDS_LIKE, NONE;
```
brynrhodes, line 4: The option `CASE` here should be `EXACT`

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/model/Match.java
```diff
@@ -0,0 +1,27 @@
...
+
+    public static Match checkMatch(String val, String checkVal) {
```
brynrhodes, line 6: The name of this function, to me at least, implies that it will throw an error if there is no match. Suggest naming this something that more clearly indicates this is getting the match type. Say `getMatchType()`?

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/model/Match.java
```diff
@@ -0,0 +1,27 @@
+package org.cqframework.cql.cql2elm.model;
+
+public enum Match {
```
brynrhodes, line 3: This is an enum of Match _types_, and so all the code in general would read better if this was called `MatchType`

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/model/MatchIdentifier.java
```diff
@@ -0,0 +1,75 @@
+package org.cqframework.cql.cql2elm.model;
+
+import org.apache.commons.lang3.tuple.Pair;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class MatchIdentifier {
```
brynrhodes, line 9: Should be IdentifierResolution

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/model/MatchIdentifier.java
```diff
@@ -0,0 +1,75 @@
...
+
+    private Match matchType;
+    private Object resolvedIdentifier;
```
brynrhodes, line 12: Is this the resolved _element_? If so it should be named as such, this makes it sound like it should be a string and it's the actual resolved identifier, not the element that was resolved.

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/model/MatchIdentifier.java
brynrhodes: Could reasonably have an `IdentifierMatch` class which was just the identifier, the matchType and the resolved Element, and then an `IdentifierResolution` class that collected the IdentifierMatches for a given identifier.

Src/java/tools/cql-formatter/src/main/java/org/cqframework/cql/tools/formatter/CqlFormatterVisitor.java
```diff
@@ -219,6 +219,7 @@ private boolean needsWhitespaceBefore(String terminal) {
             case ")": return !inFunctionDefinition() && !inFunctionInvocation();
```
brynrhodes, line 219: Can we move the fix for this to a separate PR?

brynrhodes (changes requested):
Overall this looks good and I like the approach. I'm seeing some failing tests, and I am a bit worried about the number of warnings that are being generated, we might want to limit reporting hiding to only some specific contexts. We also might want to make it an "Info" instead of a "Warning".

Also made a few requests for changes throughout, mostly around naming and code readability, as well as separating out changes that are not related to this specific item.

PR #662, opened 2021-09-02
Elm support include lib

mdnazmulkarim:


Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/TranslatorOptionsUtil.java
```diff
@@ -0,0 +1,77 @@
...
+        requireNonNull(library, "library can not be null");
+        if (library.getAnnotation() == null || library.getAnnotation().isEmpty()) {
+            return null;
+        }
+
+        String translatorOptions = getTranslatorOptions(library.getAnnotation());
+        return parseTranslatorOptions(translatorOptions);
+    }
+
+    private static String getTranslatorOptions(List<Object> annotations){
```
JPercival, line 35: This specific code was a hack-around for the fact that the JSON and XML encodings handle translator annotations differently. We should add tests for the new ELM reader classes to ensure that those are mapped correctly.

Here's the relevant issues from the cql-engine:
https://github.com/DBCG/cql_engine/issues/436

JPercival: https://github.com/DBCG/cql_engine/issues/405

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/LibraryManager.java
```diff
@@ -154,39 +154,164 @@ else if (library != null) {
...
+            }
+            if (library.getParameters() != null && library.getParameters().getDef() != null) {
+                for (ParameterDef parameterDef : library.getParameters().getDef()) {
+                    translatedLibrary.add(parameterDef);
+                }
+            }
+            if (library.getStatements() != null && library.getStatements().getDef() != null) {
+                for (ExpressionDef expressionDef : library.getStatements().getDef()) {
+                    translatedLibrary.add(expressionDef);
+                }
```
JPercival, line 296: One of the issues with running from ELM is that the operator maps may not be populated correctly. There should be some code here to ensure that, and some tests to that effect as well.

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/LibrarySourceProvider.java
```diff
@@ -5,5 +5,5 @@
 import java.io.InputStream;
 
 public interface LibrarySourceProvider {
-    InputStream getLibrarySource(VersionedIdentifier libraryIdentifier);
+    LibraryContentMeta getLibrarySource(VersionedIdentifier libraryIdentifier);
```
JPercival, line 8: The implication here is that a LibrarySourceProvider will return all the types of content that it can. IOW, the translator can't request a specific type of content, e.g.:
`InputStream getLibraryContent(VersionedIdentifier libraryIdentifier, ContentType.ELM);`

I think that's not a desirable characteristic. For example, a LIbrarySourceProvider may integrate with an ELM decompiler to produce CQL. But we _only_ want to do that if absolutely necessary because it's expensive to do. Additionally, certain platforms like Android can't support XML even if it's available, so it's a waste to load it. I think we should be able to specify which content to load, and start with the content that we know works best for a given platform, and progressively fall back to CQL.  Alternatively, the LibrarySourceProvider can report the types of content available without loading it (this would be useful for subsetted FHIR Libraries, as an example) and the translator can then make the decision about which to load.

Secondly, this is a breaking change for the public API of the translator. I think it'd be better to implement a second method, and have the existing API remain the same and default to loading CQL.

JPercival (changes requested):

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/LibrarySourceProvider.java
mdnazmulkarim: Can the following be a solution?

`public enum LibraryContentType {
    CQL,
    JSON_ELM,
    XML_ELM,
    XML,
    JSON,
    JXSON,
    COFFEE,
    ANY
}`

`public interface LibrarySourceProvider {
    InputStream getLibrarySource(VersionedIdentifier libraryIdentifier);
}`



`public interface LibrarySourceProviderExt extends LibrarySourceProvider {

        boolean isLibrarySourceAvaliable(VersionedIdentifier libraryIdentifier, LibraryContentType type);
	InputStream getLibrarySource(VersionedIdentifier libraryIdentifier, LibraryContentType type);
	
	default InputStream getLibrarySource(VersionedIdentifier libraryIdentifier) {
            return getLibrarySource(libraryIdentifier, LibraryContentType.CQL);
    }
}`

`public interface LibrarySourceLoader {

     void clearProviders();
  
     InputStream getLibrarySource(VersionedIdentifier libraryIdentifier);

     void registerProvider(LibrarySourceProvider provider);
  
}`

`public interface LibrarySourceLoaderExt extends LibrarySourceLoader {

    InputStream getLibrarySource(VersionedIdentifier libraryIdentifier, LibraryContentType type);
    boolean isLibrarySourceAvaliable(VersionedIdentifier libraryIdentifier, LibraryContentType type);

    default InputStream getLibrarySource(VersionedIdentifier libraryIdentifier) {
            return getLibrarySource(libraryIdentifier, LibraryContentType.CQL);
    }
}`

mdnazmulkarim: I have updated as above not it is backward compatible.

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/LibraryManager.java
mdnazmulkarim: I believe those checking are inside TranlatedLibrary.
image
I can add some test. Is that ok?

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/TranslatorOptionsUtil.java
mdnazmulkarim: Acknowledging those, I am thinking if we can fix/take those in different PRs if you agree.

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/LibrarySourceProvider.java
brynrhodes: Yes this makes sense

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/LibraryManager.java
```diff
@@ -154,39 +159,179 @@ else if (library != null) {
...
+
+        TranslatedLibrary result;
+        InputStream librarySource = null;
+        boolean loadSuccess = true;
+        LibraryContentType contentType= null;
+        try {
+            if (librarySourceLoader instanceof LibrarySourceLoaderExt) {
+                LibrarySourceLoaderExt sourceLoaderExt = (LibrarySourceLoaderExt) librarySourceLoader;
+                if (sourceLoaderExt.isLibrarySourceAvailable(libraryIdentifier, LibraryContentType.JSON_ELM)) {
+                    contentType = LibraryContentType.JSON_ELM;
```
JPercival, line 215: There are two types of JSON ELM currently. The JXSON and the JSON. So the translation would need to try both to be complete.

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/LibraryManager.java
```diff
@@ -154,39 +159,179 @@ else if (library != null) {
...
+        return null;
+    }
+
+    private TranslatedLibrary generateTranslatedLibraryFromEml(VersionedIdentifier libraryIdentifier, InputStream librarySource, LibraryContentType type, CqlTranslatorOptions options) {
+
+        Library library = null;
+        TranslatedLibrary translatedLibrary = null;
+        try {
+            if (type.equals(LibraryContentType.JSON_ELM)) {
+                library = CqlJsonLibraryReader.read(new InputStreamReader(librarySource));
```
JPercival, line 243: This is there are actually two different JSON readers

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/LibraryManager.java
JPercival: Yep.

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/LibrarySourceLoader.java
```diff
@@ -1,6 +1,8 @@
 package org.cqframework.cql.cql2elm;
 
 import java.io.InputStream;
+import java.util.List;
```
JPercival, line 4: Unused import. Please remove.

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/TranslatorOptionsUtil.java
JPercival: The issue is that line 37 is based on a bug that's present in only one of the 3 types of ELM serialization. It's not deserialized to the correct type (Annotation). Instead, it uses a generic HashMap. Once that bug is fixed, this code will break. I'm saying to do the right thing here, which is to use the Annotation type and then fix the serialization bugs so that the behavior progressively improves as the other bugs are fixed.

JPercival (changes requested):

Issue #507, opened 2020-04-14
Complete mapping expansion for QICore measure


brynrhodes:
Specifically, support complete translation of these measures:

* [EXM124 - Cervical Cancer Screening](https://github.com/cqframework/draft-measures/blob/master/pages/cql/in-progress/qicore/EXM124_QICore4-8.2.000.cql)
* [EXM165 - Controlling Blood Pressure](https://github.com/cqframework/draft-measures/blob/master/pages/cql/in-progress/qicore/EXM165_QICore4-8.5.000.cql)

brynrhodes:
Included in the 1.5.4 release

PR #663, opened 2021-09-02
#507: Completed mapping expansion for model info for USCore and QICore

brynrhodes:


Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/LibraryBuilder.java
```diff
@@ -2475,19 +2485,28 @@ else if (sourceProperty.getScope() != null) {
...
+            if (indexerEnd + 1 < targetMap.length()) {
                 // There are additional paths following the indexer, apply them
                 String targetPath = targetMap.substring(indexerEnd + 1);
                 if (targetPath.startsWith(".")) {
                     targetPath = targetPath.substring(1);
                 }
 
+                if (!targetPath.isEmpty()) {
+                    query.setReturn(of.createReturnClause()
+                            .withDistinct(false).withExpression(of.createProperty()
```
JPercival, line 2497: Why is this distinct "false" as opposed to true?

brynrhodes: Because the source data may have duplicates and they would be expected to be returned. It may be an extension, for example, and we wouldn't be justified in applying a distinct. There may be cases where we would be justified, but we're not detecting those currently.

JPercival (approved):

bryantaustin13 (approved):
LGTM

PR #668, opened 2021-09-16
made defaultResult take the element and context as parameters

JPercival:
* Made the baseVisitor defaultResult take the element and context as parameters
* Noted that AccessModifier is not Trackable (oversight?)

Supporting Use Case:
I the cql-langauge-server I want a list of all the ELM nodes within the hierarchy at a given CQL location. The defaultResult for all nodes is themself if they cover the position, and null if they do not. The default aggregration behavior is to append child nodes to the end of the list. With this change, I only have to override two functions as opposed to all of them.


mdnazmulkarim (approved):

bryantaustin13 (approved):
LGTM

Issue #607, opened 2021-03-02
Add a "warning" when a string literal would match an identifier


brynrhodes:
A common error is the use of string literals in place of identifiers. A helpful warning would be when a string literal would match an identifier.

PR #671, opened 2021-09-20
#607: hidden identifier work and framework for future match warning work

echicoine-icf:
This work encompasses the ongoing development of methodology for maintaining a list of exact matches, case insensitive matches, and lays ground work for future "sounds like" matches.


Src/java/cql-to-elm/src/test/resources/org/cqframework/cql/cql2elm/CMS146v2_Expected.json
```diff
@@ -3,6 +3,36 @@
     "annotation" : [ {
       "translatorOptions" : "",
       "type" : "CqlToElmInfo"
+    }, {
```
echicoine-esac, line 6: This file ended up being the perfect test for the hidden identifier work

echicoine-esac:
> This PR now has conflicts. Can you resolve those?

Conflicts should be resolved currently.

JPercival:
I do understand the need for the `hiddenIdentifier` work here, but I don't understand the use case for "sounds like" matches in a compile. Would you mind elaborating on that?

JPercival:
Closed in favor of #1277

PR #673, opened 2021-09-29
Add library deserializer for xml, json, jxson

mdnazmulkarim:


Src/java/cql-to-elm/build.gradle
```diff
@@ -11,6 +11,7 @@ dependencies {
     compile group: 'org.fhir', name: 'ucum', version: '1.0.3'
     compile group: 'com.fasterxml.jackson.module', name: 'jackson-module-jaxb-annotations', version: '2.10.1'
     compile group: 'com.fasterxml.jackson.core', name: 'jackson-databind', version: '2.10.1'
+    compile group: 'org.glassfish', name: 'javax.json', version: '1.1.4'
```
brynrhodes, line 14: A little concerned about dependencies here, is this something that we should actually put into a separate package? @JPercival thoughts?

brynrhodes (commented):
Looks great overall, just a question about dependencies

Src/java/cql-to-elm/build.gradle
brynrhodes: @mdnazmulkarim , we already have a JAXB deserializer, why do we need to add this glassfish dependency?

mdnazmulkarim: I get this `java.lang.NoClassDefFoundError: javax/json/JsonException` without this dependency. Don't know how jackson missed implementing JsonException

mdnazmulkarim: I have replaced glass-fish with something else here :https://github.com/cqframework/clinical_quality_language/pull/673/commits/ecf6f4eb06f577ec245321aa67a59572c41287a7
Found that jackson-data-bind does not implement JSR373 so we need to an implementation of javax.json-api. Kind of weird that it does not complain during serialization but breaks during de-serialization.

brynrhodes: Okay, I've tried this locally with several different approaches and the original glassfish.org dependency you had here seems like the right one to include. I would make the change, but I don't have write access to your repository. @mdnazmulkarim can you add the glassfish.org dependency back in and I'll approve the PR.

mdnazmulkarim: Added the dependency. Thanks!

brynrhodes (approved):

bryantaustin13 (approved):
LGTM

Issue #657, opened 2021-07-30
Concatenating Comments to Defines in ELM XML


jkotanchik-SB:
For FHIR and QDM measures, the XML generated by the Translator (1.5.2, 1.5.3, 1.5.4) is concatenating line and block comments to the following define statement which causes the define declaration to appear commented out in the ELM.

### Example CQL:
![Screen Shot 2021-07-06 at 11 35 05 AM](https://user-images.githubusercontent.com/56264529/127700057-20fb97f2-4162-48da-af58-285bce3032d8.png)

### Generated XML ELM:
![Screen Shot 2021-07-06 at 11 36 07 AM](https://user-images.githubusercontent.com/56264529/127700075-b89c2cd9-c653-4cff-b350-b7fdd4ea5144.png)

### Generated JSON ELM with the comment separated within the Array:
![Screen Shot 2021-07-06 at 11 39 02 AM](https://user-images.githubusercontent.com/56264529/127700099-f7f2f3cd-a8dc-42f8-964d-5ec518d6381f.png)

jkotanchik-SB:
@mdnazmulkarim

Thanks for checking out this issue! 

> When translator option EnableNarrative is on, in the annotation narrative it includes comments and function definitions which is normal.

I'm not sold that general CQL comments should be included in the ELM annotation narratives. In version `1.4.x` , only the CQL text was included in the ELM annotation narrative when the `EnabledAnnotations` was set (I couldn't find any EnableNarrative option).  Though, I have yet to find a source defining ELM annotation narratives, I am not aware of an explicit change in what is expected in the narrative. 

Was an issue reported requesting general CQL comments be added to the annotation narrative? Or perhaps some change in the CQL 1.5 specification that encouraged general comments be added?

Perhaps @brynrhodes or @cmoesel can clarify the expected contents of an ELM annotation narrative, or if you happen to know of a spec/issue/conversation covering the term, please let me know!

brynrhodes:
The ELM specification supports registering annotations to extend the base ELM functionality, including the source CQL for an expression as an annotation is part of that extended functionality. The original intent when we worked out this functionality with the tooling and measure stakeholders was to include the entire text of the definition (i.e. everything from the end of the previous definition to the beginning of the current one), which would capture comments if they were present in the CQL source. So it wasn't that CQL comments should be included, it was a fix that sometimes CQL comments weren't being included when they should have been. If there is a need, we could look at introducing a translator option to cover whether or not they are included.

[...]

PR #674, opened 2021-09-30
Add line break after comment in annotation narrative

mdnazmulkarim:
Referring issue : https://github.com/cqframework/clinical_quality_language/issues/657

Actually it is not an issue as it has been described, as the reporter state "...which causes the define declaration to appear commented out in the ELM". 
Actually translator renders the define declaration good.
When translator option EnableNarrative is on, in the annotation narrative it includes comments and function definitions which is normal. 

In this PR, added a line break between comment and definition for clarity. The output is as below:

image

### DIFF
--- a/Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/Cql2ElmVisitor.java
+++ b/Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/Cql2ElmVisitor.java
@@ -2,9 +2,9 @@ package org.cqframework.cql.cql2elm;
 
 import org.antlr.v4.runtime.ParserRuleContext;
 import org.antlr.v4.runtime.TokenStream;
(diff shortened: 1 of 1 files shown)


p9g:
Does this fix issue in Bonnie not displaying some CQL https://oncprojectracking.healthit.gov/support/browse/BONNIEMAT-1031 and https://oncprojectracking.healthit.gov/support/browse/BONNIEMAT-1032 ?

brynrhodes:
It will fix the output of the translator, but whether it addresses the issues referenced above would be a question for the MAT/BONNIE teams.

jkotanchik-SB:
@p9g We don't believe it will.

brynrhodes:
@jkotanchik-SB is there some other issue in the translator that needs to be addressed, or is it a MAT/Bonnie issue specifically?

brynrhodes (approved):

brynrhodes (approved):

PR #676, opened 2021-10-12
Elm support translator as included library

mdnazmulkarim:
With JSON, JXSON, XML elms and with translator options checking.
This branch is on top of [link](https://github.com/cqframework/clinical_quality_language/pull/673)


Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/LibraryManager.java
```diff
@@ -154,39 +172,179 @@ else if (library != null) {
...
+                    translatedLibrary.add(parameterDef);
+                }
+            }
+            if (library.getStatements() != null && library.getStatements().getDef() != null) {
+                for (ExpressionDef expressionDef : library.getStatements().getDef()) {
+                    translatedLibrary.add(expressionDef);
+                }
+            }
+        } catch (Exception e) {
+            translationSuccess = false;
```
brynrhodes, line 332: This needs a check to make sure that the resultType is set for each node. This is probably best done as a visitor implementation that would actually have utility elsewhere as well, so I suggest implementing an ElmTypeInferencingVisitor and then using that. I don't think that should hold up this PR, but we at least need a check here to make sure that if the ELM doesn't have resultType for at least each expressionDef that it can't be used as a source, and translationSuccess should be false.

brynrhodes (changes requested):
Overall this looks great, we just need to check to make sure that each ExpressionDef has a resultType before it can be used in the translator. And we should add a test to validate that that is working. From what I can see, the tests are just validating that an ELM library can be loaded, we need a test that validates that a loaded ELM library can be used successfully as part of a translation.

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/ElmJsonLibraryReader.java
```diff
@@ -15,7 +16,7 @@ private ElmJsonLibraryReader() {
     public static JAXBContext getJaxbContext() {
         if (jaxbContext == null) {
             try {
-                jaxbContext = JAXBContext.newInstance(Library.class, Annotation.class);
+                jaxbContext = JAXBContext.newInstance(Library.class, Annotation.class, TypeSpecifier.class);
             } catch (JAXBException e) {
```
brynrhodes, line 20: I don't think we need to do anything like this here, the ELM deserialization should just proceed as normal.

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/LibraryManager.java
```diff
@@ -325,7 +326,17 @@ private TranslatedLibrary generateTranslatedLibrary(Library library) {
             }
             if (library.getStatements() != null && library.getStatements().getDef() != null) {
                 for (ExpressionDef expressionDef : library.getStatements().getDef()) {
-                    translatedLibrary.add(expressionDef);
+
+                    //to do implement an ElmTypeInferencingVisitor; make sure that the resultType is set for each node
+                    if(!(expressionDef.getExpression() instanceof  SingletonFrom)) {
```
brynrhodes, line 331: This can just be a simple test against resultType. If any expressionDef has a null resultType, the ELM cannot be used in the translator.

brynrhodes (changes requested):

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/LibraryManager.java
mdnazmulkarim: Done

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/LibraryManager.java
```diff
@@ -328,11 +328,10 @@ private TranslatedLibrary generateTranslatedLibrary(Library library) {
                 for (ExpressionDef expressionDef : library.getStatements().getDef()) {
 
                     //to do implement an ElmTypeInferencingVisitor; make sure that the resultType is set for each node
-                    if(!(expressionDef.getExpression() instanceof  SingletonFrom)) {
+                    if (!(expressionDef.getExpression() instanceof SingletonFrom)) {
                         if (expressionDef.getResultTypeSpecifier() != null || expressionDef.getResultType() != null) {
```
brynrhodes, line 332: I still don't understand why we're looking specficially at SingletonFrom nodes here? It's any expression

brynrhodes, line 332: And it needs to be looking at the `resultType` property, not the `resultTypeSpecifier` or `resultTypeName`. ELM has classes for representings types in the TypeSpecifier hierarchy, and if the type specifier is only a name, ELM elements can use a QName attribute directly, but the actual `resultType` element is not in ELM (since it;s a graph, and has references to in-memory objects). The `resultType` property is introduced on the Trackable base class, and the translator uses that to store the results of type inference as it goes, and then ensures that the type there is output as a typeSpecifier in the ELM. So in general, a freshly deserialized ELM library will not have `resultType` set, and cannot be used directly by the translator.

brynrhodes: So basically, in the absence of an implementation of the TypeInferencingVisitor, we can't use deserialized ELM in the translator. So this test should fail, and when we get that in place we can add a passing test.

brynrhodes (changes requested):

brynrhodes (approved):

bryantaustin13 (approved):
LGTM

Issue #405, opened 2019-03-27
Missing LocalId in Not(Equivalent()) ELM


brynrhodes:
CMS 157 and 161 (possibly others) have an Equivalent as a child of a Not expression. In the serialized ELM JSON, localId is not present in both cases.

CMS157 snippet:

"localId" : "47",
"locator" : "74:8-74:50",
"type" : "Not",
"operand" : {
"locator" : "74:8-74:50",
"type" : "Equivalent",
"operand" : [
{ "localId" : "45", "locator" : "74:8-74:27", "name" : "ChemoBeforeEncounter", "type" : "AliasRef" }

,
{ "localId" : "46", "locator" : "74:32-74:50", "name" : "ChemoAfterEncounter", "type" : "AliasRef" }

]
}
}

CMS 161 snippet:

"localId" : "55",
"locator" : "66:14-66:67",
"type" : "Not",
"operand" : {
"locator" : "66:14-66:67",
"type" : "Equivalent",
"operand" : [
{ "localId" : "53", "locator" : "66:14-66:37", "name" : "PriorMDDEpisodeEncounter", "type" : "AliasRef" }

,
{ "localId" : "54", "locator" : "66:42-66:67", "name" : "NewOrRecurrentMDDEncounter", "type" : "AliasRef" }

]
}
}

There may be others. Other cases of Equivalent DO have localId.

In this case, the same omission exists in XML:

<operand localId="47" locator="74:8-74:50" xsi:type="Not">
<operand locator="74:8-74:50" xsi:type="Equivalent">
<operand localId="45" locator="74:8-74:27" name="ChemoBeforeEncounter" xsi:type="AliasRef"/>
<operand localId="46" locator="74:32-74:50" name="ChemoAfterEncounter" xsi:type="AliasRef"/>
</operand>
</operand>

Originally submitted as a CQL Jira tracker: https://oncprojectracking.healthit.gov/support/browse/CQLIT-156

brynrhodes:
Included in 1.5.5 Release

Issue #395, opened 2019-03-05
Missing LocalIds In Union Statement


mayerm94:
Certain Union statements are missing localId and locator fields. In the ELM snippet below you'll notice that 2 of the Union statements are missing localId and locator fields. The statements with the missing fields happen to be leaf nodes within the binary-tree representation of the series of union statements.

Currently, the CQL logic:

```
define "Denominator Exclusions":
  "Encounter Less Than Two Days"
    union "Encounter with Thrombolytic Therapy Medication or Procedures"
    union "Encounter with Thrombolytic Medication"
    union "Encounter with Comfort Measures"
    union "Encounter With An INR Greater Than 3.5"
    union "Encounter With Thrombolytic Therapy Prior to Arrival"
    union "Encounter With Thrombolytic Therapy Documented As Already Given"
```

Is being represented in the ELM as follows:

```
"expression" : {
    "localId" : "217",
    "locator" : "103:2-109:73",
    "type" : "Union",
    "operand" : [
        {
            "localId" : "215",
            "locator" : "103:2-108:62",
            "type" : "Union",
            "operand" : [
                {
                    "localId" : "190",
                    "locator" : "103:2-106:41",
                    "type" : "Union",
                    "operand" : [
                        {
                            "localId" : "186",
                            "locator" : "103:2-104:70",
                            "type" : "Union",
                            "operand" : [
                                {
                                    "localId" : "184",
                                    "locator" : "103:2-103:31",
                                    "name" : "Encounter Less Than Two Days",
                                    "type" : "ExpressionRef"
                                },
                                {
                                    "localId" : "185",
                                    "locator" : "104:9-104:70",
                                    "name" : "Encounter with Thrombolytic Therapy Medication or Procedures",
                                    "type" : "ExpressionRef"
                                }
                            ]
                        },
                        {
                            "type" : "Union",
                            "operand" : [
                                {
                                    "localId" : "187",
                                    "locator" : "105:9-105:48",
                                    "name" : "Encounter with Thrombolytic Medication",
                                    "type" : "ExpressionRef"
                                },
                                {
                                    "localId" : "189",
                                    "locator" : "106:9-106:41",
                                    "name" : "Encounter with Comfort Measures",
                                    "type" : "ExpressionRef"
                                }
                            ]
                        }
                    ]
                },
                {
                    "type" : "Union",
                    "operand" : [
                        {
                            "localId" : "212",
                            "locator" : "107:9-107:48",
[...]

PR #677, opened 2021-10-14
Add missing localId for equality and union expression

mdnazmulkarim:
Fixing issue : https://github.com/cqframework/clinical_quality_language/issues/405
and 
https://github.com/cqframework/clinical_quality_language/issues/395


brynrhodes (approved):
This works and so I'll approve, but what it's making me think is that we need a tweak to the ObjectFactory to decorate ELM nodes as they are creating, rather than trying to react to them being created in all the different places that they are. I'll approve on these grounds, but logging a tracker to add an ObjectFactoryEx pattern with a `decorate()` method that gets called on each node as it is created by the factory.

JPercival (approved):

Issue #641, opened 2021-05-19
Alphanumeric versions not allowed on included libraries


jward-123:
Hi,

I have a test library using a mix of alphanumeric characters e.g. "Lib2020". The translator will throw the following error when including a library with this type of version:

![2021-05-19 08_39_29-Window](https://user-images.githubusercontent.com/83230317/118822519-dc37cf80-b87d-11eb-8ed2-6e25538ad8ac.png)

Note it doesn't restrict the version on the library being translated itself, just the libraries it includes. It does allow special characters like ".". I don't see this kind of restriction in the CQL handbook. Should it exist?

brynrhodes:
@jward-123 , no, you are correct, the spec is intentionally silent on version formats, so the translator should not be requiring this... marking as an issue. Is this blocking for you?

jward-123:
Thanks Bryn! No it's not blocking. Right now I have a work around to replace the version with a placeholder before and after the translation.

brynrhodes:
Included in v1.5.7

PR #679, opened 2021-10-28
Support alphanumeric in version specifier

mdnazmulkarim:
Fix for Issue : https://github.com/cqframework/clinical_quality_language/issues/641


Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/model/Version.java
```diff
@@ -12,7 +12,7 @@
     public Version(String version) {
         if(version == null)
             throw new IllegalArgumentException("Version can not be null");
-        if(!version.matches("[0-9]+(\\.[0-9]+)*"))
+        if(!version.matches("[0-9A-Za-z]+(\\.[0-9A-Za-z]+)*"))
```
brynrhodes, line 15: This makes sense on the validation, but the concern is the impact on the comparison behavior. If I'm playing interpreter correctly, this will result in the following versions being "equal": `1.1.0A` and `1.1.0B`. We need to surface this in a way that items that use Versions can detect whether they are comparable at run-time. And then make sure we use that when we do version comparison. So an `isComparable` that would return true if the version consists only of numbers and periods, and false otherwise. Then use that flag in the version loader to determine whether version comparison can be performed.

brynrhodes (changes requested):

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/model/Version.java
```diff
@@ -9,18 +9,23 @@
 
     private String version;
 
+    private boolean isComparable;
+
     public Version(String version) {
         if(version == null)
             throw new IllegalArgumentException("Version can not be null");
         if(!version.matches("[0-9A-Za-z]+(\\.[0-9A-Za-z]+)*"))
```
brynrhodes, line 17: We should not be requiring a specific version format here, the CQL spec doesn't pu9t any limitations on what a valid version string is, so just remove this check and exception

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/DefaultLibrarySourceProvider.java
```diff
@@ -50,9 +50,8 @@ public boolean accept(File path, String name) {
                 if (indexOfVersionSeparator >= 0) {
                     Version version = new Version(fileName.substring(indexOfVersionSeparator + 1));
                     // If the file has a version, make sure it is compatible with the version we are looking for
-                    if (indexOfVersionSeparator == libraryName.length() && requestedVersion == null ||
-                            (version.isComparable(requestedVersion) && version.compatibleWith(requestedVersion))) {
-                        if (mostRecent == null || (version.isComparable(requestedVersion) && version.compareTo(mostRecent) > 0)) {
+                    if (indexOfVersionSeparator == libraryName.length() && requestedVersion == null || version.compatibleWith(requestedVersion)) {
+                        if (mostRecent == null || version.compareTo(mostRecent) > 0) {
```
brynrhodes, line 54: This can't use compareTo without checking isComparable, or it will throw. If both versions are not comparable, then it's based on strict equality only.

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/DefaultModelInfoProvider.java
```diff
@@ -52,10 +52,8 @@ public boolean accept(File path, String name) {
                     String[] fileNameComponents = fileName.split("-");
                     if (fileNameComponents.length == 3) {
                         Version version = new Version(fileNameComponents[2]);
-                        if (requestedVersion == null ||
-                                (version.isComparable(requestedVersion) && version.compatibleWith(requestedVersion))) {
-                            if (mostRecent == null ||
-                                    (version.isComparable(requestedVersion) && version.compareTo(mostRecent) > 0)) {
+                        if (requestedVersion == null || version.compatibleWith(requestedVersion)) {
+                            if (mostRecent == null || version.compareTo(mostRecent) > 0) {
```
brynrhodes, line 56: Same comment applies here, the compareTo method can only be used on comparable versions, otherwise strict equality.

brynrhodes (changes requested):

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/model/Version.java
mdnazmulkarim: Updated this!

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/DefaultLibrarySourceProvider.java
mdnazmulkarim: Updated this!

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/DefaultModelInfoProvider.java
mdnazmulkarim: Updated this. Thanks!

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/DefaultLibrarySourceProvider.java
brynrhodes: This change addresses the comparable case, but it's still not checking the strict equality case. If the versions aren't comparable, but they are equal then use that version.

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/DefaultModelInfoProvider.java
brynrhodes: Same comment here, needs to use equality if the versions aren't comparable

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/DefaultLibrarySourceProvider.java
```diff
@@ -51,7 +51,10 @@ public boolean accept(File path, String name) {
                     Version version = new Version(fileName.substring(indexOfVersionSeparator + 1));
                     // If the file has a version, make sure it is compatible with the version we are looking for
                     if (indexOfVersionSeparator == libraryName.length() && requestedVersion == null || version.compatibleWith(requestedVersion)) {
-                        if (mostRecent == null || version.compareTo(mostRecent) > 0) {
+                        if (mostRecent == null ||
+                                ((version != null && version.isComparable()) &&
+                                        (mostRecent != null && mostRecent.isComparable()) &&
+                                        version.compareTo(mostRecent) > 0)) {
                             mostRecent = version;
```
brynrhodes, line 58: This is correct for the case that both are comparable, but if either or both are not comparable, then it should fall back to strict equality matching.

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/DefaultModelInfoProvider.java
```diff
@@ -53,7 +53,10 @@ public boolean accept(File path, String name) {
                     if (fileNameComponents.length == 3) {
                         Version version = new Version(fileNameComponents[2]);
                         if (requestedVersion == null || version.compatibleWith(requestedVersion)) {
-                            if (mostRecent == null || version.compareTo(mostRecent) > 0) {
+                            if (mostRecent == null ||
+                                    ((version != null && version.isComparable()) &&
+                                            (mostRecent != null && mostRecent.isComparable()) &&
+                                            version.compareTo(mostRecent) > 0)) {
                                 mostRecent = version;
```
brynrhodes, line 60: Same here, needs to fall back to strict equality matching if either or both are not comparable.

brynrhodes (changes requested):
Almost there, one more case to consider

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/DefaultLibrarySourceProvider.java
mdnazmulkarim: Updated this. Thanks!

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/DefaultModelInfoProvider.java
mdnazmulkarim: Updated this. Thanks!

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/model/Version.java
```diff
@@ -74,6 +81,24 @@ public boolean compatibleWith(Version that) {
         return true;
     }
 
+    public boolean matchStrictly(Version that) {
+        return this.version.equals(that.version);
```
brynrhodes, line 85: Maybe a null check to return false if `that` is null?

brynrhodes (changes requested):
So close, one minor comment on null checking in the new matchStrictly

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/model/Version.java
mdnazmulkarim: Thanks! Done.

brynrhodes (approved):

JPercival (approved):
LGTM

Issue #637, opened 2021-04-28
MeasureReport primaryCodePath in modelinfo should not be measure.topic


p9g:
CQL Retrieve [MeasureReport:"SomeValueSet"] returns an error "Could not resolve code path measure.topic for the type of the retrieve FHIR.MeasureReport"

MeasureReport.measure is a canonical, essentially a reference to a Measure.url
MeasureReport.measure.topic does not make sense for the primaryCodePath for MeasureReport since it would need to retrieve the Measure then check Measure.topic.

suggest MeasureReport.type should be the primaryCodePath

brynrhodes:
Agree that makes sense as the primary code path.

brynrhodes:
Included in 1.5.6 release

PR #681, opened 2021-11-02
#637: primaryCodePath in modelInfo pointing to MeasureReport.type

echicoine-esac:


brynrhodes (approved):

brynrhodes (changes requested):
Actually, wait, the primaryCodePath here should be just "type", not "MeasureReport.type", the MeasureReport qualifier is unnecessary.

echicoine-esac:
> Actually, wait, the primaryCodePath here should be just "type", not "MeasureReport.type", the MeasureReport qualifier is unnecessary.

Updated entries to point to type

brynrhodes (approved):

JPercival (commented):
Same question as #682. Are there tooling changes to support this generation?

brynrhodes:
Agreed, added an issue: https://github.com/cqframework/cqf-tooling/issues/324

bryantaustin13 (approved):
LGTM

Issue #667, opened 2021-09-11
Add an option to the LibraryManager to enable caching


brynrhodes:
In an authoring environment, caching of libraries is undesirable since the content of those libraries is changing frequently. Add library caching as an option in the LibraryManager to allow authoring environments to disable the functionality.

JPercival:
One consideration here is that certain platforms may need to be able to expire the cache externally. For example, if there's a resident global LibraryManager in a HAPI FHIR server, updates to Library resources should expire the LibraryManager cache.

The cql-evaluator and HAPI fhir server implemented this behavior by providing a cache-aware wrapper for the LibraryManager.

brynrhodes:
Included in 1.5.6 release

PR #691, opened 2021-11-18
LibraryManager Enable cache option

mdnazmulkarim:
https://github.com/cqframework/clinical_quality_language/issues/667


brynrhodes (changes requested):
I don't think this should be exposed as a Translator Option, it's an option on the LibraryManager, which anything that is calling the translator can provide. It's behavior that ought to be set by the environment, not the user, so doesn't belong in the translator options. It should be exposed as an option on the LibraryManager only.

brynrhodes (approved):

bryantaustin13 (approved):
LGTM

Issue #707, opened 2021-12-22
Backwards compatibility issues with terminology references


brynrhodes:
CQL 1.5 introduced support for terminology types to allow references to terminology to be passed as arguments, rather than forcing an expand whenever a value set or code system is referenced. This capability is backwards-compatible, but there are cases where the translator is not accounting for the new capability correctly, resulting in errors when translating CQL that makes use of value set references as arguments to functions.

The following library illustrates this issue:

```
library TestVSCastFunction

using FHIR version '4.0.1'

include FHIRHelpers version '4.0.1'

valueset "Narcolepsy": 'https://www.ncqa.org/fhir/valueset/2.16.840.1.113883.3.464.1004.1182'

define function "VS Cast Function"(VSet List<System.Code>):
  ( ( cast { "VSet", 1 }[0]as Tuple {
      codes List<System.Code>,
      oid System.String,
      version System.String
    }
  ).codes ) VSetCodes
    return System.Code { code: VSetCodes.code, system: VSetCodes.system }

define function "Conditions in ValueSet"(conditions List<Condition>, codes List<System.Code>):
  conditions C
    where FHIRHelpers.ToConcept(C.code) in codes

define TestConditions:
  [Condition] C
    where C.code in "Narcolepsy"

define TestConditionsViaFunction:
  "Conditions in ValueSet"([Condition], "VS Cast Function"("Narcolepsy"))

define TestConditionsDirectly:
  "Conditions in ValueSet"([Condition], "Narcolepsy")
```

Specifically, in 1.5, the invocation signature of the value set references is `ValueSet`, whereas in 1.4 it is `list<System.Code>`.

brynrhodes:
Some of these changes impact trial-use aspects of the CQL and ELM specifications, An HL7 Jira tracker has been submitted to address those changes: [FHIR-34804](https://jira.hl7.org/browse/FHIR-34804)

brynrhodes:
Included in 1.5.6 release

PR #708, opened 2021-12-30
#707: Fixed several issues related to backwards compatibility with 1.…

brynrhodes:
…4 and the new terminology types:

1. In 1.4 compatibility mode, a reference to a `codesystem` or `valueset` declaration is typed as a `List<Code>`, rather than the `ValueSet` or `CodeSystem` types introduced in 1.5.
2. In 1.4 compatibility mode, references to the new terminology types results in a compile-time error.
3. In 1.4 compatibility mode, terminology membership operators can only be resolved with references to `codesystem` or `valueset` declarations
4. In 1.5 or greater, a `ValueSetRef` will be output with the new `preserve` attribute set to true.

For a more detailed discussion, see https://github.com/cqframework/CQL-Formatting-and-Usage-Wiki/wiki/Implementing-the-CQL-1.5-Terminology-Types

### DIFF
--- a/Src/cql-lm/schema/elm/clinicalexpression.xsd
+++ b/Src/cql-lm/schema/elm/clinicalexpression.xsd
@@ -476,6 +476,7 @@ Note that the recommended approach to statically binding to an expansion set is 
 			<xs:extension base="Expression">
 				<xs:attribute name="name" type="xs:string"/>
 				<xs:attribute name="libraryName" type="xs:string" use="optional"/>
+				<xs:attribute name="preserve" type="xs:boolean" use="optional"/>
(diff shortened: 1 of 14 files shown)


bryantaustin13 (approved):
LGTM

JPercival (approved):

brynrhodes:
@cmoesel , I would definitely like your review here, but given the timeframe and the need to push a release, I'm going ahead with the 2 reviews we have. I'll leave the branch intact to facilitate your review if you have some time in the coming weeks. I also plan to do a session on these changes at the upcoming connectathon.

Issue #713, opened 2022-01-07
Incorrect ELM generated for ValueSet in Referenced library


JPercival:
The following CQL generates incorrect ELM. The reference to VS."Female" should be an expression rather than a ValueSet directly.

```
library ValueSets

valueset "Female": '2.16.840.1.113883.3.560.100.2'
```

```
library Test

include ValueSets called VS

define ListOfCodes: { Code { code: 'M' } }

define IsIncluded: ListOfCodes in VS."Female"
```

```elm
         <expression localId="9" locator="8:20-8:45" xsi:type="AnyInValueSet">
            <codes localId="6" locator="8:20-8:30" name="ListOfCodes" xsi:type="ExpressionRef"/>
            <valueset localId="8" locator="8:35-8:45" name="Female" libraryName="VS"/>
         </expression>
```

JPercival:
Actually an engine bug

PR #714, opened 2022-01-07
Bug 713 included vs ref

JPercival:
Resolves issue #713 

* Non-local refs are treated as an expression


JPercival:
Actually an engine bug

Issue #715, opened 2022-01-10
Expected ElmExpressionRequirement when processing data requirements


brynrhodes:
When processing data requirements, certain expressions result in the error:

`expected ElmExpressionRequirement`

This error prevents processing of data requirements, but is the result of incorrect assumptions made in the requirements gathering step that determines sargability of requirements.

brynrhodes:
Included in v1.5.7

PR #716, opened 2022-01-10
#715: Fixed the error "expected ElmExpressionRequirement" when proces…

brynrhodes:
…sing data requirements for certain expressions

### DIFF
--- a/Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/Cql2ElmVisitor.java
+++ b/Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/Cql2ElmVisitor.java
@@ -3368,7 +3368,7 @@ DATETIME
             // Resolve the terminology target using an in or ~ operator
             try {
                 if (codeComparator == null) {
-                    codeComparator = ((!libraryBuilder.isCompatibleWith("1.5") && terminology.getResultType() instanceof ListType)
+                    codeComparator = (terminology.getResultType() instanceof ListType
                             || (libraryBuilder.isCompatibleWith("1.5")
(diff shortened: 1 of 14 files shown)


bryantaustin13 (approved):
LGTM

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/Cql2ElmVisitor.java
```diff
@@ -3368,8 +3368,10 @@ else if (classType.getPrimaryCodePath() != null) {
             // Resolve the terminology target using an in or ~ operator
             try {
                 if (codeComparator == null) {
-                    codeComparator = terminology.getResultType() instanceof ListType ||
-                            terminology.getResultType().isSubTypeOf(libraryBuilder.resolveTypeName("System", "Vocabulary")) ? "in" : "~";
+                    codeComparator = ((!libraryBuilder.isCompatibleWith("1.5") && terminology.getResultType() instanceof ListType)
+                            || (libraryBuilder.isCompatibleWith("1.5")
+                                    && terminology.getResultType().isSubTypeOf(libraryBuilder.resolveTypeName("System", "Vocabulary"))))
+                            ? "in" : "~";
```
JPercival, line 3374: I see the case that it's 1.5 and a Vocabulary type, and I see the case where it's 1.4 and a List type. 1.4 and Vocabulary type isn't possible (because Vocabulary is 1.5). Is 1.5 and List type possible? I'm thinking cases where the ValueSet is expanded.

brynrhodes: Yes, good point, relaxed.

JPercival (approved):

Issue #672, opened 2021-09-23
Add an option to the translator to ignore embedded ModelInfos


JPercival:
The cql-translator ships with a number of embedded ModelInfos for FHIR, QUICK, etc. The translator also supports dynamic ModelInfo loading, so in principle, these could be externalized and/or ignored in scenarios where someone wanted to use a custom ModelInfo.

brynrhodes:
Initial capability included in v1.5.7

PR #717, opened 2022-01-13
Add option to load embedded model info

mdnazmulkarim:
https://github.com/cqframework/clinical_quality_language/issues/672

### DIFF
--- a/Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/ModelManager.java
+++ b/Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/ModelManager.java
@@ -28,7 +28,7 @@ public class ModelManager {
 
     public ModelManager(boolean enableDefaultModelInfoLoading) {
         namespaceManager = new NamespaceManager();
-        setEnableDefaultModelInfoLoading(enableDefaultModelInfoLoading);
+        this.enableDefaultModelInfoLoading = enableDefaultModelInfoLoading;
         initialize();
     }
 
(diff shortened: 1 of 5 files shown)


Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/ModelManager.java
```diff
@@ -42,6 +55,23 @@ public ModelInfoLoader getModelInfoLoader() {
...
+    public boolean isDefaultModelInfoLoadingEnabled() {
+        return enableDefaultModelInfoLoading;
+    }
+
+    public void setEnableDefaultModelInfoLoading(boolean enableDefaultModelInfoLoading) {
+        this.enableDefaultModelInfoLoading = enableDefaultModelInfoLoading;
+    }
+
+    public void disableDefaultModelInfoLoading() {
+        this.enableDefaultModelInfoLoading = false;
```
brynrhodes, line 67: This would never have any effect, since the option is used in the initialize. For that reason, it's not clear that it should be a settable option.

brynrhodes (changes requested):
I don't think the option should be settable on the ModelManager, since it will only have an effect as part of the constructor. Also, it's not clear how you would set this from the command line or options. Do we need to surface this as an option in the translator options? And if so, what is the use case we are going for here? This is probably worth some discussion. @JPercival

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/ModelManager.java
mdnazmulkarim: Removed the setter.
Also Added an Translator option `DisableDefaultModelInfoLoad` . 
Added a command line argument.
The only use case I found Translator options can dictate modelManager creation is in CqlTranslator main() method.
Thanks!

brynrhodes (approved):

Issue #709, opened 2022-01-02
Document CQL Options file


brynrhodes:
Add documentation for the cql options file format the translator uses. Include a JSON Schema file and use it to address the issue in the VSCode plugin: https://github.com/cqframework/vscode-cql/issues/2

brynrhodes:
The CqlTranslatorOptions class represents the options that can be set for the translator:
https://github.com/cqframework/clinical_quality_language/blob/master/Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/CqlTranslatorOptions.java

1. Document the class with JavaDoc tags
2. Create a JSON Schema file that documents the structure (would be ideal if it could be derived from the CqlTranslatorOptions class and included in the build somehow)
3. Add a discussion about the CqlTranslatorOptions class and the use of cql-options files in directories with the translator to the Translator documentation here: https://github.com/cqframework/clinical_quality_language/blob/master/Src/java/cql-to-elm/OVERVIEW.md

brynrhodes:
Included in v1.5.7

PR #725, opened 2022-02-15
#709: Added JavaDoc entries to CqlTranslatorOptions

echicoine-esac:


brynrhodes (changes requested):
The changes to the test look good, but the class is still called CqlTranslatorOptionsToJSON, which implies that this is a class to serialize CqlTranslatorOptions, but it's not, it's to create the JsonSchema. Can we change the name of the class to CqlTranslatorOptionsToJsonSchema?

brynrhodes (approved):
Looks good, thank you!

PR #737, opened 2022-04-05
Fixed an issue with data requirements processing resulting in a NullP…

brynrhodes:
…ointerException when inconsistent library aliases are used in a collection of related libraries.


sliver007 (approved):
Looks good. Resolves the issue for the repro case.

JPercival (approved):

PR #738, opened 2022-04-15
CQF-1351 - Add QICore 4.1.1 modelinfo to translator

mholck:
Copied over the generated modelinfo file and then updated the Provider to make it available


brynrhodes (changes requested):
This looks great, the only thing I would say is can we add a test for the 4.1.1 model info? You can copy the test fixture from here: https://github.com/cqframework/clinical_quality_language/blob/master/Src/java/cql-to-elm/src/test/java/org/cqframework/cql/cql2elm/qicore/v410/BaseTest.java and the resources from here: https://github.com/cqframework/clinical_quality_language/tree/master/Src/java/cql-to-elm/src/test/resources/org/cqframework/cql/cql2elm/qicore/v410
It should just work, but we should have a test for the modelinfo.

brynrhodes (approved):

rob-reynolds:
Reviewed.  Builds locally.  LGTM.

Approved.

(couldn't assign myself as a reviewer because I don't have rights on the fork)

Issue #739, opened 2022-04-18
Incorrect target mapping for interval-valued elements in a profile-informed authoring model


brynrhodes:
Elements of types in profile-informed authoring models (e.g. US Core and QI Core) that are mapped to interval types in CQL (e.g. Interval<DateTime>) do not output with the conversion function necessary, typically resulting in run-time errors when the translated ELM is executed.

See the "Qualifying Encounters" query in the AdultOutpatientEncounters library of the ecqm-content-qicore-2022 for a reproduction.

brynrhodes:
Included in v1.5.7

PR #740, opened 2022-04-18
#739: The translator now correctly outputs a conversion function when…

brynrhodes:
… accessing a profile-informed authoring model element mapped to an interval type in CQL


JPercival (approved):
Discussed the `target` changes in the QICore model info that point to a conversion function with Bryn. Those may break depending on how the FhirHelpers library is included, but we don't currently have a good story for backwards compatibility there yet. We discussed a few options over chat, and we'll follow-up when we have a design.

PR #743, opened 2022-04-25
Feature 742 fhir platform api

brynrhodes:
Initial definition of the FhirPlatform API


brynrhodes:
Merging without review because this is new functionality so doesn't impact the existing production code base and I need it in place so I can release 1.5.8 ahead of the connectathon.

PR #745, opened 2022-05-03
Using Jackson's XML reader instead of JAXB.unmarshal for ModelInfo files

### LINKED ISSUE #640: Support JSON as a serialization format for ModelInfo
The translator currently only supports XML deserialization of model info, which means there is not a good story for deserialization of model info files in the Android space, which effectively prevents the translator from being used in an Android environment.

vitorpamplona:
This PR replaces JAXB with Jackson for loading ModelInfo files, making it suitable for use on Android. It fixes #640

Notice that:
1 - There were typos on the modelinfo files: `primaryCocdePath` and `primaryCOdePath` (Jackson XML is case sensitive)
2 - There is a [bug](https://github.com/FasterXML/jackson-dataformat-xml/issues/525) on Jackson's XML mapper that requires `xsi:type` to be the first property on XML elements with polymorphic nested objects.


brynrhodes (approved):
Looks great, builds locally, all tests pass

rob-reynolds:
LGTM.  Builds locally, all tests pass.

rob-reynolds:
Also, researched the performance implications and I didn't run performance metrics personally, but from what I can tell the Jackson reader should be more performant.

Issue #582, opened 2020-11-04
Gradle/build issue w/ Java 15


brynrhodes:
Attempting to build w/ Java 15 and the latest Gradle gives the following error:

    java.lang.NoClassDefFoundError: Could not initialize class org.codehaus.groovy.vmplugin.v7.Java7

Issue #512, opened 2020-04-17
Cannot build with Java 14


drdavec:
I am installing on a Mac using Java 14 and get a build error with gradle. A quick search indicated that gradle 6.3 is compatible with Java 14.  I changed gradle version in the wrapper, but still unsuccessful.  Apparently this code is not compatible with 6.3.  I installed on a laptop with Java 12 and build is successful.

Any plans to update for compatibility with gradle 6.3 and Java 14?

Thanks!
  Dave

brynrhodes:
Hi @drdavec! We haven't tried with Java 14, we've been focusing on Java 11 as the stable LTS. And at least one challenge here is that we still have some stakeholders that required Java 8 compatibility. Happy to add this to the backlog though as something to look at pursuing. And if you can provide details about the failures you encountered (dependency issues or the like), that would help with the analysis.

PR #747, opened 2022-05-10
Migrating to Java11 and Gradle 7.4

### LINKED ISSUE #741: Update to Java 11 as minimum required version
Version 6.0 of HAPI (in prerelease as of the time of this writing) requires Java 11. The cql-translator uses some HAPI APIs to generate data-requirements for CQL files. In order to update to the latest HAPI releases, the cql-translator will need to be updated to 11 as a minimum as well.

https://hapifhir.io/hapi-fhir/docs/introduction/changelog.html#hapi-fhir-600-prerelease

To do the upgrade we need to change gradle build to the correct Java version, ensure the CI pipeline is updated to use the correct version, and gather feedback from the community on the timing of the change from 8 to 11.

vitorpamplona:
Fixes #741: Updates the source code compatibility requirements to Java 11. 

Since Java11 removes JAXB, the upgrade requires migrating `JAXB commons` from `0.12.0` to `1.11.1` and the JAXB runtime dependency below to the `xds-to-modelInfo` module: 
- ```    implementation group: 'org.jvnet.jaxb2_commons', name: 'jaxb2-basics-runtime', version: '1.11.1'```

Related to #582 and #512, I also tested migrating to Gradle from `5.6.2` to `7.4.2`, which requires us to: 
1. Rework the [deprecated maven publishing configuration](https://docs.gradle.org/6.9.2/userguide/upgrading_version_5.html#legacy_publication_system_is_deprecated_and_replaced_with_the_publish_plugins) to use the [new plugin](https://docs.gradle.org/current/userguide/publishing_maven.html)
2. Update a these scopes: 
  - `compile` to `implementation`
  - `testCompile` to `testImplementation`
  -  `runtime` to `runtimeOnly`
  
This PR also updates the lib to use HAPI 6.0.1 and FHIR core 5.6.36. 
  
I wasn't able to successfully test the new Maven Publishing plugin. Maybe we do a few snapshot releases to make sure everything is being signed and is correctly packaged?


brynrhodes:
@vitorpamplona , looks like the most recent merges resulted in some conflicts, can you resolve those when you get a chance?

vitorpamplona:
Ready to go, @brynrhodes! You might have some issues with the publishing plugin. I did my best to translate to the new version, but couldn't really test it.

Src/java/build.gradle
```diff
@@ -183,79 +185,89 @@ configure(subprojects.findAll {it.name in ['model', 'elm', 'quick', 'qdm', 'cql-
...
+                    password project.hasProperty("ossrhPassword") ? ossrhPassword : ""
+                }
+
+                /* Use these to test locally (but don't forget to comment out others!)
+                def releasesRepoUrl = "file://${buildDir}/repo"
+                def snapshotsRepoUrl = "file://${buildDir}/ssRepo"
+                */
+
+                // change URLs to point to your repos, e.g. http://my.org/repo
+                def releasesRepoUrl = "file://${buildDir}/repo" //"https://oss.sonatype.org/service/local/staging/deploy/maven2/"
```
brynrhodes, line 250: This works for me locally if I just use the commented out URLs, so go ahead and uncomment them and commit that to the PR and I'll merge it

brynrhodes (changes requested):

JPercival (commented):
HAPI is migrating to Java 11 being required as of 6.0. Do we want to make this change at the same time for the sake of being consistent across the different frameworks? Also, this is a backwards incompatible change. Do we need to consider updating project major versions? It's currently tied to the cql specification level (i.e. 1.5).

brynrhodes:
So @JPercival , are you saying we should also update to HAPI 6.0 with this change?

JPercival:
Other way around. I'm saying we should do this when we update to 6.0. None of the rest of the stack is stable right now due to fhir core and hapi mismatches that got propagated through. I don't want to add a 6.0 change on top of that at the moment, myself.

brynrhodes:
I see, so keep this PR on the backlog for now, agreed. Vitor, can you mark this PR draft until we switch to HAPI 6.0?

JPercival:
@vitorpamplona - What's the state of Java 11 support on Android?

vitorpamplona:
We already use Java11 on the SDK. Google requires Java11 for Android Dev since July 2021. It's very stable at the moment.

JPercival:
Can I change my mind and suggest we go ahead and roll this in since we decided to do a major release on the other PR? There are at least 3 breaking changes under consideration:

* Java 11 - New platform requirements
* Jackson Serialization - Breaks old JSON ELM
* Complier/Translator split - Public API changed

Might as well get them all done. It'll take me a while to update downstream projects but the translator has wider usage so it's probably best to minimize churn here.

JPercival:
@vitorpamplona - Can you update this PR to HAPI 6.0.1 and FHIR core 5.6.36?

https://github.com/hapifhir/hapi-fhir/blob/v6.0.1/pom.xml#L757

JPercival:
@mdnazmulkarim - Can you take a look at this one too?

vitorpamplona:
@JPercival Migration to HAPI 6.0.1 and FHIR core 5.6.36 done.

mdnazmulkarim (approved):

JPercival:
@vitorpamplona - I see some conflicts. Can you resolve and I'll merge?

Issue #746, opened 2022-05-04
Extract CqlCompiler from CqlTranslator


vitorpamplona:
The idea is simple: CQL to ELM compilation should be a separate class from all the other formats (JSON/JXSON/XML/COFFEE) available in the translator. The translator then should be moved out of the current module, taking the dependencies of import and export as well as the utility `main` function with it. 

In that way, the Android app can declare a dependency on [cql-to-elm](https://mvnrepository.com/artifact/info.cqframework/cql-to-elm/) without having to manually remove the JAXB/Moxy components. 

Pros: 
- Simpler dependency chain
- `CqlTranslator` will use a `CqlCompiler` internally. Most, if not all, of the CQL Translator interface and way of use will remain the same. 

Cons: 
- Good amount of work to rename/reorganize current tests between Compilation or Translation. 
- The widespread renaming of classes, enums, and exceptions might require users to have to manually upgrade their code to the newest versions.

vitorpamplona:
I have a branch ready to go: https://github.com/Path-Check/clinical_quality_language/tree/compiler-from-translator-separation

PR #748, opened 2022-05-18
Extract CqlCompiler from CqlTranslator

vitorpamplona:
Extracts the compilation of CQL to ELM from the CqlTranslator class and moves CqlTranslatorOptions outside the CqlTranslator class. 

Fixes: #746


Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/CqlCompiler.java
```diff
@@ -0,0 +1,255 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class CqlCompiler {
+    private Library library = null;
+    private TranslatedLibrary translatedLibrary = null;
+    private Object visitResult = null;
+    private List<Retrieve> retrieves = null;
+    private List<CqlTranslatorException> exceptions = null;
```
JPercival, line 27: With the separation of "Compiler" (which transforms CQL to ELM) and "Translator" (which converts ELM to a given serialization format) it seems most of these exceptions should be "CqlCompilerException". I'd only expect a "CqlTranslatorException" as part of serialization given the separation of the two concepts.

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/CqlCompiler.java
```diff
@@ -0,0 +1,255 @@
...
+        return retrieves;
+    }
+    public Map<String, TranslatedLibrary> getTranslatedLibraries() {
+        return libraryManager.getTranslatedLibraries();
+    }
+
+    public Map<String, Library> getLibraries() {
+        Map<String, Library> result = new HashMap<String, Library>();
+        for (String libraryName : libraryManager.getTranslatedLibraries().keySet()) {
+            result.put(libraryName, libraryManager.getTranslatedLibraries().get(libraryName).getLibrary());
```
JPercival, line 95: Similar to the above comment, this is now "getCompiledLibraries" or "getElmLibraries".

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/CqlTranslatorOptions.java
```diff
@@ -174,30 +186,11 @@ public void setOptions(CqlTranslator.Options... options) {
...
         setOptions(options);
         return this;
     }
 
-    /**
-     * Returns instance of CqlTranslatorOptions formats
-     *
-     * @return
-     */
-    public List<CqlTranslator.Format> getFormats() {
```
JPercival, line 187: This is used as part of the publication process to determine the ELM that FHIR Libraries should be generated with. Not sure we can remove it completely. Several IGs are already using this.

It also raises the question of "CompilerOptions" vs "TranslatorOptions". Not sure we necessarily have to do anything about that, just raising it as a consideration.

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/CqlCompiler.java
vitorpamplona: That was my first impression as well. But then I realized that the class `CqlTranslatorException` is a catch-all term and parent class for many exceptions in this project. And since they are exposed to users, I believe the idea was to simply label exceptions from the CqlTranslator project.

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/CqlTranslatorOptions.java
vitorpamplona: Are people using this? I deleted it because the Format parameter doesn't do anything inside the Compiler or translator. The format is generally passed as another parameter in the call, not as an option of the Translator or Compiler. But I can bring it back.

JPercival: Yes, this is being used in the IG publisher. @brynrhodes would have more details. I'm open to alternatives for setting this.

brynrhodes: Yes, the translator itself doesn't do anything with it, but it's an important option that the environments use to control serialization outputs. It's definitely a "translator" option, not a "compiler" option, but I'm not sure how much mileage we'd actually get out of separating those options at this point.

brynrhodes:
@vitorpamplona , does this PR depend on the Java11 change or could we reasonably move forward with it without that?

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/CqlTranslatorOptions.java
vitorpamplona: Cool, no worries. I will revert the change.

vitorpamplona:
It does not depend on the Java11 change. Actually, none of my recent PRs depend on one another.

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/CqlCompiler.java
vitorpamplona: Converted TranslatedLibrary to CompiledLibrary

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/CqlCompiler.java
vitorpamplona: Coming back to this one. I converted CqlTranslatorException to CqlCompilerException again. I realized that the "Translated" exceptions were all on the code to compile dependencies of a given CQL, which before was still called "TranslatedLibraries".

vitorpamplona:
Ok, @brynrhodes @JPercival I made significant changes in the latest commits. See if you like the idea of renaming `TranslatedLibrary` to `CompiledLibrary`.

brynrhodes (approved):
Very nice work, thank you!

JPercival (approved):

brynrhodes:
This is breaking, so I'm sympathetic to changing the major/minor version, but.... I'd really like to keep major/minor in sync with the CQL version supported, like we've always done. So, what about 1.5.20 for this version? Leave us some room on the 1.5.9... in case we need to do something there, but go to 1.5.20 for this change?

vitorpamplona:
Should we consider doing just one version with this one as well? https://github.com/cqframework/clinical_quality_language/pull/752

JPercival:
> This is breaking, so I'm sympathetic to changing the major/minor version, but.... I'd really like to keep major/minor in sync with the CQL version supported, like we've always done. So, what about 1.5.20 for this version? Leave us some room on the 1.5.9... in case we need to do something there, but go to 1.5.20 for this change?

Hasn't tying the implementation version to the specification version bitten us plenty? It's actually the same issue we face with model infos and fhir helpers:

```cql
using FHIR version '4.0.0'
include FHIRHelpers version '4.0.0'
```
vs

```cql
using FHIR4 version '1.0.0'
include FHIRHelplers4 version '1.0.0'
```

cmoesel:
I have to agree with @JPercival -- I feel like aligning w/ spec versions has been more confusing than helpful.  Continuing on his example, the current approach makes it very difficult to know if I have the latest `FHIRHelpers` version -- as the only way to know is to compare the contents of the library -- which is not ideal.

This PR is about software, which is a little different, but I still feel like there is a lot of value in semantic versioning.  No matter how well you document it, there will be end users who are surprised when a 3rd-level version bump introduces breaking changes (and systems that automatically upgrade the dependency on the assumption that it is safe).  We initially tried to align CQL Execution w/ CQL spec versioning, but made the jump to semantic versioning with cql-execution 2.0.0.  Personally, I'm glad we did!

All that said, if now isn't the time for semantic versioning, then I would suggest you continue w/ the current approach and go from `1.5.8` to `1.5.9` (not `1.5.20`).  I think jumping to `1.5.20` would confuse people (and maybe software too) -- especially if you then release `1.5.9` _after_ `1.5.20`.  If you really need to patch the `1.5.8` line, you could go to a fourth versioning component (e.g., `1.5.8.1`).

But again, I do agree w/ JP.  I think it would be beneficial to seriously reconsider moving to semantic versioning.

brynrhodes:
Consensus is that we should move to semantic versioning, so I will bump the version on this to 2.0.0, and note in the release that we are no longer aligning version of the translator with version of the specification. I will also add to the documentation of each release the supported CQL specification version, maybe even add it as metadata to the package?

cmoesel:
I think that if we need to maintain separate release lines based on which version of the spec a translator release supports, we probably need to think through a strategy for that.  Ideally, the spec version would probably be in the namespace somewhere (e.g., translator4 w/ it's own version sequence, translator5 with it's own version sequence, etc).  I've been away from Java and gradle for too long to make any specific suggestions though.

vitorpamplona:
This is how Google keeps Gradle and Gradle Plugin in sync: https://developer.android.com/studio/releases/gradle-plugin#updating-gradle 

Not great. Not terrible.

brynrhodes:
And a table of translator versions and supported specification versions would be good to document for sure

brynrhodes:
@vitorpamplona , do any of your other PRs have high priority items that you would want to see in a 1.5.9 release before we jump to 2.0.0 with this refactor of the compiler/translator?

brynrhodes:
@p9g , I'm not sure why that version only has major/minor, that is supposed to be the full version of the translator used to do the translation. Can you submit an issue for that?

cmoesel:
> I don't think we need to support separate version lines for the specification at this point, since you can provide the specification version as an option to the compiler.

Perfect.  I think I misunderstood one of your previous comments.  This sounds great and _definitely_ simplifies things!

brynrhodes:
Okay, I've verified the refactor works all the way through the language server. I've refreshed the 1.5.9-SNAPSHOT with what is currently in master and I'll bump the version to 2.0.0-SNAPSHOT on the master and post a fresh snapshot. Then if I can get one more approving review on this PR I'll merge it. Then we can merge the rest of the PRs (less the Java 11 change) into the 2.0.0-SNAPSHOT.

brynrhodes:
Actually, I see JP did approve, so I'll merge.

cmoesel:
You might want to consider creating a `v15` branch now as well.  You can always go back and do it later, but it's probably easier to do now before you merge an 2.0.0 stuff in.

brynrhodes:
Oooh, good point. Too late :)

vitorpamplona:
Thank you, @brynrhodes

p9g:
> @p9g , I'm not sure why that version only has major/minor, that is supposed to be the full version of the translator used to do the translation. Can you submit an issue for that?

done: https://github.com/cqframework/clinical_quality_language/issues/753

JPercival:
The Java way to distinguish between Specification Version (1.5 in this case) and Implementation Version (2.0.0) is to add certain metadata to the jar. That's done here:
https://github.com/cqframework/clinical_quality_language/blob/master/Src/java/build.gradle#L33

See also #753. There might be a way to surface this as a top-level thing on the Readme, or Maven central, or something.

Relevant Java docs (see the sections for Package Versioning and Specification Versioning)

https://docs.oracle.com/javase/7/docs/technotes/guides/versioning/spec/versioning2.html

Issue #523, opened 2020-04-23
Annotation serialization with JXSON


brynrhodes:
Annotations are not serializing correctly with the JXSON serializer

JPercival:
Resolved by #752

Issue #522, opened 2020-04-23
Additional regression testing on JSON/JXSON serialization differences


brynrhodes:
The JXSON format type was introduced to support using Jackson to serialize JSON ELM. A different format is being used to ensure backwards compatibility. This issue is to track that the JXSON format should be regression tested with the JSON format to verify complete backwards compatibility, and then the JSON format deprecated.

PR #752, opened 2022-05-26
Uses Jackson instead of JAXB to Read/Write ELM

vitorpamplona:
This PR completely removes all JAXB dependencies from the runtime of this lib. Fixes [Android SDK's 1423](https://github.com/google/android-fhir/issues/1423)

Major changes:
1. Uses Jackson to read and write XML-encoded ELM libraries

2. Change the type of `Element.annotation` from `List<Object>` to `List<CqlToElmBase>`. 
2.1 Adds a new abstract base type on [cqlannotations.xsd](https://github.com/cqframework/clinical_quality_language/compare/master...Path-Check:master#diff-1edd7f83a5e9a665ee9706bf93b471baa036f8df9ed330454bc265c15bebb725) and changes the associated element on [expression.xsd](https://github.com/cqframework/clinical_quality_language/compare/master...Path-Check:master#diff-279c6b262663547a1dc386c1e12ab4ca8cbaba26bd011c700dc17b22c48866e5)
2.2 Adjusts the XMLMapper and JSONMapper to read and write annotations correctly, changing tests to match. 
2.3 Removes the possibility of getting a `LinkedHashMap` in the `Element.annotation`. 

3. Removes the distinction between `CqlTranslator`'s `toJson` and `toJxson`. Now, both use the same exporter (Jxson).

4. `ElmJsonLibraryReader` and `ElmJxsonLibraryReader` now call exactly the same reader (Jxson). 

5. Differentiates between ModelInfo read/write and ELM read/write. There are now two ObjectMappers that cannot be used interchangeably. 

6. Reviews all associated dependencies and includes only the necessary ones in each module. 
6.1. Eclipse's Moxy has been removed from the compile-level dependencies.


Src/cql-lm/schema/elm/cqlannotations.xsd
```diff
@@ -1,24 +1,37 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema" xmlns="urn:hl7-org:cql-annotations:r1" xmlns:sch="http://purl.oclc.org/dsdl/schematron" targetNamespace="urn:hl7-org:cql-annotations:r1" elementFormDefault="qualified" attributeFormDefault="unqualified">
+	<xs:complexType name="CqlToElmBase" abstract="true">
```
JPercival, line 3: Need to make sure these changes are also synced to the cql ig repo:

https://cql.hl7.org/elm.html

https://github.com/HL7/cql

Src/cql-lm/schema/elm/expression.xsd
```diff
@@ -1,14 +1,20 @@
 <?xml version="1.0" encoding="utf-8"?>
-<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema" xmlns="urn:hl7-org:elm:r1" xmlns:sch="http://purl.oclc.org/dsdl/schematron" targetNamespace="urn:hl7-org:elm:r1" elementFormDefault="qualified">
+<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema"
```
JPercival, line 2: Same as prior comment.

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/ElmJxsonLibraryReader.java
```diff
@@ -1,24 +1,62 @@
...
+import org.hl7.elm_modelinfo.r1.TypeSpecifier;
 
+import java.io.File;
 import java.io.IOException;
+import java.io.InputStream;
 import java.io.Reader;
+import java.net.URI;
+import java.net.URL;
 
 public class ElmJxsonLibraryReader {
```
JPercival, line 24: Is this class redundant now given the changes to ElmJsonLibraryReader?

Src/java/cql-to-elm/src/test/resources/org/cqframework/cql/cql2elm/ElmDeserialize/ANCFHIRDummy.json
```diff
@@ -57,6 +57,7 @@
           "dataType" : "{http://hl7.org/fhir}Observation"
         },
         "annotation" : [ {
+          "type" : "Annotation",
```
JPercival, line 60: Given there is ELM encoded as JSON out in the wild that is missing this property, whats the implication for backwards compatibility?

JPercival (commented):
Resolves #523 , #522

Related discussion at: DBCG/cql_engine/issues/405

JPercival:
Closes #640

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/ElmJxsonLibraryReader.java
vitorpamplona: Yep, but I kept it for backward compatibility. Same for the `toJson` and `toJxson` methods in the `CQLTranslator`

Src/java/cql-to-elm/src/test/resources/org/cqframework/cql/cql2elm/ElmDeserialize/ANCFHIRDummy.json
vitorpamplona: They will need to fix their JSON. :( 

This would only happen on JXSON-generated JSONs from this project. The JSON generation using JAXB always places a `type` attribute. So, the only affected instances would be from the past 6 months or so. That is why the JSON and JXSON reader/writer could not be used interchangeably until now.  We didn't have enough test cases (with annotations that are not the default type) to see this JXSON bug.

Src/cql-lm/schema/elm/cqlannotations.xsd
vitorpamplona: Should I do a PR there? I am not sure what are the procedures to change that repo.

vitorpamplona: hum.. I realized that the [expression.xsd at HL7/CQL](https://github.com/HL7/cql/blob/master/spec/elm/schema/expression.xsd) is not in sync with the one here.

Besides many documentation changes, these two types are not in the HL7/CQL repo

```xml	
<xs:complexType name="ParameterTypeSpecifier">
<xs:annotation>
	<xs:documentation>A type which is generic class parameter such as T in MyGeneric&lt;T extends SomeType&gt;.</xs:documentation>
</xs:annotation>
<xs:complexContent>
	<xs:extension base="TypeSpecifier">
		<xs:attribute name="parameterName" type="xs:string" use="required"/>
	</xs:extension>
</xs:complexContent>
</xs:complexType>
```

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/ElmJxsonLibraryReader.java
JPercival: Probably fine to kill the "toJxson" methods and the redundant LibraryReader. We've had some discussions with implementors on the cql_engine repo and the consensus is that it's fine to standardize on one JSON format. JXSON was specifically added to support Android so the only user of that API is the cql-evaluator/cql-engine.

Src/cql-lm/schema/elm/cqlannotations.xsd
JPercival: @brynrhodes - Do we need to push those new types upstream to the CQL repo?

vitorpamplona: FYI, PR has been created: https://github.com/HL7/cql/pull/76

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/ElmJxsonLibraryReader.java
vitorpamplona: I just removed all JXSON references in https://github.com/cqframework/clinical_quality_language/pull/752/commits/2f34b3617a5415254685759b12b88fb1a9af298c. JSON now is Jackson. Tests using JXSON only were moved to JSON. Duplicated tests were deleted.

JPercival:
Whoops. I merged this before there was a second review. I got confused between a couple of different PRs I was reviewing simultaneously. May need to revert this to get a second approval. @brynrhodes

PR #754, opened 2022-06-09
Removing JAXB.properties from the release files

vitorpamplona:
If implementers want to use JAXB unmarshallers on ELM libraries (packages: `org.hl7.elm.r1` and `org.hl7.cql_annotations.r1`) and ModelInfo (package: `org.hl7.elm_modelinfo.r1`), they must create a `jaxb.properties` file locally, at each package folder, pointing to their JAXB implementation

For instance, if using EclipseMoxy the file should have: 
```properties
javax.xml.bind.context.factory=org.eclipse.persistence.jaxb.JAXBContextFactory
```


JPercival (approved):

vitorpamplona:
With #760 this PR is not needed anymore. :)

PR #757, opened 2022-06-13
Library Manager lookup by LibraryVersion instead of Library

csenn:


Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/NamespaceManager.java
```diff
@@ -72,13 +73,22 @@ public NamespaceInfo getNamespaceInfoFromUri(String namespaceUri) {
...
+        ArrayList<String> pathItems = new ArrayList<>();
         if (namespaceUri != null) {
-            return String.format("%s/%s", namespaceUri, name);
+            pathItems.add(namespaceUri);
         }
-
-        return name;
+        pathItems.add(name);
+        if (version != null) {
+            pathItems.add(version);
```
brynrhodes, line 86: The version can't be part of the URL, since the canonical URL of a library and/or namespace is a version-independent identifier. This is because the functionality here is modeled the same way that FHIR knowledge artifacts are managed, where the "url" and "version" elements of the resource are managed separately, and a versioned canonical reference is constructed using a pipe. We would need to mimic that behavior here to ensure we worked the same way.

brynrhodes (changes requested):
Thank you for the contribution, apologies for the delay in getting to the review. I've made a comment about the namespace url construction.

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/NamespaceManager.java
csenn: Hi @brynrhodes, I've seen this for Value Sets before:
```
valueset "Diabetes": 'https://cts.nlm.nih.gov/fhir/ValueSet/2.16.840.1.113883.3.464.1003.103.12.1001|20190315'
```

So for the path would it be?:
```
https://www.my-namsepace.com/MyLibrary|2.0.0
```

JPercival:
Closed due to incompatibility with the CQL spec

Issue #759, opened 2022-06-15
Support different serialization frameworks for ModelInfo and ELM


brynrhodes:
The current translator uses the JAXB serialization framework for reading ModelInfo as well as for writing ELM for both XML and JSON formats. It can also use the Jackson framework for writing ELM using the made-up JXSON format specifier.

There are known issues with the JAXB JSON format that prevent it from being deserialized by a Jackson reader. In addition, the JAXB serialization framework is not supported on Android.

To better support usage of the translator in different environments, refactor the usage of serialization frameworks to use a service provider pattern so that the dependencies of different serialization frameworks can be isolated to separate packages.

brynrhodes:
Included in v2.0.0

PR #760, opened 2022-06-15
#759: Added service loader pattern for serialization frameworks

brynrhodes:
This PR introduces ModelInfoReader and ElmLibraryReader/Writer interfaces in the cql-to-elm package and a service loader pattern to support using different serialization frameworks based on which packages are included.

model-jackson: ModelInfoReader implementations using Jackson
model-jaxb: ModelInfoReader implementations using JAXB
elm-jackson: ElmLibraryReader/Writer implementations using Jackson
elm-jaxb: ElmLibraryReader/Writer implementations using JAXB

In addition, this PR refactors the CLI into a new cql-to-elm-cli package that has the command-line translator only.


Src/java/build.gradle
```diff
@@ -129,6 +134,7 @@ configure(subprojects.findAll {it.name in ['model', 'elm', 'quick', 'qdm', 'cql-
         xjc group: 'org.slf4j', name: 'slf4j-simple', version: '1.7.30'
         implementation group: 'org.jvnet.jaxb2_commons', name: 'jaxb2-basics-runtime', version: '1.11.1'
         implementation group: 'jakarta.xml.bind', name: 'jakarta.xml.bind-api', version: '2.3.3'
+        implementation group: 'org.eclipse.persistence', name: 'org.eclipse.persistence.moxy', version: '2.7.7'
```
vitorpamplona, line 137: Any chance we can get rid of Moxy here and move to a JAXB module? Android can't import it.

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/DefaultModelInfoProvider.java
```diff
@@ -81,7 +82,7 @@ public boolean accept(File path, String name) {
             if (modelFile != null) {
                 InputStream is = new FileInputStream(modelFile);
 
-                return ModelInfoXmlReader.readValue(is, ModelInfo.class);
+                return ModelInfoReaderFactory.getReader("application/xml").read(is);
```
vitorpamplona, line 85: Wouldn't be better to define these MimeTypes somewhere?

vitorpamplona: I am assuming this needs to be aligned with some sort of capabilities statement

Src/java/elm/src/main/java/org/cqframework/cql/elm/serializing/ElmLibraryWriter.java
```diff
@@ -0,0 +1,14 @@
...
+
+import java.io.*;
+import java.net.URI;
+import java.net.URL;
+
+public interface ElmLibraryWriter {
+
+    public void writeValue(Library library, Writer writer) throws IOException;
+
+    public String writeValueAsString(Library library);
```
vitorpamplona, line 13: If the Reader interface is just `read` methods, I would use just `write` methods here for consistency.

Src/java/elm-jackson/src/main/java/org/cqframework/cql/elm/serializing/jackson/ElmLibraryReaderProvider.java
```diff
@@ -0,0 +1,18 @@
...
+
+public class ElmLibraryReaderProvider implements org.cqframework.cql.elm.serializing.ElmLibraryReaderProvider {
+    @Override
+    public ElmLibraryReader create(String contentType) {
+        if (contentType == null) {
+            contentType = "application/elm+json";
+        }
+
+        switch (contentType) {
+            case "application/elm+xml": return new ElmXmlLibraryReader();
```
vitorpamplona, line 13: We should probably add an issue to create add a `text/cql` reader that compiles the code. Would it make sense?

vitorpamplona: Should this `contentType` also check for different CQL versions and send it to different readers?

vitorpamplona:
I really like the idea of XML and JSON Elm Library readers/writers not even being a part of cql-to-elm. Maybe they shouldn't even be a thing in the long run.

Src/java/build.gradle
vitorpamplona: Resolved on  https://github.com/cqframework/clinical_quality_language/pull/760/commits/e503ad2096ad72fde5948769106c22e839217a3b

Src/java/elm/src/main/java/org/cqframework/cql/elm/serializing/ElmLibraryWriter.java
vitorpamplona: Resolved on [26af72a](https://github.com/cqframework/clinical_quality_language/pull/760/commits/26af72a105ec8e5ec0e4c4eb25f0ea0b2f7d9eb4)

Src/java/build.gradle
```diff
@@ -111,14 +111,19 @@ subprojects {
     }
 }
 
-configure(subprojects.findAll {it.name in ['model', 'elm', 'quick', 'qdm', 'cql-to-elm']}) {
+/*
+JAXB dependencies:
+https://mkyong.com/java/jaxbexception-implementation-of-jaxb-api-has-not-been-found-on-module-path-or-classpath/
+ */
+
+configure(subprojects.findAll {it.name in ['model', 'model-jackson', 'model-jaxb', 'elm', 'elm-jackson', 'elm-jaxb', 'quick', 'qdm', 'cql-to-elm', 'cql-to-elm-cli']}) {
```
vitorpamplona, line 119: Need to add the new projects to the publishing list down below.

brynrhodes: Added

Src/java/elm-jackson/src/main/java/org/cqframework/cql/elm/serializing/jackson/ElmLibraryReaderProvider.java
brynrhodes: I think maybe keep compilation and serialization separate? We have facilities for this in the loader layer in the evaluator and ruler, so I don't think we need to introduce that here

brynrhodes: The versions question is a good one, but we should maybe deal with that separately. At this point all ELM is backwards compatible, so it shouldn't be an issue with using the latest readers

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/DefaultModelInfoProvider.java
brynrhodes: Probably, but given the standard nature, and the need to introduce a dependency to define them somewhere, I'm okay with them being magic values for now

vitorpamplona (approved):

JPercival (approved):

PR #772, opened 2022-07-18
Add handling for empty source interval

JPercival:
ANTLR produces an interval of i..i-1 in cases where the interval is empty. This causes the current Annotation code to break since it doesn't know how to compare and create parent/child relationships between empty intervals. This PR skips creating annotations for intervals that are empty. An empty interval _should_ only occur if there are syntax errors (i.e. an invalid CQL library).


JPercival:
Also note the 2.0 branch of the translator needs this as well

vitorpamplona (approved):

PR #783, opened 2022-08-02
develop snapshot version 2.1.0

mdnazmulkarim:


brynrhodes (changes requested):
We usually increment a minor version, rather than a revision, saving those for hotfixes, so 2.1.0-SNAPSHOT, rather than 2.0.1-SNAPSHOT

brynrhodes (approved):

PR #788, opened 2022-08-04
Clean up LibrarySourceProvider API

JPercival:
In updating the cql-evaluator to use the 2.0 translator some inconsistencies were discovered in the API:
* The LibrarySourceProvider defines both
`boolean isLibrarySourceAvailable(VersionedIdentifier libraryIdentifier, LibraryContentType type)`
which implies support for multiple content types, and 
`LibraryContentType getLibrarySourceType();`
which implies support for only a single content type per LibrarySourceProvider.

To fix this, the following was done:
* Remove LibrarySourceProviderExt and LibraryLoaderExt interfaces
* Updated the existing interfaces to allow support of multiple content types.
* Add default implementations of new methods to ensure backwards compatibility.

There was also a bug in the ELM loading in the Translator:
* The Translator would load `ANY` content and assume it was CQL.
* I removed the `ANY` enumeration. The translator can't load `ANY` content, only the content that's specifically enumerated
  * In the future, it might be wise to do further refactoring to allow pluggable content types.

Additionally, a typo was fixed:
`compilatonStack` -> `compilationStack`

And errors for included libraries were made consistent. And included library while always give an `CqlCompilerException` instead of an `IllegalArgumentException` now.


mdnazmulkarim (approved):
Awesome!

brynrhodes (approved):

Issue #771, opened 2022-07-12
CQL-to-ELM: Annotation Parsing


bkaney:
When using cql-to-elm version 1.5 there are a few issues when processing annotations

## 1. Tag-only annotations do not appear in ELM

CQL:
```
/**
 * Annotations for define
 * 
 * @test
 */
```

ELM actual:
```
<annotation xsi:type="a:Annotation">
  <a:s r="8">
    <a:s>/**
 * Annotations for define
 *
 * @test
 */
     </a:s>
     ...snip...
  </a:s>
</annotation>
```

ELM expected:
```
<annotation xsi:type="a:Annotation">
  <a:s r="8">
    <a:t name="test"/>
    <a:s>/**
 * Annotations for define
 *
 * @test
 */
     </a:s>
     ...snip...
  </a:s>
</annotation>
```

(There is a workaround to use `@test:`, but this it not documented. Should the spec allow `@test` without colon?)

## 2. Handling end of line / annotation

CQL
```
/**
 * Annotations for define
 *
 * @test:
 * @data: has-t2dm-cohort
 * @context: Patient
 */
```

ELM acutual
```
<annotation xsi:type="a:Annotation">
   <a:t name="test" value="*"/>
   <a:t name="data" value="has-t2dm-cohort&#xa; *"/>
   <a:t name="context" value="Patient=123"/>
   <a:s r="13">
      <a:s>/**
* Annotations for define
*
* @test:
* @data: has-t2dm-cohort
* @context: Patient=123
*/
      </a:s>
     ...snip...
  </a:s>
</annotation>
```

ELM expected
```
<annotation xsi:type="a:Annotation">
   <a:t name="test"/>
   <a:t name="data" value="has-t2dm-cohort"/>
   <a:t name="context" value="Patient=123"/>
   <a:s r="13">
      <a:s>/**
* Annotations for define
*
* @test:
* @data: has-t2dm-cohort
* @context: Patient=123
*/
      </a:s>
     ...snip...
  </a:s>
</annotation>
```

## 3. Handling annotation values with `@`

CQL
```
/**
 * Annotations for define
 *
 * @asof: @2020-10-01 
 * @parameter: "Measurement Interval" [@2019,@2020]
 */
```

ELM actual
```
<annotation xsi:type="a:Annotation">
   <a:t name="asof" value=""/>
   <a:t name="parameter" value="&quot;Measurement Interval&quot; ["/>
   <a:s r="8">
      <a:s>/**
[...]

PR #790, opened 2022-08-12
Fix tag parsing

mdnazmulkarim:
Related [issue](https://github.com/cqframework/clinical_quality_language/issues/771): 
- With the principle, if a tag has a value, it needs the colon to separate it
- Support ` : `as tag delimiter e.g. `@test : value`
- `@test @2010-10-01` produce tag name = test and tag value = @2010-10-01 
- supports multi tags in one line
- For this: `@test  @pertinence: strongly-positive`,  no tag value for test and tag value for pertinence = "strongly-positive"
-For this: `@parameter: "Measurement Interval" [@2019,@2020]`, tag value for parameter= "Measurement Interval" [@2019,@2020]
- Limitation: for parameter it expects to be in one line or be the last tag of a multi tag line.

### DIFF
--- a/Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/Cql2ElmVisitor.java
+++ b/Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/Cql2ElmVisitor.java
@@ -411,7 +411,8 @@ public class Cql2ElmVisitor extends cqlBaseVisitor {
 
     // this method returns Pair<tag name, next value lookup index> starting from startFrom
     // can return null in cases.
-    // supports both `:` and ` ` for delimiter
+    // if a tag has a value, it needs the colon to separate it
+    // supports `:` for delimiter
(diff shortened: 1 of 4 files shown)


Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/Cql2ElmVisitor.java
```diff
@@ -350,30 +351,126 @@ private boolean isValidIdentifier(String tagName) {
         return true;
     }
 
+    // this method returns Pair<tag value, next tag name lookup index> starting from startFrom
+    // can return null in cases.
+    // for @1980-12-01, it will potentially check to be treated as value date
+    private Pair<String, Integer> looKForTagValue(String header, int startFrom) {
```
JPercival, line 357: LooK -> Look

JPercival (approved):

brynrhodes (approved):

Issue #786, opened 2022-08-04
Dependency cycle in cql2elm


brynrhodes:
FHIRModelInfoProvider depends on ModelInfoProvider defined in cql2elm, but the cql2elm tests depend on FHIRModelInfoProvider. I used testImplementation to declare the dependency, but I intended testRuntimeOnly. Might also make sense to move ModelInfoProvider to the model package.

brynrhodes:
Addressed in 2.1.0 release

PR #791, opened 2022-08-22
#786: Resolve dependency cycle in cql2elm

barhodes:
Move ModelInfoProvider into model and create ModelIdentifier in model to be used in place of VersionedIdentifier.


vitorpamplona:
Nothing like the sight of `var` to make me happy.

brynrhodes (approved):
Looks great, builds locally, tests pass.

PR #793, opened 2022-08-22
Removes  Antlr's code generation dependencies from the runtimeClasspath.

vitorpamplona:
Dependency cleanup: 
- Removing Antlr's dependencies
- Removing hapi-fhir-client (http client not available on android)
- Removing duplicated entires on the build.


brynrhodes (approved):
Looks good, builds locally, tests pass.

PR #797, opened 2022-08-30
Swap xpp3 implementation to support Java 9+

JPercival:
The HAPI/UCUM dependencies pull in a version of the `xpp3` parser that is no longer maintained. This PR swaps to a fork that has been more recently updated. This fork also removes the `javax.xml.namespace` package from the jar, fixing the "The package javax.xml is accessible from more than one module: <unnamed>, java.xml" error when compiling on recent versions of Java.

### DIFF
--- a/Src/java/.vscode/settings.json
+++ b/Src/java/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "java.configuration.updateBuildConfiguration": "automatic"
+}
--- a/Src/java/build.gradle
+++ b/Src/java/build.gradle
@@ -31,6 +31,14 @@ subprojects {
         testImplementation group: 'org.slf4j', name: 'slf4j-simple', version: '1.7.30'
     }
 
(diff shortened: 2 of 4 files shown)


vitorpamplona:
I wish they would just take my Service Loader PR over there https://github.com/FHIR/Ucum-java/pull/29 

FYI, an implementation of the XPP3 Pull Parser is offered in the Android SDK directly. Ideally, we want to force neither UCUM's version nor the `org.codelibs` version.

barhodes (approved):

JPercival:
@vitorpamplona yeah, we'd need that PR and also one for the HAPI validation framework. I suppose we could just exclude the dependency entirely. Then it'd be up to downstream projects to supply one.

brynrhodes (approved):

brynrhodes:
Regardless of what happens with the UCUM library, I should merge this PR to address the dependency issues in the short term, yes?

PR #807, opened 2022-09-15
Make service loaders error if multiple implementations are found on classpath

JPercival:
This PR changes the ServiceLoader behavior so that if an end-user either neglects to include a jackson/jaxb implementation or if both implementations are include they get an error message. This required rearranging a few projects and dependencies to make it work, since the cql-to-elm project tests referenced both. Additionally, I killed the fhir-r4 project temporarily because it's making downstream updates confusing for end users (it's duplicated from quick). Will revert that change once things are stablized.


brynrhodes (approved):

barhodes (approved):

Src/java/elm-jackson/build.gradle
```diff
@@ -6,4 +6,6 @@ dependencies {
     implementation group: 'org.apache.commons', name: 'commons-text', version: '1.9'
     implementation group: 'com.fasterxml.jackson.dataformat', name: 'jackson-dataformat-xml', version: '2.13.2'
     implementation group: 'com.fasterxml.jackson.module', name: 'jackson-module-jaxb-annotations', version: '2.13.2'
+        // needs javax.json.JsonException when using SAXUnmarshaller.getNewXMLReader
+    implementation group: 'org.glassfish', 'name': 'javax.json', 'version': '1.0.4'
```
vitorpamplona, line 10: is this in the right place? javax.json is supposed to be a JAXB dependency not Jackson.

JPercival: Apparently, jackson uses it as well because I get that same exception without it.

Issue #799, opened 2022-09-06
Gradle to Pom.xml for info.cqframework.cql is wrong


vitorpamplona:
The compile-time dependency on `org.antlr:antlr4` should not be [listed as a runtime dependency](https://mvnrepository.com/artifact/info.cqframework/cql/2.1.0).

The pom file generated from Gradle correctly contains: 
```
  <dependencies>
    <dependency>
      <groupId>org.antlr</groupId>
      <artifactId>antlr4-runtime</artifactId>
      <version>4.10.1</version>
      <scope>compile</scope>
    </dependency>
  </dependencies>
  ```
  
But when [published](https://mvnrepository.com/artifact/info.cqframework/cql/2.1.0), the full-fat `org.antlr:antlr4` dependency came back  

Maybe there was some weird dependency clean up when publishing this version?

**Comment by vitorpamplona** [comment]
Documenting this here as well: 

There's something really strange with the <dependencies> section

[info/cqframework/cql 2.2-SNAPSHOT](https://oss.sonatype.org/service/local/repositories/snapshots/content/info/cqframework/cql/2.2.0-SNAPSHOT/cql-2.2.0-20220927.215651-3.pom) had only one dep: antlr4-runtime
```xml
<dependencies>
  <dependency>
    <groupId>org.antlr</groupId>
    <artifactId>antlr4-runtime</artifactId>
    <version>4.10.1</version>
    <scope>compile</scope>
  </dependency>
</dependencies>
```

But when the final version [info/cqframework/cql 2.2](https://oss.sonatype.org/service/local/repositories/releases/content/info/cqframework/cql/2.2.0/cql-2.2.0.pom) arrives, it gets another dependency on full antlr4
```xml
<dependencies>
  <dependency>
    <groupId>org.antlr</groupId>
    <artifactId>antlr4</artifactId>
    <version>4.10.1</version>
    <scope>compile</scope>
  </dependency>
  <dependency>
    <groupId>org.antlr</groupId>
    <artifactId>antlr4-runtime</artifactId>
    <version>4.10.1</version>
    <scope>compile</scope>
  </dependency>
</dependencies>
```

There was no change in code to justify this.

vitorpamplona:
Apparently, this has been fixed in 2.4.

PR #809, opened 2022-09-17
Use 'java-library' plugin, clean up dependencies

JPercival:
Apologies for the churn on the `xpp3` bits, still discovering better ways to do things.

This PR changes from the `java` gradle plugin to the `java-library` plugin. Documentation for this plugin is available here, but the gist of it is that it extends the `java` plugin with some more flexible dependency options:
https://docs.gradle.org/current/userguide/java_library_plugin.html

Specifically, it introduces a new dependency scope called `api`, which means that a dependency should be included at both compile time and runtime, and should be included transitively for downstream users. This is in contrast to `implementation` which includes a dependency at compile-time only and is not transitive. This meant redeclaring many dependencies throughout the repo, and caused several to be imported at compile time when they were needed only at runtime (xpp3 being the biggest culprit here).

I reworked the dependencies in the cql-translator such that a user of `cql-to-elm` will get all the required upstream dependencies (`cql`, `elm`, `model`, etc.) automatically, which is a common stumbling block for integrators ("why doesn't `cql-to-elm` work out of the box?"). This in conjunction with the previous PR that prompts users to include the specific `elm-jaxb` (or jackson) and `model-jaxb` (or jackson) libraries minimize the amount of churn required to get up and running with the cql-compiler. Best case, a user needs to declare only 3 dependencies:

```
implementation 'info.cqframework:cql-to-elm:2.1.0'
implementation 'info.cqframework:elm-jackson:2.1.0'
implementation 'info.cqframework:model-jackson:2.1.0'
```

With respect to `xpp3`, it's now only being pulled in via the `ucum` dependency. It's been excluded as compile-time dependency and added a test-runtime-only dependency. This should allow end-users to specify their own pull-parser where needed and maximize compatibility with existing use cases.

### DIFF
--- a/Src/java/build.gradle
+++ b/Src/java/build.gradle
@@ -31,18 +31,6 @@ subprojects {
         testImplementation group: 'org.slf4j', name: 'slf4j-simple', version: '1.7.36'
     }
 
-    // The FHIR ucum service depends on an
-    // XML pull-parser being present on the classpath, and have a compile-time reference to xpp3.
-    // However, some platfoms such as Android provide their own pull-parser implementations.
-    // Additionally, The xpp3.xpp3 dependency includes the javax.xml packagename, which is invalid in
-    // Java 9+ because split packages (i.e. the same packagename being available in multipe jar) are
-    // disallowed. For those reasons xpp3 is excluded here, and included as a runtime dependency for
-    // tests. Users of the CQL translator will need to supply their own implementation or add xpp3
-    // at runtime.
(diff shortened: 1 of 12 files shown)


vitorpamplona:
I should have caught this `java-library` issue. 

Let's make sure the poms represent what we want in gradle. I still don't know how this issue was possible: https://github.com/cqframework/clinical_quality_language/issues/799

vitorpamplona (approved):

PR #811, opened 2022-09-26
Fix qicore modelinfo

brynrhodes:
Expanded retrieve translation to support the case that a primary code path is a choice of Concept and ValueSet, resulting in the union of retrieves against the code element as a Concept and the code element as a ValueSet.

Expanded retrieve translation to support the case that a primary code path is a ValueSet, resulting in the use of a "contains" codeComparator.
Added a special-case exception to the use of the ValueSet type for FHIRHelpers in 1.4. It won't actually use it, but it has to be able to compile in order to be used backwards-compatibly.
Updated FHIRHelpers to add the ToValueSet function.
Updated QICore 4.0.0, 4.1.0, and 4.1.1 model infos.
Updated USCore 3.1.0 and 3.1.1 model infos.


mdnazmulkarim (approved):

bryantaustin13 (approved):
LGTM

Issue #818, opened 2022-10-07
Outputting to System.Error on Success 


alexgwalley:
In src/java/cql-to-elm-cli/src/main/java/org/cqframework/cql/cql2elm/cli/CqlTranslator.java:
all outputs are directed to standard error, making it impossible to check for actual errors (the program will always report failing even on success)

![image](https://user-images.githubusercontent.com/10475866/194600565-73d65270-e20b-426e-a514-8073e85f025a.png)

There are also other examples in this file where a similar issue occurs. I believe just changing every output which is not an error to System.out.println will fix the issue.

JPercival:
This is actually correct behavior according to the POSIX guidelines. The actual output of the translator is the serialized ELM, which could be written to stdout (as opposed to a file) in order to pipe commands together. Sideband information is supposed to be written to stderr.

https://unix.stackexchange.com/questions/331611/do-progress-reports-logging-information-belong-on-stderr-or-stdout

PR #819, opened 2022-10-07
Fixing bug with outputting success to StdError

alexgwalley:
Changes unnecessary System.Error.println commands to System.out.println so that developers can easily determine if translation completed successfully.

### DIFF
--- a/Src/java/cql-to-elm-cli/src/main/java/org/cqframework/cql/cql2elm/cli/CqlTranslator.java
+++ b/Src/java/cql-to-elm-cli/src/main/java/org/cqframework/cql/cql2elm/cli/CqlTranslator.java
@@ -53,8 +53,8 @@ public class CqlTranslator {
 
     private static void writeELM(Path inPath, Path outPath, org.cqframework.cql.cql2elm.CqlTranslator.Format format, CqlTranslatorOptions options) throws IOException {
 
-        System.err.println("================================================================================");
-        System.err.printf("TRANSLATE %s%n", inPath);
+        System.out.println("================================================================================");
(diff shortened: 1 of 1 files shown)


alexgwalley:
This is linked to this issue: https://github.com/cqframework/clinical_quality_language/issues/818

brynrhodes (approved):
Looks good to me, the only concern I have is whether there are downstream implications for applications that may have made allowances for the fact that the translator was outputting to stderr, especially for the evaluator CLI and the VSCode plugin.

JPercival:
posix cli conventions require such message to go to std err, while output (in this case, the actual ELM) would be written to std out:

https://pubs.opengroup.org/onlinepubs/9699919799/functions/stdin.html

>At program start-up, three streams shall be predefined and need not be opened explicitly: standard input (for reading conventional input), standard output (for writing conventional output), and standard error (for writing diagnostic output). 

This allows cli commands to be composable:

```bash
// compile and run
cql -l Test.cql | cql-execute -data whatever/data -context patient=123
```

or:
```bash
// look for ValueSetRef nodes
cql -l Test.cql | grep 'ValueSetRef'
```

or:
```bash
// upload ELM to a server
cql -l Test.cql --output json | curl -X https://some.server.com/recieve-some-elm
```

If you write status message to std out you pollute the elm output with status messages, thus creating invalid ELM that you can't immediately consume. Granted, the current cli tooling doesn't support patterns like this but that is the intent of outputting to err rather than out. Given that we're already following the standard conventions I'd prefer to keep doing that but I could be convinced otherwise.

PR #834, opened 2022-10-28
make try elm optional

mdnazmulkarim:
I am considering having an option here. However `CqlTranslatorOptions` might not be wise one as the user might use the same options object in the successive translate call. @brynrhodes would you suggest `LibraryManager.resolveLibrary()` have an extra boolean parameter or any other suggestion

### DIFF
--- a/Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/CqlTranslatorOptions.java
+++ b/Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/CqlTranslatorOptions.java
@@ -29,7 +29,7 @@ public class CqlTranslatorOptions {
     private EnumSet<Options> options = EnumSet.noneOf(Options.class);
     private boolean validateUnits = true;
     private boolean verifyOnly = false;
-    private boolean tryElm = true;
+    private boolean enableCqlOnly = false;
     private String compatibilityLevel = "1.5";
     private CqlCompilerException.ErrorSeverity errorLevel = CqlCompilerException.ErrorSeverity.Info;
     private LibraryBuilder.SignatureLevel signatureLevel = LibraryBuilder.SignatureLevel.None;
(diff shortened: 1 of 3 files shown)


vitorpamplona:
Any chance we can make the distinction between options that enable/disable CQL features (incompatible behavior) from compilation/translation runtime options? Keep in mind, the compilation process exports these options, and the translator only loads compiled ELM that fully matches the desired options.

brynrhodes:
I think this approach is fine, so long as we modify the "translatorOptionsMatch" function to only care about the options that actually impact language features:

https://github.com/cqframework/clinical_quality_language/blob/master/Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/LibraryManager.java#L356

So that would be that the following options all match:
* EnableAnnotations
* EnableLocators
* EnableResultTypes
* EnableDetailedErrors
* DisableListTraversal
* DisableListDemotion
* DisableListPromotion
* EnableIntervalDemotion
* EnableIntervalPromotion
* DisableMethodInvocation
* RequireFromKeyword
* compatbilityLevel
* errorLevel
* signatureLevel

vitorpamplona:
Why would these options matter for the option check? 

- EnableAnnotations
- EnableLocators
- EnableResultTypes
- EnableDetailedErrors
- compatbilityLevel
- errorLevel
- signatureLevel

Aren't they designed for debugging functions within the LSP? They don't change CQL behavior, right?

cmoesel:
@vitorpamplona - I don't know the full context of this PR, but want to note that `EnableResultTypes` and `signatureLevel` are options that can be useful _outside of debugging_.  For example, the TypeScript engine will use result types and signatures to ensure that it calls the right function in cases where we have function overloads (e.g., functions w/ the same names but different arguments).

vitorpamplona:
Hah! I didn't know that. Shall we create a page somewhere to describe the impact of each of these annotations?

cmoesel:
Well, there is already some documentation of options in the [OVERVIEW.md](https://github.com/cqframework/clinical_quality_language/blob/master/Src/java/cql-to-elm/OVERVIEW.md#usage), but it doesn't go so far as to suggest _why_ people might use these options.

vitorpamplona:
Ok, so my current understanding is: 

Master option: 
- EnableAnnotations

Language Features: 
- DisableListTraversal
- DisableListDemotion
- DisableListPromotion
- EnableIntervalDemotion
- EnableIntervalPromotion
- DisableMethodInvocation
- RequireFromKeyword

Annotated Helpers for Interpreters: 
- EnableResultTypes - function overloads
- SignatureLevel - function overloads

Dependency Loader Control: 
- EnableCqlOnly - Doesn't load JSON or XML dependencies
- CompatibilityLevel - Compatibility level for the translator, valid values are 1.3, 1.4, and 1.5

Debbuging Information for Language Servers. 
- EnableLocators - Find source of errors 
- ErrorLevel - Error, Warnings, Info, etc. 
- EnableDetailedErrors - 
- EnableResultTypes - Code completion

--- 

I think the most difficult one is CompatibilityLevel, which is currently used in the dependency library loader but in a way that only loads libraries that declare exactly the same compatibility level value the loader is expecting as opposed to just using the annotation in the controller to check if the library is compatible with that version (removing the need to compare the annotations themselves)

brynrhodes:
Added a new issue to consider more granular use cases for translatorOptionsMatch, but as is it is inclusive of all use cases and so is functional, merging on those grounds.

https://github.com/cqframework/clinical_quality_language/issues/843

JPercival (approved):

brynrhodes (approved):

PR #1111, opened 2023-01-13
Feature 850 datetime datarequirements

brynrhodes:


Src/java/elm-fhir/src/test/java/org/cqframework/cql/elm/requirements/fhir/DataRequirementsProcessorTest.java
```diff
@@ -893,6 +900,53 @@ public void TestDataRequirementsAnalysisCase2a() throws IOException {
...
+        Query q = (Query)ed.getExpression();
+        assertTrue(q.getSource() != null && q.getSource().size() == 1);
+        AliasedQuerySource source = q.getSource().get(0);
+        assertTrue(source.getExpression() instanceof Retrieve);
+        Retrieve r = (Retrieve)source.getExpression();
+        assertTrue(r.getDateFilter() != null && r.getDateFilter().size() == 1);
+        DateFilterElement dfe = r.getDateFilter().get(0);
+        assertEquals(dfe.getProperty(), "onset");
+        assertTrue(dfe.getValue() instanceof Interval);
+
```
mdnazmulkarim, line 931: Just for my learning, don't how translator knew DateFilterElement is Interval rather than other potential types.

Src/java/engine/src/main/java/org/opencds/cqf/cql/engine/execution/Context.java
```diff
@@ -644,6 +644,14 @@ public void setParameter(String libraryName, String name, Object value) {
         }
     }
 
+    public void setParameters(Library library, Map<String, Object> parameters) {
+        if (parameters != null) {
+            for (Map.Entry<String, Object> parameterValue : parameters.entrySet()) {
+                setParameter(null, parameterValue.getKey(), parameterValue.getValue());
```
mdnazmulkarim, line 650: Just thinking : is setParameter(), libraryName always be null

mdnazmulkarim (approved):

Src/java/elm-fhir/src/test/java/org/cqframework/cql/elm/requirements/fhir/DataRequirementsProcessorTest.java
brynrhodes: DateFilter values are always represented as an interval. (i.e. if the comparison being captured is to a single point, the point is represented as a unit interval on that point. It makes the downstream processing a bit easier if it can always know that it's dealing with an interval.

Src/java/engine/src/main/java/org/opencds/cqf/cql/engine/execution/Context.java
brynrhodes: It won't always be null, but if it is, it means the parameter will resolve in any library. If it's not, (i.e. it's qualified), it will only resolve in that library.

JPercival (approved):

PR #1122, opened 2023-02-04
Cache Function Defs

JPercival:
* Cache the functionDefs for a given FunctionRefEvaluator.

The CQL engine uses run-time function overload resolution because type information is frequently missing from the ELM graph. However, the CQL spec actually requires function overloads to be distinguishable at compile-time. This implies that for a given FunctionRef in the ELM graph the CQL engine will only ever resolve one and only one FunctionDef (given a valid set of ELM). This means we can resolve it once and then cache it for future FunctionRef evaluations in the event it's evaluated multiple time (for example, in the case of a `where` clause or a `let` clause).

### DIFF
--- a/Src/java/engine/src/main/java/org/opencds/cqf/cql/engine/elm/execution/FunctionRefEvaluator.java
+++ b/Src/java/engine/src/main/java/org/opencds/cqf/cql/engine/elm/execution/FunctionRefEvaluator.java
@@ -5,10 +5,15 @@ import java.util.Optional;
 
 import org.cqframework.cql.elm.execution.Expression;
 import org.cqframework.cql.elm.execution.FunctionDef;
+import org.cqframework.cql.elm.execution.FunctionRef;
 import org.opencds.cqf.cql.engine.execution.Context;
 import org.opencds.cqf.cql.engine.execution.Variable;
 
 public class FunctionRefEvaluator extends org.cqframework.cql.elm.execution.FunctionRef {
+
+    // For a given instance of the FunctionDefEvaluator, we'll only resolve to one FunctionDef
+    // Even if it's evaluated multiple times in a given ELM graph.
+    private FunctionDef functionDef;
 
     @Override
     protected Object internalEvaluate(Context context) {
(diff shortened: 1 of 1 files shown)


mdnazmulkarim (approved):

Src/java/engine/src/main/java/org/opencds/cqf/cql/engine/elm/execution/FunctionRefEvaluator.java
```diff
@@ -5,11 +5,16 @@
...
 import org.cqframework.cql.elm.execution.FunctionDef;
+import org.cqframework.cql.elm.execution.FunctionRef;
 import org.opencds.cqf.cql.engine.execution.Context;
 import org.opencds.cqf.cql.engine.execution.Variable;
 
 public class FunctionRefEvaluator extends org.cqframework.cql.elm.execution.FunctionRef {
 
+    // For a given instance of the FunctionDefEvaluator, we'll only resolve to one FunctionDef
+    // Even if it's evaluated multiple times in a given ELM graph.
+    private FunctionDef functionDef;
```
brynrhodes, line 16: I'm concerned that this might not always be correct, because the resolution is based on the runtime argument types. Consider:

```cql
define function TestAny(a Integer):
  a

define function TestAny(b String):
  b

define TestAnyFunctionWithInteger:
  TestAny(1)

define TestAnyFunctionWithString:
  TestAny('joe')
```

Currently, this works as expected because of the runtime resolution:

```cql
TestAnyFunctionWithInteger=1
TestAnyFunctionWithString=joe
```
The caching here would have to be expanded to include the resolved signature.

brynrhodes (changes requested):
This change would silently break current runtime behavior. We would need the cache to at least include the runtime signature that was resolved.

Src/java/engine/src/main/java/org/opencds/cqf/cql/engine/elm/execution/FunctionRefEvaluator.java
JPercival: Agreed that we need to make some adjustments to workaround some issues but _conceptually_ we _should_ be able to do this, right? The fact that we need to rely on runtime type resolution is actually an issue with the ELM generation. The 2 `FunctionRef` elements that are generated for the 2 separate calls to 2 separate overloads `TestAny` are indistinguishable and we can end up with ambiguous function resolutions.

The resolved signature is included in the `FunctionDesc` class that I added a couple PRs back. We could potentially use those but that doesn't actually save us from the most expensive part of the function evaluation, which is determining the types and compatibility of the function arguments via reflection. So, I think we come via a solution to address that.

Issue #838, opened 2022-11-07
modelinfo for qicore 5.0.0 needed


p9g:
https://github.com/cqframework/clinical_quality_language/tree/master/Src/java/quick/src/main/resources/org/hl7/fhir currently does not have a modelinfo for qicore 5.0.0.

See also https://jira.hl7.org/browse/FHIR-37579

JSRankins:
@brynrhodes, I believe the plan was to remove model info with Translator 2.x version. I know that hasn't happened yet. Was the plan to do that beginning with QI-Core 5.0.0?

PR #1124, opened 2023-02-07
#838: Added QICore modelinfo for 5.0.0

brynrhodes:


p9g:
Please take a look at https://github.com/projecttacoma/fqm-execution/issues/195 
Wondering if it could be a modelinfo issue and if the qicore 5.0.0 may have the same issue.
FHIR 4.0.1 modeinfo has `patientBirthDatePropertyName="birthDate.value">  ... <element name="birthDate" elementType="FHIR.date"/> `

but qicore 4.1.1 has `patientBirthDatePropertyName="birthDate" ... <element name="birthDate" elementType="System.Date"`

JPercival (approved):

c-schuler:
It looks like NutritionOrder is not in either 4.1.1 or 5.0.0. Is that expected?

c-schuler (approved):

PR #1139, opened 2023-03-27
use signature if available in engine when evaluate FunctionRefEval

mdnazmulkarim:
FunctionDef has types. FunctionRef can have both types/signature and operands. 
It is good to match types if present.
In case of function overload, FunctionRef must have type/signature information, otherwise it should prompt Exception
FunctionDef:
Screenshot 2023-03-29 at 4 39 09 PM
FunctionRef:
Screenshot 2023-03-29 at 4 40 54 PM


Src/java/engine-jackson/src/test/java/org/opencds/cqf/cql/engine/execution/TranslatorHelper.java
```diff
@@ -50,4 +53,21 @@ public Library translate(String file) throws UcumException, IOException {
 
         return new JsonCqlLibraryReader().read(new StringReader(json));
     }
+
+    public Library readJson(String json) throws IOException {
```
chalmarm, line 57: 1. Not a big one, but should we follow the hapi-fhir standards here too, if so can we rename the input parameter.
2. Can we also rename the method name to reflect its job. It is reading an input string to return a Library. Maybe readForLibrary, getLibrary, or readLibraryFromInputJson.

Src/java/engine-jackson/src/test/resources/org/opencds/cqf/cql/engine/execution/methodOverloadSignatureNone.json
```diff
@@ -0,0 +1,124 @@
+{
```
JPercival, line 1: suggestion: Could we commit the original CQL rather than the ELM? Then in the tests use the translation options for SignatureLevel to control how ELM is generated. That way we don't have to add new, otherwise unrelated code to the repository.

Src/java/engine/src/main/java/org/opencds/cqf/cql/engine/elm/execution/FunctionRefEvaluator.java
```diff
@@ -19,8 +21,15 @@ protected Object internalEvaluate(Context context) {
 
         boolean enteredLibrary = context.enterLibrary(this.getLibraryName());
         try {
-            // TODO: Use type specifiers from the operands here if they are available
-            FunctionDef functionDef = context.resolveFunctionRef(this.getName(), arguments, this.getLibraryName());
+
+            FunctionDef functionDef;
+            if (!this.getSignature().isEmpty()) {
+                List<Object> types = this.getSignature().stream().map(e -> (Object) e).collect(Collectors.toList());
```
JPercival, line 27: question: I don't understand what the cast to `Object` does here.

Src/java/engine/src/main/java/org/opencds/cqf/cql/engine/execution/Context.java
```diff
@@ -566,7 +566,11 @@ private FunctionDef resolveFunctionDesc(FunctionDesc functionDesc, List<Object>
         }
 
         for (var i = 0; i < arguments.size(); i++) {
-            isMatch = isType(resolveType(arguments.get(i)), operands.get(i));
+            if (arguments.get(i) instanceof NamedTypeSpecifier) {
```
JPercival, line 569: suggestion: There isn't enough information available to unambiguously resolve the operator, so the behavior should be to fail if overloads are detected and there's no type information.

JPercival (changes requested):

Issue #1126, opened 2023-02-09
Translator is not expanding mappings when unions of different FHIR resources occur in CQL


srankins:
The Translator is not expanding mappings when unions of different FHIR resources occur in CQL. Refer to https://github.com/projecttacoma/fqm-execution/issues/191.

mgramigna:
Minimal replication CQL:

``` cql
library UnionTest
 
using QICore version '4.1.1'

include FHIRHelpers version '4.1.000'

context Patient

define "Union of Different Types":
  ([Procedure] union [ServiceRequest]) R
    return {
      performed: R.performed,
      authoredOn: R.authoredOn
    }

```

brynrhodes:
Fixed in 2.7.0

mcmcphillips:
@brynrhodes 
After updating the translator to 2.7, the unintended behavior described in https://oncprojectracking.healthit.gov/support/browse/BONNIEMAT-1377 is still present.

PR #1140, opened 2023-03-28
#1126: Fixed incorrect mapping application when expanding mappings th…

brynrhodes:
…rough choice types


c-schuler (approved):
Looks good, builds and tests pass locally

mdnazmulkarim (approved):

PR #1150, opened 2023-04-12
Fixed FHIR Query Generation test cases that involved dates

sliver007:
Made the data calculation logic for the expected query calculation consistent with what the engine is doing for query generation to resolve an issue with test failing when an interval's start and end date are in different with respect to daylight savings time.

### DIFF
--- a/.github/workflows/check-pr.yml
+++ b/.github/workflows/check-pr.yml
@@ -15,3 +15,9 @@ jobs:
       with:
         arguments: check
         build-root-directory: Src/java
+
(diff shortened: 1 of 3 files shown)


brynrhodes (changes requested):
Getting test failures after updating from main branch

sliver007:
> Getting test failures after updating from main branch

Updated test cases and they are now passing

JPercival (approved):

brynrhodes (approved):

PR #1163, opened 2023-05-01
Bug model info hack

JPercival:
* Added some tests to verify signature generation on overloads
* Added a workaround for loading the current FHIR model when generating signatures for FhirHelpers
* Added the ability to conditionally cache the resolved function def in the CQL engine


brynrhodes (approved):
Agree it's a hack, but given where we are with the model info generation, it's functional.

Src/java/engine/src/main/java/org/opencds/cqf/cql/engine/execution/Context.java
```diff
@@ -617,7 +617,7 @@ private FunctionDef getResolvedFunctionDef(final String libraryName, final Strin
         }
 
         if (candidateDefs.size() > 1 && !hasSignature) {
-            throw new CqlException(String.format("Signature not provided for overloaded function '%s'", mangledFunctionName));
+            logger.debug("Using runtime function resolution for '{}'. It's recommended to always signatures in ELM", mangledFunctionName);
```
barhodes, line 620: Error message seems like it's missing some words.  "always provide signatures in ELM"?

JPercival: Whoops. Thanks.

barhodes (approved):

PR #1165, opened 2023-05-02
Engine as visitor

mdnazmulkarim:


brynrhodes (changes requested):
This looks fantastic, nice work! I like the rename of runtime context to State, and I like the introduction of the Environment. The only comment I have is that we should update the comment on the CqlEngine class to document that we switched from a deserialization approach to a visitor approach, since the documentation still says we're opting for the deserialization approach.

Src/java/engine/src/main/java/org/opencds/cqf/cql/engine/elm/visiting/FunctionRefEvaluator.java
```diff
@@ -0,0 +1,59 @@
...
+import org.hl7.elm.r1.FunctionRef;
+import org.opencds.cqf.cql.engine.execution.CqlEngine;
+import org.opencds.cqf.cql.engine.execution.State;
+import org.opencds.cqf.cql.engine.execution.Variable;
+
+import java.util.ArrayList;
+
+public class FunctionRefEvaluator {
+
+    private FunctionDef cachedFunctionDef;
```
JPercival, line 14: Making this an instance breaks the function caching. We should make this class static, and add a  `Map<FunctionRef, FunctionDef> functionCache` the `State` object. Then this evaluator should be able to use/update the state as needed to resolve the cache.

Src/java/engine/src/main/java/org/opencds/cqf/cql/engine/elm/visiting/QueryEvaluator.java
JPercival, line 1: This class can be used statically.

Src/java/engine/src/main/java/org/opencds/cqf/cql/engine/elm/visiting/ExpressionDefEvaluator.java
```diff
@@ -0,0 +1,42 @@
...
+
+            if (state.getCache().isExpressionCachingEnabled()) {
+                var er = new ExpressionResult(value, state.getEvaluatedResources());
+                state.getCache().cacheExpression(libraryId, expressionDef.getName(), er);
+            }
+
+            return value;
+
+        } catch (Exception e) {
+            visitor.processException(e, expressionDef);
```
JPercival, line 33: `processException` sneakily throws an exception. We should make that clear by calling it something like `throwExceptionWithSourceLocators` or by doing something like "throw addSourceLocatorsToException(e)"

Additionally, we should centralize the exception handling. It should only be done at the top-level `evaluate` method in the cql engine.

Src/java/engine/src/main/java/org/opencds/cqf/cql/engine/execution/CqlEngine.java
```diff
@@ -315,4 +372,1416 @@ private Set<String> getExpressionSet(Library library) {
...
+    }
+
+    @Override
+    public Object visitExpressionDef(ExpressionDef expressionDef, State state) {
+
+        //logger.info(String.format("expression def visit: %s", expressionDef.getName()));
+        return ExpressionDefEvaluator.internalEvaluate(expressionDef, state, this);
+    }
+
+    public Object validateOperand(Object operand) {
```
JPercival, line 404: It's unclear to me why we need this. Shouldn't `ExpressionDefs` be visited already when resolving the operand?

Src/java/engine/src/main/java/org/opencds/cqf/cql/engine/execution/CqlEngine.java
```diff
@@ -315,4 +372,1416 @@ private Set<String> getExpressionSet(Library library) {
...
+    }
+
+    @Override
+    public Object visitGreater(Greater elm, State state) {
+        try {
+            Object left = validateOperand(visitExpression(elm.getOperand().get(0), state));
+            Object right = validateOperand(visitExpression(elm.getOperand().get(1), state));
+
+            return GreaterEvaluator.greater(left, right, state);
+        } catch (Exception e) {
```
JPercival, line 575: we should move exception handling up to a central location. All the visit methods should be trivial.

Src/java/engine/src/main/java/org/opencds/cqf/cql/engine/elm/visiting/FunctionRefEvaluator.java
JPercival: I mean the `Cache` object, not the `State` object.

JPercival:
Approved. Let's address the remaining issues as part of new PRs.

brynrhodes (approved):
Overall it looks really good and I'm comfortable it's at least as functional as the previous engine based on regressions

PR #1182, opened 2023-08-06
Fixes for list Unions

### LINKED ISSUE #887: Union of lists with a null is incorrect
The following expression:
```
define Test: null union { 4, 5 } // Should be  // { 4, 5 }
```
Currently evaluates to `null`, but should evaluate to `{ 4, 5}`

Correct the issue and a test to cover the case.

Note carefully that union with intervals _does_ propagate nulls:
```
define Test: null union Interval[4, 5] // Should be null
```
Ensure that there is a test case covering this as well and that this fix does not break that test.

### LINKED ISSUE #1132: union/distinct operators do not match spec for lists when null elements are involved.
The specification for the union operator states (Appendix B, section 10.25): **This operator uses equality semantics _with the exception that null elements are considered equal._** Similar verbiage is found in the distinct operator (Appendix B, section 10.2). After finding issues in my code (which uses lists containing nulls), I ran two tests that checked these assertions:

define Test2: {{'A', 'B', null}, {'A', 'B', null}} union {'A', 'B', 'C'}
define Test3: distinct {{'A', 'B', null}, {'A', 'B', null}}

According to the defined semantics of the union and distinct operators, duplicate list objects should be removed and the outputs should be:
Test2=[[A, B, null], A, B, C]
Test3=[[A, B, null]]

However, the CQL evaluator gives the output:
Test2=[[A, B, null], [A, B, null], A, B, C]
Test3=[[A, B, null], [A, B, null]]

It appears as if the nulls are not being considered equal. There is no workaround to this, assuming that one is using lists containing nulls and wants the semantics described in the specification. I looked at the equality code for distinct and union and it appears as if when the RHS object is null, the code returns false immediately, rather than checking the LHS for a null value and returning true in this case.

JPercival:
* Correct erroneous tests for list Union behavior
  * According to the spec for List unions `null union {1, 2, 3}` should return `{1, 2, 3}` and not `null`
  * https://cql.hl7.org/09-b-cqlreference.html#union-1
* Engine fixes and additional tests for the above
  * Resolves #887
  * Partially resolves #1132 
* Added a link to the CQL spec in the Readme
* Fixed missing U of U copyright for the CQL engine

### DIFF
--- a/Src/java/engine/src/main/java/org/opencds/cqf/cql/engine/execution/CqlEngine.java
+++ b/Src/java/engine/src/main/java/org/opencds/cqf/cql/engine/execution/CqlEngine.java
@@ -2,6 +2,8 @@ package org.opencds.cqf.cql.engine.execution;
 
 import org.apache.commons.lang3.tuple.Pair;
 import org.cqframework.cql.elm.visiting.ElmBaseLibraryVisitor;
+import org.hl7.cql.model.IntervalType;
+import org.hl7.cql.model.ListType;
 import org.hl7.cql.model.NamespaceManager;
 import org.hl7.elm.r1.*;
 import org.hl7.elm.r1.Date;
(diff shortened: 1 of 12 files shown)


mdnazmulkarim (approved):

Src/java/engine/src/main/java/org/opencds/cqf/cql/engine/execution/CqlEngine.java
```diff
@@ -430,9 +432,23 @@ public Object visitUpper(Upper elm, State state) {
...
+        Object leftResult = visitExpression(left, state);
+        Object rightResult = visitExpression(right, state);
+
+
+        // This will attempt to use the declared result types from the ELM to determine which type of Union
+        // to perform. If the types are not declared, it will fall back to checking the values of the result.
+        if (left.getResultType() instanceof ListType || right.getResultType() instanceof ListType || elm.getResultType() instanceof ListType) {
+            return UnionEvaluator.unionIterable((Iterable<?>)leftResult, (Iterable<?>)rightResult, state);
+        }
+        else if (left.getResultType() instanceof IntervalType || right.getResultType() instanceof IntervalType || elm.getResultType() instanceof ListType) {
```
brynrhodes, line 446: Shouldn't this last `elm.getResultType() instanceof ListType` be IntervalType as well?

brynrhodes (changes requested):
Builds locally and tests pass, overall looks good, but I think there is a copy/paste error in the UnionEvaluator?

Src/java/engine/src/main/java/org/opencds/cqf/cql/engine/execution/CqlEngine.java
JPercival: Yep. You are right.

brynrhodes (approved):

Issue #1243, opened 2023-10-12
Change handling of potentially ambiguous method overloads


lukedegruchy:
See https://github.com/cqframework/clinical_quality_language/issues/1127

After this fix was merged, @brynrhodes suggested some fixes:

1. Leverage the OperatorMap to implement this check in `LibraryBuilder#resolveInvocation` instead of `validateAmbiguousOverloadedForwardDeclarationsSignatureNone()`, specifically, around the block that checks the SignatureLevel
2. Consider making this a warning and not an error
3. Consider checking this only if the function(s) in question are called, not if they merely exist in the CQL or in the library in question

brynrhodes:
Note that per some internal discussion, the implementation here is just to change this logic:

https://github.com/cqframework/clinical_quality_language/blob/master/Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/LibraryBuilder.java#L1176

So that in the case that we have resolved an overload with a different invocation signature than the declaration signature and the signatureLevel is None, that we issue a warning, not an error.

The warning should say "Function invocation has a different signature than the declaration and the SignatureLevel is set to none. This may result in ambiguous function resolution at runtime, consider setting the SignatureLevel to higher than None."

Specifically, I'm trying to make sure the message doesn't say anything about forwards, because it's not about forwards, this happens regardless of whether the function is a forward declaration.

Also, the default options in the translator were changed to SignatureLevel.Overloads, and we should switch that back to None.

JPercival:
Resolved by #1251

PR #1248, opened 2023-10-13
#1243: The translator will now return a warning if a function with ov…

brynrhodes:
…erloads is invoked when signature level is none

### DIFF
--- a/Src/java/cql-to-elm/src/test/java/org/cqframework/cql/cql2elm/fhir/dstu2/BaseTest.java
+++ b/Src/java/cql-to-elm/src/test/java/org/cqframework/cql/cql2elm/fhir/dstu2/BaseTest.java
@@ -273,7 +273,7 @@ public class BaseTest {
 
     @Test
     public void testFHIRNamespaces2() throws IOException {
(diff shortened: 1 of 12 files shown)


Src/java/cql-to-elm/src/test/java/org/cqframework/cql/cql2elm/fhir/dstu2/BaseTest.java
```diff
@@ -273,7 +273,7 @@ public void testFHIRNamespaces() throws IOException {
 
     @Test
     public void testFHIRNamespaces2() throws IOException {
```
lukedegruchy, line 275: Suggestion:  Rename this to testFHIRNamespaces_SignatureLevel_None()   I meant to do this on the PR where I introduced this.

lukedegruchy:
This looks good to me.  I think it's probably a good idea to get another member of the team to look at it, so I won't explicitly approve it.

JPercival (approved):

mdnazmulkarim (approved):

PR #1249, opened 2023-10-13
Tests to ensure generic overloads are sufficiently specified

JPercival:
* Adds tests to ensure that the compiler emits multiple levels of type specifiers when there are generic overloads.

### DIFF
--- a/Src/java/cql-to-elm/src/test/java/org/cqframework/cql/cql2elm/GenericOverloadsTests.java
+++ b/Src/java/cql-to-elm/src/test/java/org/cqframework/cql/cql2elm/GenericOverloadsTests.java
@@ -1,13 +1,11 @@
 package org.cqframework.cql.cql2elm;
 
 import org.cqframework.cql.cql2elm.LibraryBuilder.SignatureLevel;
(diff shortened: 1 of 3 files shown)


Src/java/cql-to-elm/src/test/java/org/cqframework/cql/cql2elm/GenericOverloadsTests.java
```diff
@@ -0,0 +1,112 @@
...
+    @Test
+    public void TestResultTypesSignatureNone() throws IOException {
+        Library library = getLibrary(true, SignatureLevel.None);
+
+        var stringifies = stringifies(library);
+        stringifies.forEach(this::validateResultTypes);
+    }
+
+    @Test
+    public void TestNoResultTypesSignatureNone() throws IOException {
```
lukedegruchy, line 106: For completeness I suggest doing the same for the other 2 signature types, and consider doing this as part a parameterized test.

lukedegruchy (approved):
Approved with one suggestion.

mdnazmulkarim (approved):

PR #1251, opened 2023-10-17
#1243: Fixed forward function overloads resolving incorrectly

brynrhodes:


Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/preprocessor/FunctionDefinitionInfo.java
```diff
@@ -19,12 +19,12 @@ public void setName(String value) {
         name = value;
     }
 
-    public void setPreCompileOutput(PreCompileOutput preCompileOutput) {
-        this.preCompileOutput = preCompileOutput;
+    public void setPreCompileOutput(FunctionHeader functionHeader) {
```
JPercival, line 22: Suggestion: This should be get/setFunctionHeader

Src/java/cql-to-elm/src/test/resources/org/cqframework/cql/cql2elm/fhir/r401/TestOverloadForward.cql
```diff
@@ -0,0 +1,18 @@
+library TestOverloadForward
+
+using FHIR version '4.0.1'
```
JPercival, line 3: Question: Do we need to consider the case of header-only declarations for an included library? I _think_ given the current compiler architecture it's expected that an included Library will be fully compiled but I'm not 100% certain.

Src/java/cql-to-elm/src/test/java/org/cqframework/cql/cql2elm/fhir/r401/BaseTest.java
```diff
@@ -831,4 +832,30 @@ public void testInclude() throws IOException {
         assertThat(qlr.getName(), equalTo("M"));
         assertThat(eqv.getOperand().get(1), instanceOf(ToConcept.class));
     }
+
+    @Test
+    public void testOverload() throws IOException {
+        CqlTranslator translator = TestUtils.runSemanticTest("fhir/r401/TestOverload.cql", 0);
```
JPercival, line 838: Suggestion: We should extend the tests to ensure that all functions are compiled, even if they aren't referenced. We now compile on-demand, so there may be dangling definitions that would "disappear" from the ELM.

JPercival (approved):
Approved with some questions/suggestions.

Src/java/cql-to-elm/src/test/resources/org/cqframework/cql/cql2elm/fhir/r401/TestOverloadForward.cql
brynrhodes: Yes, the expectation is that included libraries are fully compiled and the current translator will ensure that is the case. I tried to do lazy compilation even for included libraries, but it proved to be more trouble than it was worth at this point.

Src/java/cql-to-elm/src/test/java/org/cqframework/cql/cql2elm/fhir/r401/BaseTest.java
brynrhodes: We could construct a specific test, but I definitely encountered this issue in getting past the regression tests, I initially tried to leave some functions uncompiled, but reconstructing the state necessary to do nested compilation was too big a lift and I opted for ensuring the Library is 100% compiled instead.

mdnazmulkarim (approved):

Issue #648, opened 2021-06-24
Type inference through conditionals (case-when, if-else, etc)


JPercival:
If the statements of a conditional return disparate types the cql-translator throws an error because it can't correctly infer that it should be a choice type:

Doesn't work (expected return type is `Choice<String,Boolean>`):
```cql
define "Cases": 
  case
    when true then 'true'
    when false then false
    else null
  end
```
Doesn't work (expected return type is `Choice<String,Boolean>`):
```cql
  define "If":
    if true then 'true'
    else if false then false
    else null
```

As a workaround you can cast the first conditional to the expected type:
```cql
define "If":
    if true then ('true' as Choice<Boolean, String>)
    else if false then false
    else null
 ```

JPercival:
closed by #1252

PR #1252, opened 2023-10-18
#648: added tests for validating type inference through conditional

iasmile:
Added the separate test cases for case and if type conditionals. The returned result types are verified as ChoiceType for following snippets.

``` 
define "Cases": 
  case
    when true then 'true'
    when false then false
    else null
  end
```

```
  define "If":
    if true then 'true'
    else if false then false
    else null
```

Github issue: https://github.com/cqframework/clinical_quality_language/issues/648

### DIFF
--- a/Src/java/cql-to-elm/src/test/java/org/cqframework/cql/cql2elm/SemanticTests.java
+++ b/Src/java/cql-to-elm/src/test/java/org/cqframework/cql/cql2elm/SemanticTests.java
@@ -6,9 +6,7 @@ import org.hl7.cql.model.NamedType;
 import org.hl7.elm.r1.*;
 import org.testng.annotations.Test;
 
-import java.io.ByteArrayOutputStream;
 import java.io.IOException;
-import java.io.PrintStream;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
@@ -17,7 +15,6 @@ import java.util.Map;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.*;
 import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertSame;
 import static org.junit.Assert.assertTrue;
 import static org.testng.Assert.assertThrows;
 
--- a/Src/java/cql-to-elm/src/test/java/org/cqframework/cql/cql2elm/SemanticTests.java
+++ b/Src/java/cql-to-elm/src/test/java/org/cqframework/cql/cql2elm/SemanticTests.java
@@ -6,7 +6,10 @@ import org.hl7.cql.model.NamedType;
 import org.hl7.elm.r1.*;
 import org.testng.annotations.Test;
 
+import java.io.ByteArrayOutputStream;
 import java.io.IOException;
+import java.io.PrintStream;
+import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -14,8 +17,10 @@ import java.util.Map;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.*;
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertSame;
 import static org.junit.Assert.assertTrue;
 import static org.testng.Assert.assertThrows;
+
 
 public class SemanticTests {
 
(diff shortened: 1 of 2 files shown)


Src/java/cql-to-elm/src/test/java/org/cqframework/cql/cql2elm/SemanticTests.java
```diff
@@ -652,6 +653,62 @@ public void testNonExistentFileName() {
...
+        assertThat(caseDef.getResultType(), instanceOf(ChoiceType.class));
+
+        ChoiceType choiceType = (ChoiceType)caseDef.getResultType();
+
+        List<String> expectedChoiceTypes = new ArrayList<>();
+        expectedChoiceTypes.add("System.String");
+        expectedChoiceTypes.add("System.Boolean");
+        expectedChoiceTypes.add("System.Integer");
+
+        for (DataType dt : choiceType.getTypes()) {
```
JPercival, line 679: Suggestion: This requirement goes both ways. I'd suggest creating two sets of strings for expected and actual and making sure they are equivalent. This test would start passing if the compiler emitted no choice types.

Src/java/cql-to-elm/src/test/java/org/cqframework/cql/cql2elm/SemanticTests.java
```diff
@@ -652,6 +653,62 @@ public void testNonExistentFileName() {
...
+        assertThat(ifDef.getResultType(), instanceOf(ChoiceType.class));
+
+        ChoiceType choiceType = (ChoiceType)ifDef.getResultType();
+
+        List<String> expectedChoiceTypes = new ArrayList<>();
+        expectedChoiceTypes.add("System.String");
+        expectedChoiceTypes.add("System.Boolean");
+
+        for (DataType dt : choiceType.getTypes()) {
+                assertTrue("Expected return types are String and Boolean: ", expectedChoiceTypes.contains(((NamedType)dt).getName()));
```
JPercival, line 707: Suggestion: Same as above here. The set of expected types should be equivalent to the set of actual types

JPercival (changes requested):

Src/java/cql-to-elm/src/test/java/org/cqframework/cql/cql2elm/SemanticTests.java
iasmile: The sets comparison does not respect the order of types I believe using list comparison would harden the test case. suggestion?

iasmile: The requested changes have been made.

Src/java/cql-to-elm/src/test/java/org/cqframework/cql/cql2elm/SemanticTests.java
iasmile: The requested change have been made.

Src/java/cql-to-elm/src/test/java/org/cqframework/cql/cql2elm/SemanticTests.java
JPercival: Set is the correct comparison here, not List.

Choice<Integer, String> ~ Choice<String,Integer>

JPercival (changes requested):

JPercival (approved):

Src/java/cql-to-elm/src/test/java/org/cqframework/cql/cql2elm/SemanticTests.java
```diff
@@ -652,6 +654,64 @@ public void testNonExistentFileName() {
...
+        for (DataType dt : choiceType.getTypes()) {
+            actualChoiceTypes.add(((NamedType)dt).getName());
+        }
+        assertTrue("Expected types are String, Boolean, and Integer: ", actualChoiceTypes.equals(expectedChoiceTypes));
+    }
+
+    @Test
+    public void testIfConditionalReturnTypes() throws IOException {
+        CqlTranslator translator = runSemanticTest("Issue648.cql", 0);
+        org.hl7.elm.r1.Library library = translator.toELM();
```
mdnazmulkarim, line 690: There's no other Library class, thus no need to use package reference here hopefully.

Src/java/cql-to-elm/src/test/java/org/cqframework/cql/cql2elm/SemanticTests.java
```diff
@@ -652,6 +654,64 @@ public void testNonExistentFileName() {
...
+        Set<String> actualChoiceTypes = new HashSet<>();
+        for (DataType dt : choiceType.getTypes()) {
+            actualChoiceTypes.add(((NamedType)dt).getName());
+        }
+        assertTrue("Expected types are String, Boolean, and Integer: ", actualChoiceTypes.equals(expectedChoiceTypes));
+    }
+
+    @Test
+    public void testIfConditionalReturnTypes() throws IOException {
+        CqlTranslator translator = runSemanticTest("Issue648.cql", 0);
```
mdnazmulkarim, line 689: You can avoid translating the Library twice

mdnazmulkarim (approved):

PR #1267, opened 2023-10-31
Test to validate 827 is fixed

### LINKED ISSUE #827: Identifier containing escaped double-quotes does not resolve
References to a CQL expression identifier that contains escaped double quotes, will not resolve unless the definition is above/before the reference. Rearranging the CQL seems to be a possible workaround, but will not work in our particular case - we are using automation that generates the CQL as part of a transform from a proprietary representation. 

# Reproduction

###### Machine/OS
MacBook Pro M1 Max (ARM) chip.  
macOS Monterey 12.0.1

###### IDE
Visual Studio Code
Version: 1.72.2  
  
Extension: Clinical Quality Language (CQL)  
version 0.4.0 'Preview'

This following code fails with the exception: 
"org.opencds.cqf.cql.engine.exception.CqlException: Translation of library Test failed with the following message: Could not resolve identifier This is the "referenced" expression in the current library.
	at org.opencds.cqf.cql.evaluator.engine.execution.TranslatingLibraryLoader.translate(TranslatingLibraryLoader.java:134)..."

```
library Test

using FHIR version '4.0.1'

context Patient

define "Referencing Expression":     
    "This is the \"referenced\" expression"

define "This is the \"referenced\" expression":
    'The reference resolved!'
```

The following code works (note the order of the two expressions is modified such that the referenced expression is defined before it is referenced:

```
library Test

using FHIR version '4.0.1'

context Patient

define "This is the \"referenced\" expression":
    'The reference resolved!'

define "Referencing Expression":     
    "This is the \"referenced\" expression"
```

JPercival:
Closes #827

### DIFF
--- a/Src/java/cql-to-elm/src/test/java/org/cqframework/cql/cql2elm/SemanticTests.java
+++ b/Src/java/cql-to-elm/src/test/java/org/cqframework/cql/cql2elm/SemanticTests.java
@@ -503,6 +503,12 @@ public class SemanticTests {
         assertThat(ivs.getValuesetExpression(), instanceOf(OperandRef.class));
     }
 
+    @Test
+    public void testIssue827() throws IOException {
+        // https://github.com/cqframework/clinical_quality_language/issues/827
+        TestUtils.runSemanticTest("Issue827.cql", 0);
+    }
+
 
     @Test
     public void testIssueEmptySourceInterval() throws IOException {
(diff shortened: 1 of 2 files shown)


Src/java/cql-to-elm/src/test/java/org/cqframework/cql/cql2elm/SemanticTests.java
```diff
@@ -667,7 +673,7 @@ public void testCaseConditionalReturnTypes() throws IOException {
         }
 
         ExpressionDef caseDef = defs.get("Cases");
-    
+
```
lukedegruchy, line 676: Nitpick:  this change isn't needed

lukedegruchy (approved):

mdnazmulkarim (approved):

PR #1269, opened 2023-11-01
First commit.

lukedegruchy:


Src/java/engine-fhir/src/test/java/org/opencds/cqf/cql/engine/fhir/converter/R4TypeConverterTests.java
```diff
@@ -2,18 +2,16 @@
...
 import static org.testng.Assert.assertTrue;
 
 import java.math.BigDecimal;
-import java.time.ZoneOffset;
-import java.util.ArrayList;
-import java.util.Calendar;
-import java.util.Iterator;
-import java.util.List;
-import java.util.TimeZone;
+import java.time.*;
```
c-schuler, line 12: Just noting the use of wildcard imports here, which is not best practice.

Src/java/engine/src/main/java/org/opencds/cqf/cql/engine/runtime/DateTime.java
```diff
@@ -296,10 +294,8 @@ public Date toJavaDate() {
         return java.util.Date.from(dateTime.toInstant());
     }
 
-    public static DateTime fromJavaDate(Date date) {
```
c-schuler, line 299: This is a very useful function that is used in other projects. I would like to know why it has been removed.

c-schuler (approved):
Nice work. I have a couple of notes, but am willing to approve. I would like to see the fromJavaDate function restored unless there is a good reason for its removal.

Src/java/engine/src/main/java/org/opencds/cqf/cql/engine/runtime/DateTime.java
lukedegruchy: The IDE said it was unused and it's my instinct to clean up dead code.   

Feel free to add it back.

JPercival: I did a quick review of cqf-tooling and clinical-reasoning, this function is used when parsing the input reporting period. So it's only used one place I'm fine either way.

Issue #1254, opened 2023-10-20
Invalid CQL Parsing when missing comma in List construction


MikeRileyGTRI:
I believe this only happens with List types getting concatenated together in the expression. See this example below, it only seems to happen with List types. Issue is with "ListTest6"

```
// "TreatmentConcept_X" Concepts described above
define "TreatmentMR_0": [MedicationRequest: "TreatmentConcept_0"]
define "TreatmentMR_1": [MedicationRequest: "TreatmentConcept_1"]
define "TreatmentMR_2": [MedicationRequest: "TreatmentConcept_2"]
define "TreatmentMR_3": [MedicationRequest: "TreatmentConcept_3"]
define "TreatmentMR_4": [MedicationRequest: "TreatmentConcept_4"]

define "ListTest1": {1,2,3,4,5} //Valid
define "ListTest2": {1,2,3,4,5,} //Invalid, errant comma
define "ListTest3": {1,2,3 4,5} //Invalid, missing comma
define "ListTest4": {'A','B','C','D','E'} //Valid
define "ListTest5": {'A','B','C' 'D','E'} //Invalid missing comma
define "ListTest6": {"TreatmentMR_0", "TreatmentMR_1", "TreatmentMR_2", "TreatmentMR_3" "TreatmentMR_4"} //Valid?? Missing comma
define "ListTest7": {"TreatmentMR_0", "TreatmentMR_1", "TreatmentMR_2", "TreatmentMR_3", "TreatmentMR_4",} //Invalid, errant comma
```

ddieppois:
I can confirm that it seems that the translator needs a correct syntax for the list definition. If I try with the given examples I get a set of errors related to the incorrect lists. If I remove or correct them it goes through without triggering those errors, just a few related to TreatmentConcept_X.

With wrong syntax for lists:
`Error:[22:31, 22:31] Syntax error at }
Error:[23:18, 23:18] Syntax error at :
Error:[23:22, 23:22] Syntax error at ,
Error:[23:24, 23:24] Syntax error at ,
Error:[23:28, 23:28] Syntax error at ,
Error:[23:30, 23:30] Syntax error at }
Error:[24:18, 24:18] Syntax error at :
Error:[25:33, 25:33] Syntax error at 'D'
Error:[27:105, 27:105] Syntax error at }
`

Nothing when removed.

MikeRileyGTRI:
[...]

Issue #721, opened 2022-02-01
Add a warning for duplicate aliases in peer contexts


brynrhodes:
To avoid potential confusion for readers, the translator should warn about reuse of the same alias identifier in peer contexts.

This could potentially be identified as a best practice as well for CQL authors. Also logged a tracker to identify this (along with alias hiding): https://jira.hl7.org/browse/FHIR-35916

Issue #426, opened 2019-06-07
Add a warning to the translator when identifier hiding is detected


brynrhodes:
Warn authors when they create an identifier that hides an identifier in a broader scope.

p9g:
For example, if an alias is the same as a definition name.

Issue #598, opened 2021-02-09
Add a "warning" when case-mismatching an identifier


brynrhodes:
If an identifier does not resolve but differs only in case from an existing identifier, include a mention of the case-insensitive match as a potential resolution.

JP: A failed resolution is a compile-time error, so the implementation of this story needs to enhance that error message. The current message doesn't include any message to potentially matching identifiers.

Luke:  The solution is to use conditional logic to select the error message depending on whether the identifiers were a case-insensitive match or didn't match at all.

lukedegruchy:
Related or potentially related tickets:

https://github.com/cqframework/clinical_quality_language/issues/426
https://github.com/cqframework/clinical_quality_language/issues/607
https://github.com/cqframework/clinical_quality_language/issues/721
https://github.com/cqframework/clinical_quality_language/issues/1254

PR #1277, opened 2023-11-09
Full implementation of detection of identifier hiding

lukedegruchy:
- Detect identifier hiding for various structures, including expressions, functions, parameters, arguments, lets, aliases, etc
- Detect such hiding for both exact matches and case insensitive matches
- Output compiler warnings referencing the structure in question
- Ensure the context of the structure is considered.  For example, remove from consideration an alias once the function body has been compiled
- Tons of new tests and some updated tests.

Closes https://github.com/cqframework/clinical_quality_language/issues/426
Closes https://github.com/cqframework/clinical_quality_language/issues/721
Closes https://github.com/cqframework/clinical_quality_language/issues/1254


Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/LibraryBuilder.java
```diff
@@ -2935,6 +3043,60 @@ private DataType getExpressionDefResultType(ExpressionDef expressionDef) {
...
+     * For example, if an alias within an expression body has the same name as a parameter, execution would have
+     * added the parameter identifier and the next execution would consider an alias with the same name, thus resulting
+     * in a warning.
+     * <p/>
+     * Exact case matching as well as case-insensitive matching are considered.  If known, the type of the structure
+     * in question will be considered in crafting the warning message, as per the {@link Element} parameter.
+     *
+     * @param identifier The identifier belonging to the parameter, expression, function, alias, etc, to be evaluated.
+     * @param onlyOnce Special case to deal with overloaded functions, which are out scope for hiding.
+     * @param element The consturct element, for {@link ExpressionRef}.
```
JPercival, line 3059: nit: consturct -> construct

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/Cql2ElmVisitor.java
```diff
@@ -514,6 +521,12 @@ private void removeImplicitContextExpressionDef(ExpressionDef def) {
     public ExpressionDef internalVisitExpressionDefinition(cqlParser.ExpressionDefinitionContext ctx) {
         String identifier = parseString(ctx.identifier());
         ExpressionDef def = libraryBuilder.resolveExpressionRef(identifier);
+
+        // lightweight ExpressionDef to be used to output a hiding warning message
+        final ExpressionDef hallowExpressionDef = of.createExpressionDef()
```
JPercival, line 526: nit: hallow -> hollow

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/LibraryBuilder.java
```diff
@@ -2344,6 +2344,114 @@ public Expression resolveIdentifier(String identifier, boolean mustResolve) {
         return null;
     }
 
+    public QueryLetRef getQueryLetRef(LetClause let) {
+        QueryLetRef result = of.createQueryLetRef().withName(let.getIdentifier());
```
JPercival, line 2348: suggestion: most of the XYZ ref factory functions could be reused directly inline in the library builder such as here:
https://github.com/cqframework/clinical_quality_language/blob/b1b2854f80fdb5b0d342cccc92bb4a51058223aa/Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/LibraryBuilder.java#L2232

Could also take that as a second PR to avoid mixing refactoring into this one.

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/LibraryBuilder.java
```diff
@@ -2344,6 +2344,114 @@ public Expression resolveIdentifier(String identifier, boolean mustResolve) {
...
+        codesystemRef.setResultType(codeSystemDef.getResultType());
+        if (codesystemRef.getResultType() == null) {
+            // ERROR:
+            throw new IllegalArgumentException(String.format("Could not validate reference to codesystem %s because its definition contains errors.",
+                    codesystemRef.getName()));
+        }
+        return null;
+    }
+
+    public ParameterRef resolveParameterRef(ParameterDef theParameterDef) {
```
JPercival, line 2393: suggestion: for the sake of naming consistency this ought to be `createParameterRef`

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/LibraryBuilder.java
```diff
@@ -2344,6 +2344,114 @@ public Expression resolveIdentifier(String identifier, boolean mustResolve) {
...
+        final ParameterRef parameterRef = of.createParameterRef().withName(theParameterDef.getName());
+        parameterRef.setResultType(parameterRef.getResultType());
+        return parameterRef;
+    }
+
+    public OperandRef resolveOperandRef(OperandDef operandDef) {
```
JPercival, line 2399: suggestion: same as above resolve -> create

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/LibraryBuilder.java
```diff
@@ -2935,6 +3043,60 @@ private DataType getExpressionDefResultType(ExpressionDef expressionDef) {
...
+     * added the parameter identifier and the next execution would consider an alias with the same name, thus resulting
+     * in a warning.
+     * <p/>
+     * Exact case matching as well as case-insensitive matching are considered.  If known, the type of the structure
+     * in question will be considered in crafting the warning message, as per the {@link Element} parameter.
+     *
+     * @param identifier The identifier belonging to the parameter, expression, function, alias, etc, to be evaluated.
+     * @param onlyOnce Special case to deal with overloaded functions, which are out scope for hiding.
+     * @param element The consturct element, for {@link ExpressionRef}.
+     * @param nullableExpression Use strictly to comply with the signature for {@link #reportWarning(String, Expression)}.
```
JPercival, line 3060: issue: I think we can reuse the existing Defs (which are Elements) rather than creating Refs (which are Expressions). The means we don't need the final parameter for this function. We can simply relax the signature of reportWarning from `Expression` -> `Element`. The trackbacks are available there.

JPercival (changes requested):

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/LibraryBuilder.java
lukedegruchy: After changing the signature to accept Elements, I got rid of all the code I introduced to deal with Expressions, including the "get"/"resolve" methods you commented on above.

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/LibraryBuilder.java
lukedegruchy: I deleted this method since I no longer need to deal with Expressions.

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/LibraryBuilder.java
lukedegruchy: I deleted this method since I no longer need to deal with Expressions.

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/LibraryBuilder.java
lukedegruchy: I deleted this method since I no longer need to deal with Expressions.

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/Cql2ElmVisitor.java
lukedegruchy: Done

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/LibraryBuilder.java
lukedegruchy: Done

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/LibraryBuilder.java
```diff
@@ -2935,6 +2985,58 @@ private DataType getExpressionDefResultType(ExpressionDef expressionDef) {
...
+     * <p/>
+     * Exact case matching as well as case-insensitive matching are considered.  If known, the type of the structure
+     * in question will be considered in crafting the warning message, as per the {@link Element} parameter.
+     *
+     * @param identifier The identifier belonging to the parameter, expression, function, alias, etc, to be evaluated.
+     * @param onlyOnce   Special case to deal with overloaded functions, which are out scope for hiding.
+     * @param element    The construct element, for {@link ExpressionRef}.
+     */
+    void pushIdentifierForHiding(String identifier, boolean onlyOnce, Element element) {
+        final MatchType matchType = identifiersToCheckForHiding.stream()
```
brynrhodes, line 3004: I am a little concerned about performance here. For library-level identifier resolution a hash is used currently. If I'm reading this correctly, every identifier declaration anywhere will now incur a worst-case performance of O(3n) in the number of identifiers in the library. Since that hit is on every identifier declaration, that will add up quickly. For most libraries, it probably doesn't matter, but I'm concerned about libraries with large numbers of declarations, and/or lots of content. It would be pretty straightforward to recast this as a hash, even for the case-insensitive comparison, right? Or am I being overly cautious here?

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/LibraryBuilder.java
```diff
@@ -2935,6 +2985,58 @@ private DataType getExpressionDefResultType(ExpressionDef expressionDef) {
...
+                    return MatchType.NONE;
+                })
+                .filter(innerMatchType -> MatchType.NONE != innerMatchType)
+                .findFirst()
+                .orElse(MatchType.NONE);
+
+        if (MatchType.NONE != matchType && ! onlyOnce) {
+            final String message = String.format("Identifier hiding detected: Identifier for identifiers: %s%s",
+                    String.format(lookupElementWarning(element), identifier),
+                    formatMatchedMessage(matchType)+ "\n");
```
brynrhodes, line 3023: Should this be `%n` instead of `\n`?

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/LibraryBuilder.java
```diff
@@ -2935,6 +2985,58 @@ private DataType getExpressionDefResultType(ExpressionDef expressionDef) {
...
+                    }
+
+                    return MatchType.NONE;
+                })
+                .filter(innerMatchType -> MatchType.NONE != innerMatchType)
+                .findFirst()
+                .orElse(MatchType.NONE);
+
+        if (MatchType.NONE != matchType && ! onlyOnce) {
+            final String message = String.format("Identifier hiding detected: Identifier for identifiers: %s%s",
```
brynrhodes, line 3021: I'm a little concerned about the message here, can we make this something more friendly, like:

The alias identifier X is hiding the parameter identifier X

And only do this for the exact match case (because technically case-insensitive matches are not hiding, right?)

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/LibraryBuilder.java
```diff
@@ -2935,6 +2985,58 @@ private DataType getExpressionDefResultType(ExpressionDef expressionDef) {
...
+     */
+    void pushIdentifierForHiding(String identifier, boolean onlyOnce, Element element) {
+        final MatchType matchType = identifiersToCheckForHiding.stream()
+                .map(innerIdentifier -> {
+                    if (innerIdentifier.equals(identifier)) {
+                        return MatchType.EXACT;
+                    }
+
+                    if (innerIdentifier.equalsIgnoreCase(identifier)) {
+                        return MatchType.CASE_IGNORED;
```
brynrhodes, line 3011: I'm not sure this is the right implementation, a case-insensitive match isn't technically hiding. The intent of adding a warning for a case-insensitive match was actually on identifier resolution when there wasn't an exact resolution, but there is a case-insensitive resolution, so we can say something friendly like "did you mean X instead of x here?"

brynrhodes (changes requested):
Overall this looks really good, and I love all the tests, I just have some comments inline that I wonder about.

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/LibraryBuilder.java
JPercival: I think it's probably overly cautious, myself.

The reason for that is that the alternative still requires a Deque or LinkedList or some other such construct to represent identifier scoping (for parameters, aliases, and lets). A hash alone won't cut it because we'd need to pop the identifiers off the stack once they fall out of scope. Best case is a linked list of hashes, which all need to be checked.

That would actually be the same construct we use at runtime in the cql engine to represent the stack frame, and if it's "fast enough" at runtime it ought to be "fast enough" at compile time, IMO.

I'm also not sure where O(3n) comes from. Could you elaborate on that? I think it's actually worse than that, O(N^2) (check each identifier against every other identifier).

We don't have any benchmarking for the compiler in the project, only the engine. It'd be interesting to add a few tests and see where our hotspots are. How about we add a story to include some benchmarks, and we address whatever hotspots there may be? If it's this particular item we change it, if it turns out to be something else we can change that.

JPercival (approved):

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/LibraryBuilder.java
lukedegruchy: My branch now outputs two distinct messages depending on whether the hiding is for an exact or case-sensitive match.

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/LibraryBuilder.java
lukedegruchy: My branch now outputs two distinct messages depending on whether the hiding is for an exact or case-sensitive match.

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/LibraryBuilder.java
lukedegruchy: Done.

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/LibraryBuilder.java
brynrhodes: I'm still not sure this is right, I don't think the compiler should output a warning if I just create an identifier that differs only in case, that is an intentional feature of the language and shouldn't cause a warning.

The warning should be output _only_ if a failed resolution differs from an in context identifier only by case.

In other words, this logic is running when the identifier is declared, the warning should be output on identifier resolution, not identifier declaration. Am I missing something here?

JPercival: Yes, that behavior is different than originally specified in the story as I read it. The scenario you describe would be an _error_ since it’d be a non-existent identifier. A failed resolution is a failed resolution regardless of casing as far as compilation success goes. What you’re suggesting could be implemented as a more precise error message if resolution failed (e.g. “Could not find identifier X, did you mean x?”) but it’d still be an error.  

@lukedegruchy - can we remove the case-sensitivity aspect and put that in a separate story?

lukedegruchy: @JPercival Agreed.

Do you think this issue covers the requirement?  https://github.com/cqframework/clinical_quality_language/issues/598

brynrhodes: Yes, that is the story, it is incorrectly characterized there as a warning, which is the source of the confusion here, my apologies for mis-stating that in the original story definition, but yes, I think that story is the one and no need to create a separate story for this.

JPercival: I think #598 is updated to reflect that it should be a more precise error message, and then this PR does not cover that story. We’ll have a follow-on PR to cover the #598 issue separately. Does that make sense?

lukedegruchy: I've remove the case insenstive logic and adjusted the tests accordingly.

JPercival:
In review we split out a couple of additional stories here:
#598 
#1280

PR #1282, opened 2023-11-17
Use conditional logic to select the error message for an unmatched identifier based on whether it was a case-insensitive match

lukedegruchy:
Closes https://github.com/cqframework/clinical_quality_language/issues/598


lukedegruchy:
Superseded by:  https://github.com/cqframework/clinical_quality_language/pull/1406

PR #1292, opened 2023-11-24
Extend ObjectFactory to include a decoration function

JPercival:
* Create `ObjectFactoryEx` that extends `ObjectFactory` and adds a `decorate` method that assigns local ids to all `Element`s created.
* Refactor ELM visitors and builders to share the ObjectFactory.


JPercival:
Planning on cleaning up a bit more, just a draft PR to check I'm on the right track.

JPercival:
Changed my mind on the approach here.

PR #1370, opened 2024-06-20
Null handling by ProperContains, ProperIncludedIn, ProperIncludes, ProperIn evaluators. More tests for arithmetic functions.

antvaset:
This fixes how the ProperContains, ProperIncludedIn, ProperIncludes, ProperIn evaluators handle nulls. The CQL specification states that:

* The `T properly included in List<T>` expression (`ProperIn(T, List<T>)` in ELM) evaluates to null if either argument is null ([link](https://github.com/HL7/cql/blob/v1.5.2/spec/09-b-cqlreference.adoc?plain=1#L5524)).
* Same for `List<T> properly included in List<T>` (`ProperIncludedIn(List<T>, List<T>)` in ELM).
* `List<T> properly includes T` (`ProperContains(List<T>, T)` in ELM) evaluates to null if either argument is null ([link](https://github.com/HL7/cql/blob/v1.5.2/spec/09-b-cqlreference.adoc?plain=1#L5492)).
* Same for `List<T> properly includes List<T>` (`ProperIncludes(List<T>, List<T>)` in ELM).

Also adding new tests for arithmetic functions.


JPercival:
@antvaset - We have a goal of >=80% test coverage for new/updated code. Can you add tests for the Evaluators that gets the patch up to 80%?

antvaset:
Hi @JPercival, I've squeezed in some more coverage. This is the most we can do I believe.

JPercival (approved):

brynrhodes (changes requested):
The new ArithmeticFunctionsTests are good, but the changes to the "proper" operators are incorrect. List and interval membership operations throughout the language use "equality semantics, with the exception that nulls are considered equal for the purposes of membership". See the "ProperContains" and "ProperIn" operator descriptions in the Logical Specification: https://cql.hl7.org/04-logicalspecification.html#proper-contains

antvaset:
Thanks for the clarification @brynrhodes. I'm closing this because the fixes to the evaluator logic are incorrect. Also linking here the issue in JIRA https://jira.hl7.org/browse/FHIR-46283 and the pull request with the salvaged arithmetic function tests https://github.com/cqframework/cql-tests/pull/33/files.

PR #1374, opened 2024-06-28
Reporting coverage across subprojects

antvaset:
Currently, the JaCoCo test reports which are run separately for the engine-fhir, engine, cql-to-elm subprojects only include the classes from the default source sets. This means that e.g. the coverage of cql-to-elm classes is not reported when running engine-fhir tests. Concrete example: `resolveMethod` in the cql-to-elm subproject has low coverage ([link](https://app.codecov.io/gh/cqframework/clinical_quality_language/commit/e00c39673d7d13def60debe112bc19ad7f48f297/blob/Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/SystemMethodResolver.java#L253)) even though there are many tests for FHIRPath methods inside engine-fhir ([link](https://github.com/cqframework/clinical_quality_language/blob/e00c39673d7d13def60debe112bc19ad7f48f297/Src/java/engine-fhir/src/test/resources/org/hl7/fhirpath/r4/tests-fhir-r4.xml)).

Here I'm trying to get code coverage to be reported across subprojects and see if codecov can aggregate the reports which seems to be the case. See coverage for `resolveMethod` after changing `jacocoTestReport` config ([link](https://app.codecov.io/gh/cqframework/clinical_quality_language/commit/f76f939dac2b357009261bac8b8db0e36d298729/blob/Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/SystemMethodResolver.java#L253)) and overall change in coverage for the affected classes ([link](https://app.codecov.io/gh/cqframework/clinical_quality_language/pull/1374/indirect-changes)).

### DIFF
--- a/Src/java/engine-fhir/build.gradle
+++ b/Src/java/engine-fhir/build.gradle
@@ -27,12 +27,15 @@ generateSources {
 
 jacocoTestReport {
     dependsOn ':cql-to-elm:compileTestJava'
+    dependsOn ':engine:compileTestJava'
 
     sourceDirectories.setFrom(files(
         "${projectDir}/../cql-to-elm/src/main/java",
+        "${projectDir}/../engine/src/main/java",
     ))
 
     classDirectories.setFrom(files(
         "${projectDir}/../cql-to-elm/build/classes/java/main",
+        "${projectDir}/../engine/build/classes/java/main",
     ))
 }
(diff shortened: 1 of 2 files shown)


JPercival (approved):
I agree with this change since it better reflects our actual test coverage. I'll just note that it suggests that our test structure is suboptimal, since we need to includes results across several projects to get the real result.

barhodes (approved):

PR #1385, opened 2024-07-18
[do not merge!] Align CQL's grammar with that of FHIRPath to fix operator precedence inconsistencies

antvaset:
This branch contains proposed changes to CQL's grammar (and therefore spec) to make operator precedence work the same way it does in FHIRPath.

## ...some background on this issue

The following is a valid FHIRPath expression:
```
(0 | 1 | 2).take(2) = 0 | 1
```
which evaluates to true (you can use this https://hl7.github.io/fhirpath.js/ to confirm). But in CQL it currently throws an error because `=` takes precedence over `|`, so it's trying to parse `((0 | 1 | 2).take(2) = 0) | 1` and not `(0 | 1 | 2).take(2) = (0 | 1)`. The pipe operator is defined to be equivalent to `union`, and `union` is currently explicitly below `=` in CQL's operator precedence table (https://cql.hl7.org/03-developersguide.html#operator-precedence). In FHIRPath however, the pipe has higher precedence than `=` (https://build.fhir.org/ig/HL7/FHIRPath/#operator-precedence).


antvaset:
Superseded by https://github.com/cqframework/clinical_quality_language/pull/1436

PR #1391, opened 2024-07-25
Implement ToChars in the engine

antvaset:
This PR implements the ToChars operator as defined by https://cql.hl7.org/04-logicalspecification.html#tochars:

> The ToChars operator takes a string and returns a list with one string for each character in the input, in the order in which they appear in the string.
> 
> If the argument is null, the result is null.

Squeezed in as much coverage as possible with 4 additional tests.

### DIFF
--- a/Src/java/engine/src/main/java/org/opencds/cqf/cql/engine/elm/executing/ToCharsEvaluator.java
+++ b/Src/java/engine/src/main/java/org/opencds/cqf/cql/engine/elm/executing/ToCharsEvaluator.java
@@ -1,0 +1,35 @@
+package org.opencds.cqf.cql.engine.elm.executing;
+
+import org.opencds.cqf.cql.engine.exception.InvalidOperatorArgument;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/*
+ToChars(argument String) List<String>
+
+The ToChars operator takes a string and returns a list with one string for each character in the input, in the order in which they appear in the string.
+
+If the argument is null, the result is null.
+*/
+
+public class ToCharsEvaluator {
+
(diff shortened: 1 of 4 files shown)


Src/java/engine/src/main/java/org/opencds/cqf/cql/engine/elm/executing/ToCharsEvaluator.java
```diff
@@ -0,0 +1,35 @@
...
+
+        if (operand instanceof String) {
+            List<String> result = new ArrayList<>();
+            for (char c : ((String) operand).toCharArray()) {
+                result.add(String.valueOf(c));
+            }
+            return result;
+        }
+
+        throw new InvalidOperatorArgument(
```
JPercival, line 31: thought: the error path isn't tested, which is a common issue with many of the `XYZEvaluator` classes. I can think of a few ways to address this:

1. Add a set of functions to check function invariants at the start of the body. This would add overhead to each call but it would allow us to more easily test that behavior.
2. Add Elm type inference / checking as part of the initial load of an ELM graph in the engine, (cache the results). This would allow us to remove all the error checking completely since we'd validate the ELM was correct ahead of time.
3. Add a set of `ToCharsEvaluator` tests that invoke the static function with incorrect operands and validate the error behavior.
4. Expand the XML test definition to support invalid operands and exceptions as results

Each of those come with different tradeoffs. I personally think the latter would be the most beneficial since it would be implementation-independent, and would allow us to drive towards standardization of error behavior across different implementations of CQL. That's a part of the spec that's a bit mushy at the moment.

JPercival (approved):

PR #1394, opened 2024-07-31
Add tests for ProperIn

antvaset:
This PR adds tests for the ProperIn evaluator following the changes to ProperContains to align it with the spec. Internally, ProperIn uses the ProperContains logic entirely because the operators are the same except for the order of the arguments.

### DIFF
--- a/Src/java/engine-fhir/src/test/resources/org/hl7/fhirpath/cql/CqlListOperatorsTest.xml
+++ b/Src/java/engine-fhir/src/test/resources/org/hl7/fhirpath/cql/CqlListOperatorsTest.xml
@@ -720,12 +720,52 @@
 		</test>
 	</group>
 	<group name="ProperIn">
+		<test name="ProperIn1">
+			<expression>'a' properly included in null as List&lt;String&gt;</expression>
+			<output>false</output>
+		</test>
+		<test name="ProperIn2">
+			<expression>'a' properly included in {}</expression>
+			<output>false</output>
+		</test>
+		<test name="ProperIn3">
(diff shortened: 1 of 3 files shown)


antvaset:
Depends on https://github.com/cqframework/cql-tests/pull/40

JPercival (approved):

Src/java/engine-fhir/src/test/resources/org/hl7/fhirpath/cql/CqlListOperatorsTest.xml
```diff
@@ -720,12 +720,52 @@
...
+			<expression>'a' properly included in { 'a', 'a' }</expression>
+			<output>false</output>
+		</test>
+		<test name="ProperIn8">
+			<expression>'c' properly included in { 'a', 'b' }</expression>
+			<output>false</output>
+		</test>
+		<test name="ProperIn9">
+			<expression>'a' properly included in { 'a', null }</expression>
+			<output>null</output>
```
brynrhodes, line 765: Seems like this should be true?

brynrhodes (changes requested):

Src/java/engine-fhir/src/test/resources/org/hl7/fhirpath/cql/CqlListOperatorsTest.xml
antvaset: Also whichever we go with, I'll add the short explainers as comments to the XMLs.

Issue #1408, opened 2024-08-27
CQL Fluent Functions Overload function name


SeenaFa:
When overloading a fluent function, using the same function name but different arguments, it seems that the order in which the functions appear in the CQL library matters. 

Below is an example showing how you get different results when ordering the fluent functions:

`library IssueFluentFunctions

using FHIR version '4.0.1'

context Patient

define "EobFluentFunctionError": ([ExplanationOfBenefit]).ffoverload()

define fluent function ffoverload(eob List<ExplanationOfBenefit>):
    eob.item.serviced.value

define fluent function ffoverload(obs List<Observation>):
    obs.effective.value

////////////////////////////////////////////////////////////////////////////////////

define "EobFluentFunctionError2": ([ExplanationOfBenefit]).ffoverload2()

define fluent function ffoverload2(obs List<Observation>):
    obs.effective.value

define fluent function ffoverload2(eob List<ExplanationOfBenefit>):
    eob.item.serviced.value`

The results are as follows:

Patient=Patient(id=SamplePatient1)
EobFluentFunctionError=[]
EobFluentFunctionError2=[2023-03-31]

Not sure what the expected behavior is but the expected result is that the date should be returned for both.

[IssueFluentFunctions.zip](https://github.com/user-attachments/files/16769375/IssueFluentFunctions.zip)

PR #1410, opened 2024-09-09
Use type specifiers from signature for function ref resolution, if present. Raise exception if resolution is ambiguous.

antvaset:
Closes https://github.com/cqframework/clinical_quality_language/issues/1408.

The linked issue is due to the engine not being able to distinguish between two `func(List<A>)` and `func(List<B>)` overloads at runtime. At compile time, the functions are resolved correctly, and it's possible to include the disambiguated signatures in the ELM by setting signature level to Overloads or All. If the signatures are present in the ELM, we should therefore use those signatures to resolve function refs at runtime which will fix the issue.

This PR adds support for resolving function refs at runtime using the function ref's signature if the signature is included in the ELM. The changed `FunctionRefEvaluator` now also throws if it cannot disambiguate between multiple candidate function defs.

Below is an overview of the `FunctionRefEvaluator` logic before and after the change.

BEFORE: Find the function defs with the matching name and number of operands. If only one function def is found, pick that function def and stop there. If there are multiple candidate function defs, pick the first one with the matching operand types. The Java class types of the function ref's signature (if present, type specifiers are converted to Java classes) or of the evaluated argument expressions are used for comparison. If no function defs are matched in the end, an exception is thrown `Could not resolve call to operator 'FUNCTION(ARGUMENT_TYPES)' in library 'LIBRARY_NAME'.`.

AFTER: Find the function defs with the matching name. If the function ref includes a signature, use the signature to further select the matching function defs (type specifiers are used for comparison). If the function ref doesn't include a signature, the Java class types of the evaluated argument expressions are used for comparison. If no function defs are left in the end, an exception is thrown `Could not resolve call to operator 'FUNCTION(ARGUMENT_TYPES)' in library 'LIBRARY_NAME'.`. If there are multiple matching function defs, an exception is thrown `Ambiguous call to operator 'FUNCTION(ARGUMENT_TYPES)' in library 'LIBRARY_NAME'.`. If only one function def is left in the end, pick that function def (normal case).

I am reusing the type specifier comparison implementation from the `SimpleElmEvaluator` class, which was handy.


antvaset:
Hi @JPercival thanks for the tick! I've pushed some changes to the JaCoCo task so that the coverage of `SimpleElmEngine` is visible, yet it won't update the bot's comment, so please see the updated report here https://app.codecov.io/gh/cqframework/clinical_quality_language/pull/1410.

antvaset:
> Hi @JPercival thanks for the tick! I've pushed some changes to the JaCoCo task so that the coverage of `SimpleElmEngine` is visible, yet it won't update the bot's comment, so please see the updated report here https://app.codecov.io/gh/cqframework/clinical_quality_language/pull/1410.

Scratch that, it's just updated the comment.

JPercival (approved):

brynrhodes (approved):

PR #1413, opened 2024-09-13
Swap Gradle language from Groovy to Kotlin

JPercival:
* Fix usage of Java 17 API in the build
* Swap the build language from Groovy to Kotlin


antvaset (approved):

Src/java/build.gradle.kts
```diff
@@ -0,0 +1,24 @@
...
+}
+
+sonar {
+  properties {
+    property("sonar.projectKey", "cqframework_clinical_quality_language")
+    property("sonar.organization", "cqframework")
+    property("sonar.host.url", "https://sonarcloud.io")
+  }
+}
+
```
cmoesel, line 13: You've removed the `idea` task. If you do not plan to support this anymore, you should also remove it from the [Java README](https://github.com/cqframework/clinical_quality_language/tree/feature-kotlin-build2/Src/java) and [Java Quick Start documentation](https://github.com/cqframework/clinical_quality_language/blob/feature-kotlin-build2/Src/java-quickstart/README.md).

cmoesel (changes requested):
Very cool. This is a great first step! Everything makes sense to me. I only left two comments. Since this is going into `master`, I'm marking this "request changes" because I don't think the docs should reference idea project generation if it's not supported anymore.

I also noticed that when I run this, I get a warning that I'm _not_ seeing on the master branch:
> \> Task :model:compileJava
> /Users/cmoesel/git/cql/pr1413/Src/java/model/build/generated/sources/model/main/java/org/hl7/fhir/Identifier.java:108: warning: [unchecked] Possible heap pollution from parameterized vararg type JAXBElement<? extends Element>
    public Identifier withRest(JAXBElement<? extends Element> ... values) {

It's probably not a big deal, but I'm mentioning it since it does appear to be new.

Src/java/build.gradle.kts
JPercival: IntelliJ supports this natively now, so it's unneeded. I will update the docs.

Src/java/buildSrc/src/main/kotlin/cql.library-conventions.gradle.kts
```diff
@@ -0,0 +1,16 @@
...
+    id("cql.java-conventions")
+}
+
+dependencies {
+    // Various libraries for Android signatures are available, Jackson uses this one
+    signature("com.toasttab.android:gummy-bears-api-${project.findProperty("android.api.level")}:0.5.0@signature")
+}
+
+tasks.animalsnifferTest {
+    enabled = false
```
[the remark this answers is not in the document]
JPercival: Previously, the tests were disabled a different way. The test source sets were excluded from animalSniffer:

https://github.com/cqframework/clinical_quality_language/pull/1413/files/5a43d148d49fb2a96d49d35ca00ac3e7d6c10771#diff-d3a7856aeb130f9b038332f5266fcce2a72a55516858f243570a035f76394649L17-L19

 I felt that disabling it explicitly was a bit more clear than source-set manipulation. Happy to change that if you feel differently.

JPercival:
> I also noticed that when I run this, I get a warning that I'm _not_ seeing on the master branch:
> 
> > > Task :model:compileJava
> > /Users/cmoesel/git/cql/pr1413/Src/java/model/build/generated/sources/model/main/java/org/hl7/fhir/Identifier.java:108: warning: [unchecked] Possible heap pollution from parameterized vararg type JAXBElement<? extends Element>
> > public Identifier withRest(JAXBElement<? extends Element> ... values) {
> 
> It's probably not a big deal, but I'm mentioning it since it does appear to be new.

Huh. My guess is that we were previously excluding the `generated` sources from compiler warnings. I don't know what change would have caused this though. I will investigate.

Src/java/build.gradle.kts
JPercival: Modulo the grammar, actually. I'll look at that...

Src/java/buildSrc/src/main/kotlin/cql.library-conventions.gradle.kts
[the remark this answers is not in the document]
cmoesel: Oh, I see. So it's not disabling a testing capability within animalSniffer -- it's saying not to run animalSniffer on the tests in the project. Got it. I'm fine with it as you have it; I just wanted to understand the change. Now I do. Thanks!

JPercival:
@cmoesel - The heap pollution was a real issue. The xjc generation was running for all the classes in the schema in the `model` project, not just those relevant to `model`.

cmoesel (approved):
Looks good to me. Thanks, @JPercival!

PR #1422, opened 2024-10-07
Add support for converting CQL Tuples to FHIR Parameters

JPercival:
* Add support for converting CQL Tuples to FHIR Parameters
* Tests for same

### DIFF
--- a/Src/java/engine-fhir/src/main/java/org/opencds/cqf/cql/engine/fhir/converter/Dstu2FhirTypeConverter.java
+++ b/Src/java/engine-fhir/src/main/java/org/opencds/cqf/cql/engine/fhir/converter/Dstu2FhirTypeConverter.java
@@ -230,7 +230,7 @@ class Dstu2FhirTypeConverter extends BaseFhirTypeConverter {
     private static void addPartWithNameAndValue(
             Parameters.ParametersParameterComponent param, String key, Object value) {
         if (value instanceof Parameters.ParametersParameterComponent) {
-            var part = (Parameters.ParametersParameterComponent)value;
+            var part = (Parameters.ParametersParameterComponent) value;
             part.setName(key);
             param.addPart(part);
         } else {
--- a/Src/java/engine-fhir/src/main/java/org/opencds/cqf/cql/engine/fhir/converter/Dstu2FhirTypeConverter.java
+++ b/Src/java/engine-fhir/src/main/java/org/opencds/cqf/cql/engine/fhir/converter/Dstu2FhirTypeConverter.java
@@ -1,8 +1,5 @@
 package org.opencds.cqf.cql.engine.fhir.converter;
 
-import ca.uhn.fhir.model.api.IDatatype;
-import ca.uhn.fhir.model.api.IResource;
-import ca.uhn.fhir.model.dstu2.resource.Parameters;
 import java.math.BigDecimal;
 import java.util.Calendar;
 import java.util.Collection;
@@ -246,20 +243,20 @@ class Dstu2FhirTypeConverter extends BaseFhirTypeConverter {
             part.setName(key);
             var element = value.getElements().get(key);
             if (element == null) {
-                part.addUndeclaredExtension(false, NULL_EXT_URL, new BooleanType(true));
+                part.addExtension(NULL_EXT_URL, new BooleanType(true));
                 continue;
             } else if (element instanceof Collection) {
                 if (((Collection<?>) element).isEmpty()) {
-                    part.addUndeclaredExtension(false, EMPTY_EXT_URL, new BooleanType(true));
+                    part.addExtension(EMPTY_EXT_URL, new BooleanType(true));
                     continue;
                 }
             }
 
             var result = toFhirType(element);
-            if (result instanceof IResource) {
-                part.setResource((IResource) result);
-            } else if (result instanceof IDatatype) {
-                part.setValue((IDatatype) result);
+            if (result instanceof Resource) {
+                part.setResource((Resource) result);
+            } else if (result instanceof Type) {
+                part.setValue((Type) result);
             } else {
                 throw new IllegalArgumentException("Tuple contains unsupported type");
             }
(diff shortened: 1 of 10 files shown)


Src/java/engine-fhir/src/main/java/org/opencds/cqf/cql/engine/fhir/converter/Dstu2FhirTypeConverter.java
```diff
@@ -227,7 +228,35 @@ public IBase toFhirTuple(Tuple value) {
...
+            return parameters;
+        }
+
+        var param = parameters.addParameter();
+        for (String key : value.getElements().keySet()) {
+            var part = param.addPart();
+            part.setName(key);
+            var element = value.getElements().get(key);
+            if (element == null) {
+                part.addExtension()
```
brynrhodes, line 242: The data-absent-reason extension needs to be on the value, not the part:

```xml
<parameter>
  <extension url="http://hl7.org/fhir/StructureDefinition/cqf-cqlType">
    <valueString value="System.Boolean"/>
  </extension>
  <name value="Is pregnancy confirmed"/>
  <valueBoolean>
    <extension url="http://hl7.org/fhir/StructureDefinition/data-absent-reason">
      <valueCode value="unknown"/>
    </extension>
  </valueBoolean>
</parameter>
```

Src/java/engine-fhir/src/main/java/org/opencds/cqf/cql/engine/fhir/converter/Dstu2FhirTypeConverter.java
```diff
@@ -227,7 +228,35 @@ public IBase toFhirTuple(Tuple value) {
...
+                        .setUrl(DATA_ABSENT_REASON_EXT_URL)
+                        .setValue(new CodeType(DATA_ABSENT_REASON_UNKNOWN_CODE));
+            } else if (element instanceof Collection && ((Collection<?>) element).isEmpty()) {
+                part.addExtension().setUrl(EMPTY_LIST_EXT_URL).setValue(new BooleanType(true));
```
brynrhodes, line 246: Same issue here, the cqf-isEmptyList extension needs to be on the value, not the part:

```xml
<parameter>
  <extension url="http://hl7.org/fhir/StructureDefinition/cqf-cqlType">
    <valueString value="List<Tuple{id:System.Integer,name:System.String,address:List<Tuple{street:System.String,city:System.String,state:System.String}>}>"/>
  </extension>
  <name value="CQLEmptyListExample"/>
  <valueBoolean>
    <extension url="http://hl7.org/fhir/StructureDefinition/cqf-isEmptyList">
      <valueBoolean value="true"/>
    </extension>
  </valueBoolean>
</parameter>
```

Src/java/engine-fhir/src/main/java/org/opencds/cqf/cql/engine/fhir/converter/Dstu2FhirTypeConverter.java
```diff
@@ -227,7 +228,35 @@ public IBase toFhirTuple(Tuple value) {
...
+            } else {
+                var result = toFhirType(element);
```
brynrhodes, line 248: What happens in the case that the element is a Tuple or List? The expected behavior is that these would be subparts, with Lists represented as repeating elements: https://hl7.org/fhir/uv/cql/Parameters-cql-typemappingexampleresult.html

brynrhodes: If I'm reading the code right, what it will actually do is put this in as a Parameters resource

brynrhodes (dismissed):

Src/java/engine-fhir/src/main/java/org/opencds/cqf/cql/engine/fhir/converter/Dstu2FhirTypeConverter.java
JPercival: This isn't mapping all the evaluation results for a Library; it's mapping arbitrary (potentially recursive) values. There's no way to know that you need to lift or flatten the way "Using CQL with FHIR" IG specifies without that larger context. You'd need a second step to flatten the results. We don't even have the expression name here.

Src/java/engine-fhir/src/main/java/org/opencds/cqf/cql/engine/fhir/converter/Dstu2FhirTypeConverter.java
JPercival: CQL does not track any metadata about types as part of the execution. We do not even know what type the missing value is. All we know is that "X" is "null".

Src/java/engine-fhir/src/main/java/org/opencds/cqf/cql/engine/fhir/converter/Dstu2FhirTypeConverter.java
JPercival: Same situation - we don't have that type information. Is a `valueBoolean` the correct value for a List of Tuples?

Src/java/engine-fhir/src/main/java/org/opencds/cqf/cql/engine/fhir/converter/Dstu2FhirTypeConverter.java
JPercival: I don't know what I was thinking. You're right.

barhodes (approved):

brynrhodes (approved):
Nested tuples weren't being handled correctly, and in fixing that I actually found a gap in empty tuple representation. I've submitted JIRA trackers to the extensions library and CQL IG to address that gap, and pre-adopted the solution in my updates to this PR.

barhodes (approved):

barhodes (approved):

Issue #1420, opened 2024-10-03
DateTime function should use time zone of the evaluation request


brynrhodes:
The following expression:

```cql
DateTime(2024, 10, 3, 15, 54, 0, 0)
```

should result in a DateTime value with the timezone offset of the evaluation request: https://cql.hl7.org/09-b-cqlreference.html#datetime-1

However, looking at the code, the local timezone is always used:

https://github.com/cqframework/clinical_quality_language/blob/master/Src/java/engine/src/main/java/org/opencds/cqf/cql/engine/runtime/DateTime.java#L94

antvaset:
Hi @brynrhodes, I've added tests for this here: https://github.com/cqframework/clinical_quality_language/pull/1423

It looks like it picks up the timezone offset of the evaluation request as per the spec.

When you call `engine.expression(...)` or `engine.evaluate(...)` you have to pass in the evaluation date time explicitly every time, otherwise it gets reset to `ZonedDateTime.now()`.

PR #1423, opened 2024-10-08
Add test for default timezone offset of DateTime

antvaset:
Closes https://github.com/cqframework/clinical_quality_language/issues/1420.

This PR adds tests for calling `DateTime(y, m, d, h, m, s, ms)` without specifying the timezone offset. As per the spec, it should default to the timezone offset of the evaluation request.

Note that when `engine.expression(...)` or `engine.evaluate(...)` is called without passing in the evaluation date time explicitly, it is set to `ZonedDateTime.now()`.

### DIFF
--- a/Src/java/engine/src/test/java/org/opencds/cqf/cql/engine/execution/CqlDateTimeOperatorsTest.java
+++ b/Src/java/engine/src/test/java/org/opencds/cqf/cql/engine/execution/CqlDateTimeOperatorsTest.java
@@ -4,6 +4,10 @@ import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.is;
 import static org.junit.jupiter.api.Assertions.*;
 
+import java.time.ZoneId;
+import java.time.ZonedDateTime;
+import java.util.EnumSet;
+import java.util.List;
 import org.hl7.elm.r1.VersionedIdentifier;
 import org.junit.jupiter.api.Test;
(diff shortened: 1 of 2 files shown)


brynrhodes (changes requested):
This is only the test, but doesn't the DateTimeEvaluator need to be updated as well?

JPercival:
I walked through the code with Anton. The code looks correct to me. I think I was wrong before, the DateTimeEvaluator is using the timezone of the evaluation request.

brynrhodes:
Hmm.... I walked through this the other day and it wasn't, just did again and I don't see how it is using the evaluation request timestamp:

https://github.com/cqframework/clinical_quality_language/blob/master/Src/java/engine/src/main/java/org/opencds/cqf/cql/engine/elm/executing/DateTimeEvaluator.java#L33

JPercival:
The only place that's used is here:
https://github.com/cqframework/clinical_quality_language/blob/master/Src/java/engine/src/main/java/org/opencds/cqf/cql/engine/execution/EvaluationVisitor.java#L785-L786

Either the ELM has the offset embedded, or the current offset from the State object is used. That state is set up as part of the evaluation.

brynrhodes:
I see, I was assuming the evaluator tree was still controlling the visit, but I see that it's the visitor now.

brynrhodes (approved):

JPercival (approved):

PR #1425, opened 2024-10-20
Make sure `per` is set to something when visiting `expand` and `collapse`

antvaset:
Currently, the translator fails to build the call context for `collapse { null }` returning

```
Could not determine signature for invocation of operator System.Collapse.
```

This is because the `per` argument is not instantiated inside `visitSetAggregateExpressionTerm` when it's not explicitly set and `source` is not a list of intervals. In the case of `collapse { null }`, `source` is of type `List<System.Any>`. The fallback expression for `per` in all cases should be `null` of type `System.Quantity`.

### DIFF
--- a/Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/Cql2ElmVisitor.java
+++ b/Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/Cql2ElmVisitor.java
@@ -2746,7 +2746,9 @@ public class Cql2ElmVisitor extends CqlPreprocessorElmCommonVisitor {
     @Override
     public Object visitSetAggregateExpressionTerm(cqlParser.SetAggregateExpressionTermContext ctx) {
         Expression source = parseExpression(ctx.expression(0));
-        Expression per = null;
+
+        // If `per` is not set, it will remain `null as System.Quantity`.
+        Expression per = libraryBuilder.buildNull(libraryBuilder.resolveTypeName("System", "Quantity"));
         if (ctx.dateTimePrecision() != null) {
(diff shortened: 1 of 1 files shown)


Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/Cql2ElmVisitor.java
```diff
@@ -2746,7 +2746,9 @@ public Object visitAggregateExpressionTerm(cqlParser.AggregateExpressionTermCont
     @Override
     public Object visitSetAggregateExpressionTerm(cqlParser.SetAggregateExpressionTermContext ctx) {
         Expression source = parseExpression(ctx.expression(0));
-        Expression per = null;
+
+        // If `per` is not set, it will remain `null as System.Quantity`.
+        Expression per = libraryBuilder.buildNull(libraryBuilder.resolveTypeName("System", "Quantity"));
```
JPercival, line 2751: question: Seems like this actually move to line 2757? Ideally, we're not setting `per` more than once, because that implies we're doing (potentially) unnecessary work?

JPercival (approved):

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/Cql2ElmVisitor.java
antvaset: Hi @JPercival I agree this adds unnecessary work when `per` is later reassigned (reassignment itself is also bad..). I chose to do it this way because otherwise we'd repeat `libraryBuilder.buildNull(libraryBuilder.resolveTypeName("System", "Quantity"));` multiple times.

brynrhodes (approved):

PR #1426, opened 2024-10-23
Implemented Any support with wildcards instead of downcast; improved …

brynrhodes:
…instantiation behavior


JPercival:
Postponed, superseded by #1428

PR #1428, opened 2024-10-29
Improved generic instantiation behavior and added type precedence to …

brynrhodes:
…ambiguous resolution logic


Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/model/InstantiationResult.java
```diff
@@ -26,6 +35,12 @@ public Operator getOperator() {
         return operator;
     }
 
+    private Map<TypeParameter, DataType> typeMap;
+
+    public Map<TypeParameter, DataType> getTypeMap() {
+        return typeMap;
+    }
```
antvaset, line 42: The `typeMap` member or this getter aren't used anywhere yet. Should it be deleted?

antvaset:
Proposing to add this to this for completeness https://github.com/cqframework/clinical_quality_language/pull/1431

JPercival (approved):

antvaset (approved):

PR #1438, opened 2024-11-06
Kotlin Conventions Plugin

JPercival:
* Add formatting and static analysis checks for Kotlin code
* Convert a couple sample classes to Kotlin to ensure conventions work


Src/java/cql/src/main/java/org/cqframework/cql/Application.kt
```diff
@@ -0,0 +1,25 @@
...
+        val lexer = cqlLexer(chars)
+        val tokens = CommonTokenStream(lexer)
+        val parser = cqlParser(tokens)
+
+        parser.buildParseTree = true
+        val tree = parser.library()
+
+        // show tree in text form
+        println(tree.toStringTree(parser))
+    }
```
cmoesel, line 24: Since this is not a static function, if you try to run it via `gradle :cql:run`, it will fail with an error like:
> Error: Main method is not static in class org.cqframework.cql.Application, please define the main method as:
>     public static void main(String[] args)

Based on some googling, I found two approaches that work.

One approach is to declare `Application` as a class with a companion object containing an annotated main function. E.g.:
```kt
class Application {
    companion object {
        @JvmStatic
        fun main(args: Array<String>) {
            // all the good stuff goes here
        }
    }
}
```

The other option is to drop the wrapping `Application` object entirely and just define `main` as a standalone function (not inside an object or class). In this case, however, you'll need to change the `mainClass` in `cql/build.gradle.kts` like this:
```gradle
application {
    mainClass = "org.cqframework.cql.ApplicationKt"
}
```

I tried both approaches locally via `gradle :cql:run --args="/path/to/my/cql/file"` and they seem to work. I don't know Kotlin at all, so maybe one way is preferable to the other. I'll leave that to you to decide.

cmoesel (changes requested):
Cool. Kotlin looks nice!

I found one issue related to the CLI application, but it has an easy fix. See my individual comment on the `Application.kt` file.

Src/java/cql/src/main/java/org/cqframework/cql/Application.kt
JPercival: I'll do your first alternative for the sake of providing a reference in the code base for those new to Kotlin on companion objects and the @JvmStatic annotation.

antvaset (approved):

Src/java/cql/src/main/java/org/cqframework/cql/Application.kt
JPercival: Updated.

PR #1449, opened 2024-11-20
Convert Invocation hierarchy to Kotlin

JPercival:


antvaset (approved):

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/model/invocation/SkipInvocation.kt
```diff
@@ -0,0 +1,16 @@
+package org.cqframework.cql.cql2elm.model.invocation
+
+import org.hl7.elm.r1.Expression
+import org.hl7.elm.r1.Slice
+
+/** Created by Bryn on 5/17/2017. */
```
lukedegruchy, line 6: I wonder if this comment is misleading, since this is technically a brand-new Kotlin class.

lukedegruchy (approved):

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/model/invocation/OperatorExpressionInvocation.kt
```diff
@@ -0,0 +1,14 @@
+package org.cqframework.cql.cql2elm.model.invocation
+
+import org.hl7.elm.r1.OperatorExpression
+import org.hl7.elm.r1.TypeSpecifier
+
+/** Created by Bryn on 4/12/2018. */
+abstract class OperatorExpressionInvocation<O : OperatorExpression>
+protected constructor(expression: O) : AbstractExpressionInvocation<O>(expression) {
```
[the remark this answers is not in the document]
JPercival: The constructor is not needed. Missed it on cleanup. Fixed!

The wildcard suppression is because Kotlin is stricter about type safety by default than Java. This statement:

```
override var signature: List<TypeSpecifier>
```

generates this getter in Java

```
public List<? extends TypeSpecifier> getSignature()
```

because `OperatorExpressionInvocation` is not sealed, and therefore subclass could override `getSignature` with any subclass of `TypeSpecifier`. For example, a hypothetical `ListSignaturededOperatorExpressionInvocation` could override `getSignature` thusly:

```
public List<ListTypeSpecifier> getSignature()
```

So to be completely safe, the superclass code should treat `signature` as "some unknown subclass of TypeSpecifier" as opposed to "TypeSpecifier specifically".

Anyway, the rest of the code in the compiler assumes the standard Java behavior of not using wildcards, so I suppressed them, perhaps temporarily until we more of the project converted.

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/model/invocation/SkipInvocation.kt
JPercival: I debated that. It's kind of a low-value comment. But Bryn is still the #1 expert on the compiler design so I decided that I'd rather people go to him with questions on the class based on the comment than to me based on the commit log. :)

PR #1457, opened 2024-11-27
Conversion of Model hierarchy to Kotlin

JPercival:
* Converted the rest of the `cql2elm.model` package to Kotlin
* Removed dead paths for exception handling
* Fixed some null pointer exceptions resulting from stricter null handling


Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/model/CompiledLibrary.kt
```diff
@@ -0,0 +1,281 @@
...
+import org.hl7.elm.r1.IncludeDef
+import org.hl7.elm.r1.Library
+import org.hl7.elm.r1.ParameterDef
+import org.hl7.elm.r1.UsingDef
+import org.hl7.elm.r1.ValueSetDef
+import org.hl7.elm.r1.VersionedIdentifier
+
+@Suppress("TooManyFunctions")
+class CompiledLibrary {
+    var identifier: VersionedIdentifier? = null
```
lukedegruchy, line 24: I really like Kotlin's clarity about what's nullable and what isn't.  It would be nice if such a mechanism existed in Java beyond Optional and @Nullable we could use.

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/model/CompiledLibrary.kt
```diff
@@ -0,0 +1,281 @@
...
+    fun remove(expression: ExpressionDef) {
+        require(expression !is FunctionDef) { "FunctionDef cannot be removed." }
+        namespace.remove(expression.name)
+    }
+
+    private fun ensureLibrary(operator: Operator) {
+        // Functions can be defined in an anonymous library
+        if (this.identifier != null && identifier!!.id != null) {
+            if (operator.libraryName == null) {
+                operator.libraryName = identifier!!.id
```
lukedegruchy, line 98: Question:  From what I understand, Kotlin's !! is meant to throw an NPE if the variable in question is null.  Given Kotlin has "?" under which situations do we want to deliberately use "!!" in the CQL project?

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/model/CompiledLibrary.kt
```diff
@@ -0,0 +1,281 @@
...
+                }
+                results
+            }
+        }
+    }
+
+    fun resolveCall(callContext: CallContext, conversionMap: ConversionMap): OperatorResolution? {
+        val resolution = operatorMap.resolveOperator(callContext, conversionMap)
+
+        if (resolution != null) {
```
lukedegruchy, line 248: Question:  Do we not intend to check for a null resolution.getOperation() like we did in Java?

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/model/ConversionMap.kt
```diff
@@ -0,0 +1,451 @@
...
+        operatorMap: OperatorMap
+    ): Conversion? {
+        return findConversion(
+                fromType.pointType,
+                toType.pointType,
+                isImplicit = true,
+                allowPromotionAndDemotion = false,
+                operatorMap = operatorMap
+            )
+            ?.let { Conversion(fromType, toType, it) }
```
lukedegruchy, line 180: Question, is this the equivalent of Optional.ofNullable(findConversion(..)).map(it -> new Conversion(...)

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/model/OperatorEntry.kt
```diff
@@ -0,0 +1,400 @@
+package org.cqframework.cql.cql2elm.model
+
+import java.util.*
+import kotlin.collections.ArrayList
+import kotlin.collections.HashMap
+import org.hl7.cql.model.ChoiceType
+import org.hl7.cql.model.DataType
+
+class OperatorEntry(val name: String) {
```
lukedegruchy, line 9: Are we not guarding against empty name Strings as the Java code does?

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/model/ResolvedIdentifierContext.kt
```diff
@@ -0,0 +1,123 @@
...
+                return Optional.ofNullable(element)
+            }
+
+            return Optional.empty()
+        }
+
+    private val isExactMatch: Boolean
+        get() = ResolvedIdentifierMatchType.EXACT == matchType
+
+    fun warnCaseInsensitiveIfApplicable(): Optional<String> {
```
lukedegruchy, line 43: Why an Optional<String> and not a String?

lukedegruchy (approved):

lukedegruchy (approved):

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/model/CompiledLibrary.kt
JPercival: I'll take a look at this again. I removed most of the cases where this happened in the auto-generated code. Looks like I missed this one.

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/model/CompiledLibrary.kt
JPercival: `getOperation` is no longer nullable:
https://github.com/cqframework/clinical_quality_language/blob/master/Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/model/OperatorResolution.kt#L5

One of the nice thing about the explicit nulls is that the cleanliness is viral. :)

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/model/ResolvedIdentifierContext.kt
JPercival: This was how to Kotlin convertor generated it. I'll take another look and see if it can be simplified.

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/model/OperatorEntry.kt
JPercival: I'll add it back.

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/model/ConversionMap.kt
JPercival: Yes. `.let` only runs if the result is not null.

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/model/ResolvedIdentifierContext.kt
JPercival: Went with `String?`, but it's about the same as far as code clarity. Will need to experiment more with `Optional`. In principle it's made redundant by having explicit nullability. I kind of like the `map` functionality though.

PR #1458, opened 2024-12-01
Convert remaining cql-to-elm main classes to Kotlin

antvaset:
* Still working on tidying up the Kotlin code.
* So far left out the `LibrarySourceProvider` interface because it has a default method. See more [here](https://stackoverflow.com/questions/44179996/using-default-function-implementation-of-interface-in-kotlin).


JPercival:
Merged into the `feature-kotlin` branch

PR #1464, opened 2024-12-06
Replace xjc with custom XSD Java code gen

antvaset:
* Add new cql.xsd-java-gen-conventions plugin that runs `node Src/js/xsd-java-gen/generate.js`.
* `generate.js` reads the XSD files and generates Java classes with minimal imports and without annotations. Currently the input XSDs are hardcoded but should be read in from command line args.
* Add new cql.xjc-temp-conventions plugin which does the same thing as cql.xjc-conventions but skips `sourceSets` and `sourcesJar` and outputs the generated Java classes to `...-xjc-temp`. This allows us to generate the Java classes in place (in the same Java package) using the new method and have the original code also made available for comparison.
* Enable cql.xsd-java-gen-conventions and cql.xjc-temp-conventions for the elm and model modules.


antvaset:
Superseded by https://github.com/cqframework/clinical_quality_language/pull/1474

PR #1465, opened 2024-12-11
Convert the `model` module to Kotlin

JPercival:


Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/model/Model.kt
```diff
@@ -33,7 +33,7 @@ open class Model(val modelInfo: ModelInfo, modelManager: ModelManager?) {
 
         for (t in index.values) {
             if (t is ClassType && t.label != null) {
-                classIndex[casify(t.label)] = t
+                classIndex[casify(t.label!!)] = t
```
lukedegruchy, line 36: Are we sure we want to throw an NPE here? (!!)

JPercival: Ideally, no. I've been fixing the cases of these where it's reasonably straight forward to do so. In this case, the `label` property is mutable, meaning that it could change values between the null check (`t.label != null`) and the usage (`t.label!!`) if, for instance, the Model was shared between different threads. I looked at changing `label` to be immutable, but it requires a re-work of how the `ClassTypes` are initialized. I just decided it was a bigger change than I wanted do. I've _mostly_ kept the APIs and usages of the classes the same, only grabbing low-hanging fruit.

Src/java/elm-fhir/src/test/java/org/cqframework/cql/elm/requirements/fhir/DataRequirementsProcessorTest.java
```diff
@@ -1933,7 +1933,14 @@ private void assertEqualToExpectedModuleDefinitionLibrary(
...
         expectedModuleDefinitionLibrary.setDate(null);
-        assertTrue(actualModuleDefinitionLibrary.equalsDeep(expectedModuleDefinitionLibrary));
+
+        parser.setPrettyPrint(true);
+        var jsonExpected = parser.encodeResourceToString(expectedModuleDefinitionLibrary);
+        var jsonActual = parser.encodeResourceToString(actualModuleDefinitionLibrary);
+
+        assertEquals(jsonExpected, jsonActual);
+
+        // assertTrue(actualModuleDefinitionLibrary.equalsDeep(expectedModuleDefinitionLibrary));
```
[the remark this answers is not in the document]
JPercival: Fixed.

Src/java/model/src/main/java/org/hl7/cql/model/ClassType.kt
```diff
@@ -0,0 +1,298 @@
...
+
+        return super.isCompatibleWith(other)
+    }
+
+    override val isGeneric: Boolean
+        get() = genericParameters.isNotEmpty()
+
+    override fun isInstantiable(callType: DataType, context: InstantiationContext): Boolean {
+        return if (callType is ClassType && callType.elements.size == elements.size) {
+            sortedElements.zip(callType.sortedElements).all { (thisElement, thatElement) ->
```
lukedegruchy, line 281: We really need zip in Java

lukedegruchy (approved):

Src/java/model/src/main/java/org/hl7/elm_modelinfo/r1/serializing/ModelInfoReaderFactory.kt
```diff
@@ -0,0 +1,45 @@
...
+    @JvmStatic
+    fun getReader(contentType: String): ModelInfoReader? {
+        val providers = providers(false)
+        if (providers.hasNext()) {
+            val p = providers.next()
+            if (providers.hasNext()) {
+                throw RuntimeException(
+                    java.lang.String.join(
+                        " ",
+                        "Multiple ModelInfoReaderProviders found on the classpath.",
```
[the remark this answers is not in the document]
JPercival: Yes. I’ve not tried them though. I’ll give the a shot.

PR #1471, opened 2024-12-13
Break out UCUM Service into its own module

JPercival:


Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/ucum/UcumServiceFactory.kt
```diff
@@ -0,0 +1,15 @@
+package org.cqframework.cql.cql2elm.ucum
+
+import java.util.*
+
+@Suppress("TooGenericExceptionThrown")
+object UcumServiceFactory {
+    fun load(): UcumService {
+        return ServiceLoader.load(UcumService::class.java).firstOrNull()
+            ?: throw RuntimeException(
```
lukedegruchy, line 9: Shouldn't we be throwing something more specific, like IllegalStateException instead of RuntimeException in all cases, to signal the problem is with the state of the application, and not a user error?

lukedegruchy (approved):

Src/java/ucum/src/main/java/org/cqframework/cql/ucum/DefaultUcumService.kt
```diff
@@ -0,0 +1,46 @@
+package org.cqframework.cql.ucum
+
+import org.cqframework.cql.cql2elm.ucum.UcumService
+import org.fhir.ucum.Decimal
+import org.fhir.ucum.UcumEssenceService
+
+@ConsistentCopyVisibility
+data class DefaultUcumService private constructor(private val ucumService: UcumService) :
+    UcumService by ucumService {
```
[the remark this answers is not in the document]
JPercival: Singletons (`object` in Kotlin-speak) require a no-args constructor.

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/ucum/UcumServiceFactory.kt
JPercival: Fixed to use IllegalStateException

PR #1474, opened 2024-12-17
Replace xjc with custom XSD Kotlin code gen

antvaset:
* Add new cql.xsd-kotlin-gen-conventions plugin that runs `node Src/js/xsd-kotlin-gen/generate.js`.
* `generate.js` reads the XSD files and generates Kotlin classes with minimal imports and without annotations. Currently the input XSDs are hardcoded but should be read in from command line args.
* Add new cql.xjc-temp-conventions plugin which does the same thing as cql.xjc-conventions but skips sourceSets and sourcesJar and outputs the generated Java classes to `...-xjc-temp`. This allows us to generate the Kotlin classes in place (in the same package) using the new method and have the original code also made available for comparison.
* Attempt to replace model-jaxb, model-jackson, elm-jaxb, eml-jackson with model-xmlutil and elm-xmlutil which use https://github.com/pdvrieze/xmlutil to read and write XML and JSON.
* There is some stuff that is currently simply commented out to help iterate faster. This needs fixing before the branch can be merged!
* fhir-modelinfo-3.0.1.xml and quickfhir-modelinfo-3.0.1.xml appeared to be broken and the XML deserializer was throwing "Content is not allowed in prolog". This was fixed by re-creating the files.


antvaset:
Superseded by https://github.com/cqframework/clinical_quality_language/pull/1479

Issue #1475, opened 2024-12-17
Mapping expansion does not respect the signature level flag


brynrhodes:
When compiling with SignatureLevel All or Overloads, the following library using FHIR correctly outputs the signature elements for the FHIRHelpers call:

```cql
library TestEncounterPeriod

using FHIR version '4.0.1'

include FHIRHelpers version '4.1.0'

parameter "Measurement Period" Interval<DateTime> default Interval[@2024-01-01T00:00:00.0Z, @2025-01-01T00:00:00.0Z)

context Patient

define Encounters:
  [Encounter] E
    where E.period during "Measurement Period"
```

However, the equivalent library using QICore does not:

```cql
library TestEncounterPeriodQICore

using QICore version '4.1.1'

include FHIRHelpers version '4.1.0'

parameter "Measurement Period" Interval<DateTime> default Interval[@2024-01-01T00:00:00.0Z, @2025-01-01T00:00:00.0Z)

context Patient

define Encounters:
  [Encounter] E
    where E.period during "Measurement Period"
```

This appears to be because the mapping expansion output is not respecting the signature level flag.

JPercival:
Resolved by #1476

PR #1476, opened 2024-12-17
#1475: Fixed mapping expansions for profile-informed authoring not re…

brynrhodes:
…specting signature level

### DIFF
--- a/Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/LibraryBuilder.java
+++ b/Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/LibraryBuilder.java
@@ -2743,6 +2743,22 @@ public class LibraryBuilder {
             String functionArgument = targetMap.substring(invocationStart + 1, targetMap.lastIndexOf(')'));
             Expression argumentSource =
                     functionArgument.equals("%value") ? source : applyTargetMap(source, functionArgument);
+
+            // NOTE: This is needed to work around the mapping for ToInterval
+            // FHIRHelpers defines multiple overloads of ToInterval, but the type mapping
+            // does not have the type of the source data type.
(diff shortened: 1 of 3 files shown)


JPercival (approved):

Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/LibraryBuilder.java
```diff
@@ -2743,6 +2743,22 @@ public Expression applyTargetMap(Expression source, String targetMap) {
             String functionArgument = targetMap.substring(invocationStart + 1, targetMap.lastIndexOf(')'));
             Expression argumentSource =
                     functionArgument.equals("%value") ? source : applyTargetMap(source, functionArgument);
+
+            // NOTE: This is needed to work around the mapping for ToInterval
```
lukedegruchy, line 2747: General feedback:  Please fix the title and description of this PR so at a minimum it's not a truncated version of the commit message.

Src/java/cql-to-elm/src/test/java/org/cqframework/cql/cql2elm/TranslationTests.java
```diff
@@ -467,4 +467,38 @@ void abstractClassNotRetrievable() throws IOException {
                 errors.stream().map(Throwable::getMessage).collect(Collectors.toList());
         assertThat(errorMessages, contains("Specified data type DomainResource does not support retrieval."));
     }
+
+    @Test
+    void mappingExpansionsRespectSignatureLevel() throws IOException {
```
lukedegruchy, line 472: Please add a test for Signature.None to contrast with the test below.  Also, it might be good to test other non-Overloads, non-None Signature types as well.

lukedegruchy (approved):

PR #1496, opened 2025-02-04
Translator supports namespaces

bkaney:
This is a working POC of adding support for CQL namespaces, it works with the example found here https://github.com/reason-healthcare/crmi-reusable-cql

This PR adds an option `-root-dir` that when present attempts to use the IG resource dependencies and CQL namespaces.


bkaney:
@brynrhodes / @JPercival - Could you have a look and see if this helps unblock ability to translate CQL with CQL namespaces.

brynrhodes (approved):
Looks good to me

JPercival (approved):

PR #1514, opened 2025-03-14
Changes and fixes pt. 2 (Kotlin feature branch)

antvaset:
Changes and fixes discussed in https://github.com/cqframework/clinical_quality_language/pull/1462.


Src/java/buildSrc/src/main/kotlin/cql.library-conventions.gradle.kts
```diff
@@ -3,11 +3,6 @@ plugins {
     id("cql.java-conventions")
 }
 
-dependencies {
-    // Various libraries for Android signatures are available, Jackson uses this one
-    signature("com.toasttab.android:gummy-bears-api-${project.findProperty("android.api.level")}:0.10.0:coreLib2@signature")
-}
-
```
JPercival, line 10: We still need this. The comment is _not_ saying that we use Jackson. It's saying that we're copying build patterns found in the Jackson project to address Android support.

JPercival (changes requested):

Src/java/buildSrc/src/main/kotlin/cql.library-conventions.gradle.kts
antvaset: I've added it back. Thank you for catching this!

JPercival (approved):

PR #1517, opened 2025-03-18
add #negateMembershipExpression

delcroip:
to support 'not in '


brynrhodes:
Hi @delcroip ! This is potentially a good suggestion, but any change in the grammar needs to be a specification change. I've added a JIRA tracker to support adding this in CQL. If the tracker is approved it would be added for consideration in the next ballot of CQL, currently scheduled for the September cycle:
https://jira.hl7.org/browse/FHIR-50166

PR #1535, opened 2025-04-22
Change default list comparison behavior to use "equivalent" semantics.

JPercival:
* The next version of the CQL spec adds a new `~in` operator to check for list membership using `equivalence` rather than `equality`.
* This PR changes the CQL engine to behavior to default to that, and adds an option to fall back to the current `equality` behavior.
* It also fixes a couple decimal equivalence tests that were found to be wrong in testing. Decimals should use round-half behavior rather than floor rounding for equivalence.


brynrhodes (approved):

lukedegruchy (approved):

PR #1542, opened 2025-05-03
Decouple `serialization` project from `cql-to-elm`

antvaset:
Currently, the `serialization` project contains the logic for parsing cql-options.json into `CqlTranslatorOptions` defined in `cql-to-elm`, and it therefore depends on `cql-to-elm`. Because of this, `serialization` cannot be imported in the opposite direction into `cql-to-elm` (for the JS build) without creating a circular dependency.

This PR moves `CqlTranslatorOptions` to `cql` (together with `CqlCompilerOptions`, `CqlCompilerException.ErrorSeverity`, and `LibraryBuilder.SignatureLevel` which are used by `CqlTranslatorOptions`). `cql` is imported by both `serialization` and `cql-to-elm`, so `CqlTranslatorOptions` are still available in both projects. There are some breaking changes, and it might not be ideal that the CQL compilation concerns get mixed into the `cql` project.

Alternatively, `serialization` could be decoupled from `cql-to-elm` by moving the serialization logic for cql-options.json into `cql-to-elm`. This is also not ideal because the serialization concern would get partially mixed into `cql-to-elm` and not stay inside `serialization` project specifically concerned with implementing serialization.

The third option is to create a new project just for `CqlTranslatorOptions`, `CqlCompilerOptions`, `CqlCompilerException.ErrorSeverity`, and `LibraryBuilder.SignatureLevel` (not perfect).


antvaset:
Superseded by https://github.com/cqframework/clinical_quality_language/pull/1546

Issue #1287, opened 2023-11-18
Revisit CQL Engine and Result APIs


JPercival:
The CqlEngine APIs were created in an ad hoc fashion, as we were uncertain what an optimal API would look like. This has resulted in a few cases where the APIs are inconsistent, don't follow Java best practices, or suffer from code smells such as "telescoping constructor", or having many function overloads.

The engine still isn't API complete/final. Specifically, we don't have all the debugging APIs in place, so any current work on updating the APIs will be a step along the way.

Example of many overloads (potentially addressed by a params class):
https://github.com/cqframework/clinical_quality_language/blob/master/Src/java/engine/src/main/java/org/opencds/cqf/cql/engine/execution/CqlEngine.java#L110-L176

Example of brittle API, which is not extensible as new retrieve options are added:
https://github.com/cqframework/clinical_quality_language/blob/e3171ced48fd64257dc7bf1a8d5fe09639bd95db/Src/java/engine/src/main/java/org/opencds/cqf/cql/engine/retrieve/RetrieveProvider.java#L6

Example of improper hiding:
https://github.com/cqframework/clinical_quality_language/blob/e3171ced48fd64257dc7bf1a8d5fe09639bd95db/Src/java/engine/src/main/java/org/opencds/cqf/cql/engine/execution/EvaluationResult.java#L9

This story is a place holder for a design task to revise these to be more useful/consistent.

PR #1550, opened 2025-05-14
Add support for Backtraces and profiling

scymtym:
This is a draft for extending the "stack of windows" concept of the `CqlEngine` to a stack of `ActivationFrame`s which track more information than just arguments and local variables, namely

* Expression/function definitions being evaluated in particular activation frames

* Start and end times for the evaluation for activation frames

* Cache hit information for certain activation frames

On top of that extension, this pull request adds two features to the `CQLEngine`

1. Backtraces in CQL exceptions that represent runtime error. The backtrace information can be used by clients to display backtraces with expression/function names, arguments, local variables and source information. Example:

2. Optional profiling. A new engine option can be used to turn enable profiling which the engine to aggregate information from `ActivationFrame`s into a deterministic performance profile that can be used, for example, to render a flamegraph. Example:


c-schuler:
This is very cool, thank you!

A couple of enhancements that I would suggest:

1. Introduce tiered profiling levels (e.g., summary vs. detailed). This being a binary (on/off) is fine for now, but introducing a toggle for verbosity or controlling granularity will likely be a desired enhancement in the future.
2. Only push activation frames if profiling or backtracing is enabled. This avoids overhead in case both are disabled.

As this PR goes through the paces I would also like to ensure the following take place:

- Update the project’s documentation to include usage guidelines for the new backtrace and profiling features.
- Provide configuration settings to enable or disable backtracing and profiling, allowing users to control the overhead associated with these features.
- Assess the performance impact of the added features, especially in production environments, and optimize as necessary.
- Tests should cover scenarios for both backtracing and profiling, including edge cases like nested evaluations and cached results.
- Simulated errors and exceptions should be tested to verify that backtraces and debug logs provide meaningful insights.

With thorough testing and proper documentation, these features will be valuable additions to the project. Thanks again!

scymtym:
@c-schuler Thank you for your feedback and suggestions. I can address some of the suggested changes by myself but will need more guidance for others.

> Introduce tiered profiling levels (e.g., summary vs. detailed). This being a binary (on/off) is fine for now, but introducing a toggle for verbosity or controlling granularity will likely be a desired enhancement in the future.

I can see this as a useful distinction when presenting the results of profiling - the flamegraph-SVG renderer included in the pull request is mainly a demo of one particular result presentation style that works without dependencies and is not too complex (and happened to be sufficient for my use-case). One question could be whether this renderer should be omitted from the pull request so that clients have complete control and responsibility for the presentation of profiling results. None of this has any bearing on runtime performance, of course.

Regarding runtime behavior and the level of activation frames, I'm not sure what a summary level or other finer grained options could look like. The only information in activation frames that is specific to profiling are the start and end times of the respective evaluation and the cache hit/miss bit. The impact on execution time is (presumably) very minor compared to the profiling work of building the call tree and cumulative timings which are controlled by the singular engine option I added.

A related topic is coverage tracking. I remember seeing something about coverage tracking in the codebase but I can't remember where now. Since a flat profile (just call counts, no nesting) also provides coverage tracking, some integration work may be required in that area to avoid multiple mechanisms for the same functionality.

> Only push activation frames if profiling or backtracing is enabled. This avoids overhead in case both are disabled.

Since the stack of activation frames replaces the previous stack of windows, activation frames themselves are not optional.

For clients that can tolerate worse backtraces, there could be an option to only push activation frames when absolutely necessary and only with the absolutely necessary information. I doubt the impact will be large, possibly to the point that querying the engine options frequently would negate the savings, but that would have to be measured.

Also, most of the work for backtraces is done only after an error has already occurred by executing code in `catch` blocks as the stack unwinds, not in advance. Granted, establishing the extra `try/catch` blocks might in itself have a slight performance impact but again repeatedly querying engine options could be the more expensive operation.

> - Update the project’s documentation to include usage guidelines for the new backtrace and profiling features.

I think I can do that. The documentation would mainly explain how get the backtrace and profiling information, not how to present them. Something like the backtrace in the above screenshots needs some work on the client's part. The flamegraph in the screenshot can be produced with just the code in this pull request - provided we don't decide against having the rendering code in the `Profile` class.

> Provide configuration settings to enable or disable backtracing and profiling, allowing users to control the overhead associated with these features.

Such an option exists for profiling. Activation frames cannot be turned off completely since their main purpose, holding variable bindings, is not optional. Reducing the number of activation frames could be an option, as I mentioned above, if it turns out to have a noticeable performance impact.

> Assess the performance impact of the added features, especially in production environments, and optimize as necessary.

I would need help with this one since I don't have access to any relevant production environments.

> Tests should cover scenarios for both backtracing and profiling, including edge cases like nested evaluations and cached results.

Since the existing tests exercise those edge cases and activation frames are used throughout, the main missing piece would be running the existing tests with profiling enabled. Any suggestions regarding how I could do that?

> Simulated errors and exceptions should be tested to verify that backtraces and debug logs provide meaningful insights.

I can add test cases that validate the information in backtraces.

scymtym:
@c-schuler With the most recent commits, the points you brought up should be mostly addressed: 

* **Options for enabled and disabling the features**: I explained above that a single engine option should be enough.
* **Documentation**: I completed the javadoc documentation of the added classes. I'm not sure whether there are other places in which the features should be documented.
* **Performance Impact**: I ran `CqlPerformanceTest` for three configurations: without my changes, with my changes + profiling disabled and with my changes + profiling enabled. There were no significant differences in execution time. I also fixed some of the messages about unexpected results in `CqlPerformanceTest` (see [#cql chat](https://chat.fhir.org/#narrow/channel/179220-cql/topic/Extending.20the.20CQL.20reference.20implementation.20to.20track.20.22calls.22/near/522164769)) to see whether that impacted execution time of the test, but there didn't seem to be a significant impact either.
* **Unit Tests**: I added unit tests for the backtrace and profiling features

I haven't pushed the change that allows `CqlPerformanceTest` to run with profiling enabled since it seems like I should wait for #1556 (See [discussion in #cql](https://chat.fhir.org/#narrow/channel/179220-cql/topic/Extending.20the.20CQL.20reference.20implementation.20to.20track.20.22calls.22/near/523145841)).

Src/java/engine/src/main/java/org/opencds/cqf/cql/engine/exception/Backtrace.java
```diff
@@ -0,0 +1,225 @@
...
+            this.definition = definition;
+            this.arguments = arguments;
+            this.localVariables = localVariables;
+            this.contextName = contextName;
+            this.contextValue = contextValue;
+        }
+
+        /**
+         * Returns the definition in which the current expression was
+         * being evaluated when the backtrace frame was capture.
```
antvaset, line 85:
```suggestion
         * being evaluated when the backtrace frame was captured.
```

Src/java/engine/src/main/java/org/opencds/cqf/cql/engine/execution/State.java
```diff
@@ -177,24 +245,48 @@ public DateTime getEvaluationDateTime() {
...
-        if (!windows.peek().isEmpty()) getStack().pop();
+        final var topActivationFrame = this.stack.peek();
+        assert topActivationFrame != null; // stack underflow
+        topActivationFrame.variables.pop();
     }
 
     public void push(Variable variable) {
-        getStack().push(variable);
+        final var topActivationFrame = this.stack.peek();
+        assert topActivationFrame != null; // stack underflow
```
antvaset, line 267:
```suggestion
        if (topActivationFrame == null) {
            throw new IllegalStateException("Stack underflow: No activation frame available.");
        }
```

Src/java/engine/src/main/java/org/opencds/cqf/cql/engine/execution/State.java
```diff
@@ -297,7 +424,10 @@ public void popEvaluatedResourceStack() {
     }
 
     public Object resolveAlias(String name) {
-        for (Variable v : getStack()) {
+        // This method needs to account for multiple variables on the stack with the same name
+        ArrayList<Object> ret = new ArrayList<>();
+        boolean isList = false;
```
antvaset, line 429:
```suggestion
```

Src/java/engine/src/main/java/org/opencds/cqf/cql/engine/exception/Backtrace.java
```diff
@@ -0,0 +1,225 @@
...
+        // been recorded: the innermost expression, the containing
+        // definition or function and all local variables and
+        // arguments.
+        if (!frames.isEmpty()) {
+            final var currentFrame = this.frames.get(frames.size() - 1);
+            if (currentFrame instanceof FunctionoidFrame) {
+                if (((FunctionoidFrame) currentFrame).getDefinition() == containingDefinition) {
+                    return;
+                }
+            }
```
antvaset, line 184: SonarQube flags this as a maintainability issue, so:
```suggestion
            if (currentFrame instanceof FunctionoidFrame functionoidFrame) {
                if (functionoidFrame.getDefinition() == containingDefinition) {
                    return;
                }
            }
```

Src/java/engine/src/main/java/org/opencds/cqf/cql/engine/execution/EvaluationVisitor.java
```diff
@@ -8,11 +8,49 @@
...
+            exception
+                    .getBacktrace()
+                    .maybeAddFrame(
+                            (ExpressionDef) frame.element,
+                            frame,
+                            state.getStack(),
+                            state.getCurrentContext(),
+                            state.getCurrentContextValue(),
+                            expression);
+        }
```
antvaset, line 51:
```suggestion
        if (frame.element instanceof ExpressionDef expressionDef) {
            exception
                    .getBacktrace()
                    .maybeAddFrame(
                            expressionDef,
                            frame,
                            state.getStack(),
                            state.getCurrentContext(),
                            state.getCurrentContextValue(),
                            expression);
        }
```

Src/java/engine/src/main/java/org/opencds/cqf/cql/engine/execution/Profile.java
```diff
@@ -0,0 +1,356 @@
...
+                result.append(String.format(
+                        "[%s]", ((Retrieve) expression).getDataType().getLocalPart()));
+            } else if (expression instanceof ExpressionDef) {
+                result.append(((ExpressionDef) expression).getName());
+                if (expression instanceof FunctionDef) {
+                    result.append("()");
+                }
+            } else {
+                result.append("unknown expression type");
+            }
```
antvaset, line 186:
```suggestion
            if (expression == null) {
                result.append("«root»");
            } else if (expression instanceof Retrieve retrieve) {
                result.append(String.format(
                        "[%s]", retrieve.getDataType().getLocalPart()));
            } else if (expression instanceof ExpressionDef expressionDef) {
                result.append(expressionDef.getName());
                if (expression instanceof FunctionDef) {
                    result.append("()");
                }
            } else {
                result.append("unknown expression type");
            }
```

Src/java/engine/src/main/java/org/opencds/cqf/cql/engine/execution/State.java
```diff
@@ -18,6 +18,68 @@
...
+            if (this.element == null) {
+                result.append("«root»");
+            } else if (this.element instanceof ExpressionDef) {
+                result.append(((ExpressionDef) this.element).getName());
+            } else if (this.element instanceof Retrieve) {
+                result.append(String.format(
+                        "[%s]", ((Retrieve) this.element).getDataType().getLocalPart()));
+            } else {
+                result.append(this.element.getClass().getSimpleName());
+            }
```
antvaset, line 70:
```suggestion
            if (this.element == null) {
                result.append("«root»");
            } else if (this.element instanceof ExpressionDef expressionDef) {
                result.append(expressionDef.getName());
            } else if (this.element instanceof Retrieve retrieve) {
                result.append(String.format(
                        "[%s]", retrieve.getDataType().getLocalPart()));
            } else {
                result.append(this.element.getClass().getSimpleName());
            }
```

Src/java/engine/src/test/java/org/opencds/cqf/cql/engine/execution/ProfilingTest.java
```diff
@@ -0,0 +1,237 @@
...
+                    final var matcher = e.getValue();
+                    final var childElement = contextChildren.keySet().stream()
+                            .filter(expression -> {
+                                if (expression instanceof ExpressionDef) {
+                                    return ((ExpressionDef) expression)
+                                            .getName()
+                                            .equals(expressionName);
+                                } else {
+                                    return false;
+                                }
```
antvaset, line 78:
```suggestion
                                if (expression instanceof ExpressionDef expressionDef) {
                                    return expressionDef
                                            .getName()
                                            .equals(expressionName);
                                } else {
                                    return false;
                                }
```

antvaset:
Hi @scymtym , this is so cool and very well done! And thanks for merging in my suggested changes.

Agree with @c-schuler that more toggles will be needed to control future profiling features like verbosity and granularity. In terms of what those toggles could look like, this TODO is captured by https://github.com/cqframework/clinical_quality_language/issues/1287 and will be addressed separately and perhaps more holistically. We don't currently have a README for CQL engine options in this repo (only code comments), and I can do that separately as well for all engine options (see new issue https://github.com/cqframework/clinical_quality_language/issues/1574).

antvaset (approved):

scymtym:
@antvaset Thank you for the review.

I agree that some way of controlling the profiling feature, that is capturing profiling info as well as processing the finished profile, in more detail will be needed. The ad-hoc flamegraph-SVG writer I included is more a demo than anything. Whether those fleshed out details should take the form of engine options, I'm not really sure but that can be worked later as you suggest.

PR #1556, opened 2025-05-18
Improved implementation of the `expand` operator

antvaset:
According to the spec for the `expand` operator:

> If the interval boundaries are more precise than the per quantity, the more precise values will be truncated to the precision specified by the per quantity.

This PR improves the current implementation but I'll need to look into it further and fully test all potential edge cases.


antvaset:
This is failing the CqlMainSuiteTest. In draft state until fixed.

JPercival (approved):

Src/java/engine/src/test/java/org/opencds/cqf/cql/engine/runtime/DateTimeTest.java
```diff
@@ -368,4 +368,74 @@ void dateTimeEquals() {
 
         // TODO: It'd be good to extend these across different types of constructors, for example, from a string
     }
+
+    @Test
+    void roundToPrecision() {
```
lukedegruchy, line 373: Suggestion:  try with different timezones and bonus points for a weird half-hour timezone, like America/St_Johns

lukedegruchy (approved):

PR #1576, opened 2025-07-01
#1575: Fixed the implementation of `equivalent` for ratios

### LINKED ISSUE #1575: `Equivalent` for `Ratio` types should use value of the ratio, not compare the numerator and denominator.
According the the spec, ration should be equivalent if they represented the same value:

https://cql.hl7.org/09-b-cqlreference.html#equivalent

> For ratios, equivalent means that the numerator and denominator represent the same ratio (e.g. 1:100 ~ 10:1000).

Currently, rather than calculate the value, the current implementation checks the numerator and denominator independently:

https://github.com/cqframework/clinical_quality_language/blob/master/Src/java/engine/src/main/java/org/opencds/cqf/cql/engine/runtime/Ratio.java#L29-L39

Current:
```
(1 ~ 10) && (10 ~ 1000)
```

Desired:
```
(1 / 10) ~ (10 / 1000)
```

Reported on Zulip by @scymtym:
[#cql > Is Equivalent (~) wrong for Ratio?](https://chat.fhir.org/#narrow/channel/179220-cql/topic/Is.20Equivalent.20.28~.29.20wrong.20for.20Ratio.3F/with/522627993)

antvaset:
Closes #1575:

* Fix the `equivalent` logic for ratios to make e.g. `1'cm':2'cm' ~ 2'cm':4'cm'` evaluate to `true`. The result is calculated as `equivalent(a.numerator * b.denominator, b.numerator * a.denominator)`.
* Fix the `divide` evaluator to make dividing by a 0-valued quantity return `null`. Previously it threw a NullPointerException.

### DIFF
--- a/Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/model/ModelImporter.java
+++ b/Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/model/ModelImporter.java
@@ -3,7 +3,6 @@ package org.cqframework.cql.cql2elm.model;
 import java.util.*;
 import org.cqframework.cql.cql2elm.ModelManager;
 import org.hl7.cql.model.*;
-import org.hl7.cql.model.NamespaceManager;
 import org.hl7.elm_modelinfo.r1.*;
 
 public class ModelImporter {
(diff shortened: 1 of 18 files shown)


brynrhodes (approved):

scymtym:
Sorry I didn't communicate this in the bug report (only [here](https://chat.fhir.org/#narrow/channel/179220-cql/topic/Is.20Equivalent.20.28~.29.20wrong.20for.20Ratio.3F/near/526571652)) but I have a branch that fixes `Equivalent` for `Ratio` and `Quantity` as well as the other comparison operators and arithmetic operators for `Quantity` by adding the required unit conversions. Should I submit that pull request in parallel or wait for this one?

scymtym:
I submitted my branch as a [pull request](https://github.com/cqframework/clinical_quality_language/pull/1578) for reference.

antvaset:
Hi @scymtym I'd tackle this separately from the broader "unit-aware quantity operations" feature, so this PR is good to go first while yours is in review. I can jump on to reviewing your PR next (which is very impressive!).

I did like your `equivalent` implementation better so I've changed this PR to use `MultiplyEvaluator`.

JPercival (approved):

PR #1578, opened 2025-07-03
---

scymtym:
draft: true
---
# Perform unit conversion in equality, comparison and arithmetic operators for Quantity (and Ratio)

The pull request mostly achieves what is claimed but has draft status for now because of the following issues:

* `RuntimeTest.intervalOfQuantityWithDifferentUOM` fails now because the expected `InvalidInterval` exception is not thrown.
  I think the test is wrong because both interval endpoints have the unit "mass per volume" (although spelled scaled differently) and should be acceptable for defining an `Interval<Quantity>`.

* In the test `CQLOperationsR4Test.test`, the test case `testQuantity/testQuantity4` fails.
  The test case asserts that `4 'g' ~ 4040 'mg'` is true.
  I can only imagine this being true due to some precision rule, but I had the impression that for unit conversion, everything has to be converted to the "more granular" unit which would mean `4000 'mg' ~ 4040 'mg'` which has to be false.

* There are a few TODOs, mostly related to test improvements, that I would address if this pull request is otherwise acceptable.


scymtym:
@antvaset  I merged master into this a few more times but I'm not sure what the next step is. Is there anything I should be doing to make progress with this?

scymtym:
Updated to apply to the Kotlin conversion of the codebase as discussed in Zulip chat. All tests that I could run locally, including many newly enabled tests, pass for me. The issues mentioned in the original pull request description have been resolved.

PR #1596, opened 2025-08-11
evaluatedResources Bug CqlEngine

Capt-Mac:
evaluatingExpressions method had a bug that was accumulating evaluated resources from CQL expressions instead of clearing them between top level evaluated executions. This patch will allow for expected behavior of evaluated resources and fix unit test that was checking for regressions


Src/java/engine-fhir/src/test/java/org/opencds/cqf/cql/engine/fhir/data/EvaluatedResourcesTest.java
```diff
@@ -21,11 +21,9 @@ void withCache() {
         var results = engine.evaluate(library.getIdentifier(), Set.of("Union"));
 
         assertEvaluationResult(results, "Union", List.of(CONDITION, ENCOUNTER));
-        engine.getState().clearEvaluatedResources();
```
lukedegruchy, line 24: Question:  I just want to confirm why we're removing these:  that the CqlEngine is doing this at the appropriate points in the workflow now.

lukedegruchy (approved):

Src/java/engine-fhir/src/test/java/org/opencds/cqf/cql/engine/fhir/data/EvaluatedResourcesTest.java
Capt-Mac: each of these acts as a top level evaluated expression which means the method called for evaluation should handle the clearing of resources between evaluations

Capt-Mac: if a regression comes in later that undoes 'clearing' then this test should fail with too many evaluated resources

PR #1605, opened 2025-09-24
Clean up exception hierarchy

JPercival:
* Makes a CqlIncludeException a type of CqlCompilerException
* Reduces unused overloads of CqlCompilerExceptions
* Use only one list to track all exception types (as opposed to 4)
* Next steps
  * Compiler: Make locator non-nullable, ensure we have context to provide a locator everywhere an exception is thrown
  * Engine: Support returning multiple compiler exceptions, instead of flattening them all into one message


cmoesel:
@JPercival - One of our CQL Translations Service tests just started failing with this change. It is a test for properly returning an error when an included library is missing.

Whereas a missing library used to result in this annotation:
```json
{
  "type": "CqlToElmError",
  "libraryId": "CMS146",
  "libraryVersion": "2",
  "startLine": 5,
  "startChar": 1,
  "endLine": 5,
  "endChar": 37,
  "message": "Could not load source for library CMSAll, version 1, namespace uri null.",
  "errorType": "include",
  "errorSeverity": "error",
  "targetIncludeLibraryId": "CMSAll",
  "targetIncludeLibraryVersionId": "1"
}
```

It now results in this annotation:
```json
{
  "type": "CqlToElmError",
  "libraryId": "CMS146",
  "libraryVersion": "2",
  "startLine": 5,
  "startChar": 1,
  "endLine": 5,
  "endChar": 37,
  "message": "Could not load source for library CMSAll, version 1, namespace uri null.",
  "errorType": "internal",
  "errorSeverity": "error"
}
```

Note that:
* `errorType` changes from `include` to `internal`
* `targetIncludeLibraryId` is no longer provided
* `targetIncludeLibraryVersionId` is no longer provided

I see that `CqlIncludeException` still exists as a specific exception in the Java code, but it seems all its specificity is now removed in the serialization to an annotation. Is that intentional (for example, as part of an effort to normalize annotations to just a few types and common schemas)?

PR #1618, opened 2025-10-03
1617: fix(engine): correct operator precedence in validateTerminology…

amondel2:
1617:   The if condition in validateTerminologyRequirements was incorrectly evaluating due to missing parentheses. Because '&&' has higher  precedence than '||', the expression:

  ` (A || B || C && D)`
 was being evaluated as:
` (A || B || (C && D))`
instead of the intended:
` ((A || B || C) && D) `

This caused terminology requirements to throw an exception in cases when a TerminologyProvider was configured with a library with ValueSets.

Fix: Add explicit parentheses so the check correctly validates as intended.


JPercival:
We'll update in the new Kotlin code. Thanks for the contribution

brynrhodes (approved):

JPercival:
@amondel2 - We couldn't push to your branch to fix the merge conflict, so we took it in #1684

PR #1640, opened 2025-11-14
Initial implementation of AST for CQL

antvaset:
* Parsing CQL source files into ASTs modeled using CQL language constructs (distinct from ELM).
* Implementation of AST walker for read-only traversal of the CQL AST.
* WIP implementation of ELM emitter that transforms CQL AST to ELM.
* Showing the parse tree and AST in the playground.
* KMP for `SystemModelInfoProvider`.
* New `FilesToStringsTask` Gradle task with codegen enabling file contents to be inlined at compile time and used as strings (useful for e.g. inlining `system-modelinfo.xml` for `SystemModelInfoProvider`-in-JS).
* New `LoadTestResourcesTask` Gradle task with multiplatform codegen enabling access to test resources in commonTest.
* Limiting multiplatform test execution to JVM and JS-in-Node.js

### DIFF
--- a/Src/java/engine/src/main/kotlin/org/opencds/cqf/cql/engine/execution/CqlEngine.kt
+++ b/Src/java/engine/src/main/kotlin/org/opencds/cqf/cql/engine/execution/CqlEngine.kt
@@ -450,6 +450,65 @@ constructor(val environment: Environment, engineOptions: MutableSet<Options>? = 
         throw ce
     }
 
+    /**
+     * Resolves the default value of a named parameter in a CQL library.
+     *
+     * This method evaluates the `default` expression of the given parameter definition within the
+     * specified library. It uses the standard engine lifecycle — initializing the library on the
+     * state stack, beginning an evaluation, and cleaning up afterward — so that the evaluation
+     * visitor has the full context it needs (current library, evaluated resource stack, activation
+     * frames, evaluation time).
+     *
+     * If the parameter has no default expression, this method returns `null`.
+     *
+     * **Typical usage** — resolving a CQL-defined default `Measurement Period` before measure
+     * evaluation, so that callers do not need to manipulate the library stack themselves:
+     * ```kotlin
+     * val defaultPeriod = engine.resolveParameterDefault(libraryId, "Measurement Period")
+     * ```
+     *
+     * @param libraryIdentifier the versioned identifier of the library containing the parameter
+     * @param parameterName the name of the parameter whose default to resolve
+     * @param evaluationDateTime the evaluation timestamp; defaults to `ZonedDateTime.now()`
+     * @return the evaluated default value, or `null` if the parameter has no default
+     * @throws CqlException if the library or parameter cannot be resolved, or if the default
+     *   expression evaluation fails
+     */
+    @JvmOverloads
+    fun resolveParameterDefault(
+        libraryIdentifier: VersionedIdentifier,
+        parameterName: String,
+        evaluationDateTime: ZonedDateTime? = null,
+    ): Any? {
+        val library =
+            environment.resolveLibrary(libraryIdentifier)
+                ?: throw CqlException("Unable to resolve library: ${libraryIdentifier.id}")
+
+        val parameterDef = Libraries.resolveParameterRef(parameterName, library)
+        if (parameterDef.default == null) {
+            return null
+        }
+
+        initializeEvalTime(evaluationDateTime)
+        state.init(library)
+        state.beginEvaluation()
+        return try {
+            evaluationVisitor.visitExpression(parameterDef.default!!, state)
+        } catch (ce: CqlException) {
+            processException(ce, parameterDef.default!!)
+        } catch (e: Exception) {
+            processException(
+                e,
+                parameterDef.default!!,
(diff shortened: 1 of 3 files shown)


cmoesel:
I don't have any comments on the code (haven't looked at it), but I'm curious -- what's the motivation/vision for this CQL AST?

JPercival:
> I don't have any comments on the code (haven't looked at it), but I'm curious -- what's the motivation/vision for this CQL AST?

The ultimate goal is to split the compiler into a frontend (parsing, AST generation, type inference, some types of optimization like inlining) and a backend (ELM generation). The motivations for that are multi-fold.
1. The current Cql2ElmVisitor is difficult to maintain. By splitting the functionality in to phases we expect it to be easier to evolve going forward.
    * In particular, ELM generation can be "dumb" given a hypothetical fully type-checked and validated AST structure. The ELM generation itself is simple transform -- it's all the scope book-keeping, library resolution, type checking, etc that come from the CQL side that make it complex.
    * Similarly, intermingling ELM generation makes it more difficult to implement type inference.
    * As a side effect of the above, both the parsing and the ELM generation are more easily unit tested in isolation. See the preliminary unit tests in the PR. This will allow us to shrink the test deck while improving code coverage, there by improving the quality of the codebase overall.
2. CQL has begun to evolve independently of the ELM. An intermediate representation allows us to change one side of the equation without affecting the other.
    * This will allow us to more easily support various revisions of CQL while still generating the same ELM. 
    * For example, all the CQL 2.0 changes obviate the need for ModelInfo entirely. We could internally "rewrite" `using` statements from CQL 1.3-1.5 to be inlined type definitions. This allows us to have "one" way manage types in the CQL compiler that we need to maintain, while being transparent to both the CQL users and the ELM consumers
3. It allows better IDE tools to be built.
    * Currently, the VS Code plugin is driven off of the ELM graph. Certain syntax subtleties like `on or before end of` are lost, and therefore we can't effectively support autocomplete of that statement. Hooking in "earlier" in the compilation process allows us to assist with syntax, not just semantics.
    * Similarly, it may be faster without need to go through the full ELM generation.
4. It allows the CQL compiler to target things beside ELM.
    * For example, it could emit FHIR-on-SQL statements based on the AST, or .Net bytecode, or whatever.

cmoesel:
> The ultimate goal is to split the compiler into a frontend (parsing, AST generation, type inference, some types of optimization like inlining) and a backend (ELM generation). The motivations for that are multi-fold...

Thanks for additional insight. That's helpful. It's been a while since I've written any significant CQL-to-ELM code (probably nearing a decade). I didn't realize the ELM model itself caused so much complexity -- I assumed it was mainly the type inference, function resolution, and other various CQL _magic_, but it sounds like that stuff is just shifting to the AST generator now. That said, ELM was based on Health e-Decisions, so I shouldn't be surprised that conforming to that schema might introduce additional complications.

There's not much to argue with here: better separation of concerns, better testing, better consistency across versions, better IDE tools, and easier translation into other target formats. I like it.

One small note: I wanted to see what the AST looks like, so I built it and ran the CQL playground. In the process of doing that, I found that the CQL-to-ELM UI README needs to be updated. It currently says to browse to http://localhost:3000/cql-to-elm-ui, but that should be http://localhost:3000/clinical_quality_language/playground now.

antvaset:
Hi @cmoesel, sorry the README got outdated after we deployed the UI to cqframework.org. Now it's fixed.

JPercival:
>There's not much to argue with here... I like it.

Is that "I like the idea" or "I like the AST implementation?"

cmoesel:
> > There's not much to argue with here... I like it.
> 
> Is that "I like the idea" or "I like the AST implementation?"

I like the idea. I haven't looked at the implementation closely enough to comment on that. Is there any documentation that lists all of the possible AST nodes and their relationships?

JPercival:
@cmoesel - Docs updated. A few key ones:

[Src/java/cql/README.md](https://github.com/cqframework/clinical_quality_language/pull/1640/files#diff-bd85865c55550467a6dc2973cddf15f2927b2e0a0aa27fc2392048dad32edc72)

[Src/java/docs/DECISIONS.md](https://github.com/cqframework/clinical_quality_language/pull/1640/files#diff-2de0d03f1b56efa07cc153524cef0986197008e8e7bfb4c902f52a524ddd7ec1)

[Src/java/cql/AST.md](https://github.com/cqframework/clinical_quality_language/pull/1640/files#diff-32050e9405418593939a2c3690aaa214105c0961268e3dabba7cbca20b149859)

brynrhodes:
I don't disagree with the AST approach here, there are some good arguments there, but I disagree with the premise of number 3, where are you seeing that "CQL has begun to evolve independently of ELM"? I don't think that's true at all, Every change that has been proposed in CQL 2.0 has a corresponding change in the ELM representation. I don't think it changes the thrust of the arguments, but I want to clarify the statement.

JPercival:
@brynrhodes - I don't see that in the AST docs anywhere? Was there some particular statement that needed to be clarified?

brynrhodes:
It's not in the docs so far as I know, I was referring to the comment thread above (https://github.com/cqframework/clinical_quality_language/pull/1640#issuecomment-3549675987)

PR #1647, opened 2025-11-21
Rev to 4.1, update to HAPI 8.6

JPercival:


Src/java/cqf-fhir/src/main/kotlin/org/cqframework/fhir/utilities/IGContext.kt
```diff
@@ -46,7 +46,7 @@ open class IGContext {
 
     var sourceIg: ImplementationGuide? = null
 
-    var fhirVersion: String? = null
+    var fhirVersion: String = "4.0.1"
```
lukedegruchy: Question:  why are we defaulting to 4.0.1 now?

PR #1656, opened 2025-12-11
Add evaluation result setter

antvaset:
Adds a setter to map-like `EvaluationResult`. This is needed e.g. in CR tests where `EvaluationResult`s are put together manually.

### DIFF
--- a/Src/java/engine/src/main/kotlin/org/opencds/cqf/cql/engine/execution/EvaluationResult.kt
+++ b/Src/java/engine/src/main/kotlin/org/opencds/cqf/cql/engine/execution/EvaluationResult.kt
@@ -19,9 +19,14 @@ class EvaluationResult {
         return expressionResults[name]
     }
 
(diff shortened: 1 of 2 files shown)


Src/java/engine/src/main/kotlin/org/opencds/cqf/cql/engine/execution/EvaluationResult.kt
```diff
@@ -19,10 +19,15 @@ class EvaluationResult {
...
     }
 
-    /** Returns the ExpressionResult for the given expression name. */
+    /** Returns the ExpressionResult for the given expression reference. */
     operator fun get(ref: EvaluationExpressionRef): ExpressionResult? {
         return results[ref]
     }
 
+    /** Sets the ExpressionResult for the given expression reference. */
+    operator fun set(ref: EvaluationExpressionRef, result: ExpressionResult) {
```
lukedegruchy: Please add a short description to the PR

PR #1659, opened 2026-01-06
Fixed NPM model info provider not using ModelInfoReaderFactory

brynrhodes:


brynrhodes:
These failures seem to be intermittent, and they do not occur locally. In addition, this is for a v3 hotfix, so I'm going to merge it

PR #1713, opened 2026-03-25
CQL v5

antvaset:
* All non-null CQL values are represented as instances (implementors) of the sealed `Value` interface. `Boolean`, `Integer`, `Long`, `Decimal`, `String`, and `List` from the `org.opencds.cqf.cql.engine.runtime` package are used intead of `kotlin.Boolean`, `kotlin.Int`, `kotlin.Long`, `BigDecimal`, `kotlin.String`, and `kotlin.collections.Iterable<T>`.
* The evaluation visitor, all evaluators, engine API, `FhirTypeConverter`, and related tests are updated accordingly to use `Value` instead of `Any`.
* JUnit and Hamcrest are replaced with kotlin.test in the majority of tests along the way.
* The original `Value` object is renamed to `Constants`. The `CqlList` class is replaced with the `SortHelper.compare()` method.
* Non-system structured types (e.g. instances of `FHIR.Patient`, `FHIR.string`, etc.) are now represented in the engine as `ClassInstance`s. A `ClassInstance` is a structured value (like a Tuple) with a QName type tag like `{http://hl7.org/fhir}Patient`. As a result:
  * the engine only operates on a fixed set of Kotlin types, and HAPI FHIR structures and other custom types do not bleed into the runtime,
  * `resolvePath()`, `as()`, `setValue()`, `objectEqual()`, `objectEquivalent()`, `resolveType()` are no longer needed in the `ModelResolver` interface.
* Java `Class<T>` references are replaced with type specifiers and type QNames (both multiplatform) in the engine. As a result:
  * `packageName(s)` is no longer needed in `ModelResolver` and `is(value: Any?, type: Class<*>?): Boolean?` changed to `is(valueType: String, type: QName): Boolean?`,
  * type testing is improved so e.g. `Interval[1,2] is Interval<Integer>` returns true and `Interval[1,2] is Interval<Time>` returns false,
  * function resolution is more robust when function refs are compiled without signatures.
* In `FhirModelResolver : ModelResolver`, the `is(valueType: String, type: QName): Boolean?` method now returns `true` only if `{http://hl7.org/fhir}_valueType_` extends `type` or is the same as `type`. As a result, expressions like `FHIR.integer { ... } is FHIR.positiveInt` and `FHIR.Quantity { ... } is FHIR.Age` return false because `FHIR.integer` is a supertype of `FHIR.positiveInt` and `FHIR.Quantity` is a supertype of `FHIR.Age`.
* Fix `DescendentsEvaluator`.


JPercival (approved):

Zylox:
We are currently using a custom modelinfo and a custom modelresolver to implement cql over internal data types that are not fhir related.  This is well supported currently. These changes to modelresolver seem to make that impossible now. Is this engine only going to support FHIR, specifically hapi fhir, or is there some other avenue to get existing behavior that i am missing.

Fully extracting every record into a CqlClassInstance would be prohibitively poorly performing for us. Right now, the fact that we can retrieve objects at request time with `resolvePath` is very useful. 

If CqlClassInstance were an interface that allowed similar behavior we could make this work, maybe. You could even then implement the interface by something identical to the current implementation if for some reason that is performant for you, where `cqlClassInstanceImpl.resolvePath("value")` performs a get on `elements`, and others like my team can back it with our existing data structures. This approach still wouldnt be ideal as we would have to wrap every object in an adapter, but its preferable to having to fork off the engine.

antvaset:
Hi @Zylox,

> These changes to modelresolver seem to make that impossible now. Is this engine only going to support FHIR, specifically hapi fhir, or is there some other avenue to get existing behavior that i am missing.

The engine still supports any CQL model (models are made of types described in the [spec](https://build.fhir.org/ig/HL7/cql/03-developersguide.html#types)). I've documented all supported types in `Src/java/engine/cql-value-types.md` and how they are represented in the engine. `CqlClassInstance` is a container for named structured types (also referred to as [class types](https://build.fhir.org/ig/HL7/cql/03-developersguide.html#defining-class-types) or [typed tuples](https://build.fhir.org/ig/HL7/cql/02-authorsguide.html#structured-values-tuples)), and they aren't FHIR-specific or HAPI FHIR-specific.

> Fully extracting every record into a CqlClassInstance would be prohibitively poorly performing for us. Right now, the fact that we can retrieve objects at request time with `resolvePath` is very useful.
> 
> If CqlClassInstance were an interface that allowed similar behavior we could make this work, maybe. You could even then implement the interface by something identical to the current implementation if for some reason that is performant for you, where `cqlClassInstanceImpl.resolvePath("value")` performs a get on `elements`, and others like my team can back it with our existing data structures. This approach still wouldnt be ideal as we would have to wrap every object in an adapter, but its preferable to having to fork off the engine.

The `elements` of `CqlClassInstance` can be anything that implements `MutableMap<String, Any?>`, so you could instantiate `CqlClassInstance` with something like `myLazyMap` that only extracts the underlying records upon request. It would therefore materialize incrementally, as needed by the engine during evaluation, e.g. load the value of a specific element when the `Property` operator is applied or enumerate all elements when `Descendants` is applied.

antvaset:
The source branch `engine-structured-types` has been renamed to `v5` and this PR was automatically closed. https://github.com/cqframework/clinical_quality_language/pull/1755 is the new pull request.

PR #1755, opened 2026-05-01
CQL Engine v5

antvaset:
This PR includes multiple changes set to be released in v5. The changes are further documented in the [v4 to v5 migration guide](https://github.com/cqframework/clinical_quality_language/blob/v5/Src/java/engine/v4-to-v5-migration.md) included in the PR.

## Changes to CQL value classes (#1775, [ADR 005](https://github.com/cqframework/clinical_quality_language/blob/v5/Src/java/docs/DECISIONS.md#adr-005-cql-value-representation-in-the-kmp-engine))

* All non-null CQL values are represented in the engine as instances (implementors) of the sealed `Value` interface. `Boolean`, `Integer`, `Long`, `Decimal`, `String`, and `List` from the `org.opencds.cqf.cql.engine.runtime` package are used intead of `kotlin.Boolean`, `kotlin.Int`, `kotlin.Long`, `BigDecimal`, `kotlin.String`, and `kotlin.collections.Iterable<kotlin.Any?>`.
* `Value.toString()` returns the CVL representation of the CQL value (see #1782).
* The evaluation visitor, evaluators, engine API, `FhirTypeConverter`, and related tests are updated accordingly to use `Value?` instead of `kotlin.Any?`.

## New `ClassInstance` class for CQL class instances, changes to `ModelResolver`, Java platform independence

* Non-system structured types (e.g. instances of `FHIR.Patient`, `FHIR.string`, etc.) are now represented in the engine as `ClassInstance`s (and not as e.g., HAPI FHIR structures). Informally, a `ClassInstance` is a structured value (Tuple) with a `QName` type tag like `{http://hl7.org/fhir}Patient`. As a result:
  * HAPI FHIR structures and other external classes do not bleed into the engine,
  * `resolvePath()`, `as()`, `setValue()`, `objectEqual()`, `resolveType()` are no longer needed in the `ModelResolver` interface.
* Java `Class<T>` references are replaced with type specifiers and type QNames (both multiplatform) in the engine. As a result:
  * `packageName(s)` is no longer needed in `ModelResolver` and `is(value: Any?, type: Class<*>?): Boolean?` changed to `is(valueType: String, type: QName): Boolean?`,
  * type testing is improved so e.g. `Interval[1,2] is Interval<Integer>` returns true and `Interval[1,2] is Interval<Time>` returns false,
  * function resolution is more robust when function refs are compiled without signatures.

## Additional changes, fixes, improvements
* In `FhirModelResolver : ModelResolver`, the `is(valueType: String, type: QName): Boolean?` method now returns `true` only if `{http://hl7.org/fhir}_valueType_` extends `type` or is the same as `type`. As a result, expressions like `FHIR.integer { ... } is FHIR.positiveInt` and `FHIR.Quantity { ... } is FHIR.Age` return false because `FHIR.integer` is a supertype of `FHIR.positiveInt` and `FHIR.Quantity` is a supertype of `FHIR.Age`.
* JUnit and Hamcrest are replaced with kotlin.test in the majority of tests along the way.
* The original `Value` object is renamed to `Constants`. The `CqlList` class is replaced with the `SortHelper.compare()` method.
* `DescendentsEvaluator` logic is fixed along the way.

## Enhancements for the engine JS target and CQL playground

* Initial implementation of FHIR model resolver in JS with support for `createInstance()` and `is()` methods.
* JS-export CQL value classes, CQL type classes.
* Add type hints for expression results in the online playground.
image


Src/java/engine/src/commonMain/kotlin/org/opencds/cqf/cql/engine/elm/executing/EquivalentEvaluator.kt
```diff
@@ -128,32 +144,42 @@ object EquivalentEvaluator {
...
         }
 
-        // Fallback to data provider's `objectEquivalent()`
-
-        if (state != null) {
-            return state.environment.objectEquivalent(left, right)
+        if (left is ClassInstance && right is ClassInstance) {
+            if (left.type == right.type) {
+                val dataProvider =
+                    state!!.environment.resolveDataProviderByModelUri(left.type.getNamespaceURI())
```
c-schuler: By removing the null check for state, a possible NPE exists here. I believe it also is possible on line 194 as well (quantitiesEquivalent).

c-schuler (approved):
Really nice work! I really only have some small nits. I noted one of them in the EquivalentEvaluator and another is below.

The EqualEvaluator, EquivalentEvaluator, the four comparison evaluators (Greater/Less/…), and the binary-arithmetic evaluators each repeat a long when/is ladder over the same set of Value subtypes, differing only in the operation. Adding a new Value type requires editing ~10+ files. The new sealed hierarchy is exactly the structure that would let this move to polymorphic methods on Value (e.g. compareTo, equalTo). Not a blocker, but the refactor stopped one level short of the payoff it sets up.

antvaset:
> The EqualEvaluator, EquivalentEvaluator, the four comparison evaluators (Greater/Less/…), and the binary-arithmetic evaluators each repeat a long when/is ladder over the same set of Value subtypes, differing only in the operation. Adding a new Value type requires editing ~10+ files. The new sealed hierarchy is exactly the structure that would let this move to polymorphic methods on Value (e.g. compareTo, equalTo). Not a blocker, but the refactor stopped one level short of the payoff it sets up.

Hi @c-schuler thank you! At this stage, I think it's OK to keep the `Value` interface and classes as simple and dumb and possible. This helps with the transition and going forward, may be beneficial because we have a clear separation of concerns in which `Value` packages the data and evaluators implement the CQL equality, CQL arithmetic, etc. logic on `Value` instances. I like the idea of removing duplicate code so can look separately into adding more helper and extension functions within the scope of `org.opencds.cqf.cql.engine.elm.executing` so they can be shared by multiple evaluators.

raleigh-g-thompson (approved):
Looks good.

barhodes (approved):

JPercival (approved):

PR #1757, opened 2026-05-05
Added selective requirements implementation

brynrhodes:
Added selective requirements implementation, consistent with [FHIR-56426](https://jira.hl7.org/browse/FHIR-56426)

Selectivity is introduced as a concept to describe how "selective" a set of data requirements is with respect to a particular expression. Conceptually, the selectivity describes any expression in Conjunctive or Disjunctive Normal Form, and conveys the set of data requirements referenced by sargeable terms in the expression. The "coverage" of the selectivity is whether the clauses and sargeable terms of the expression are a complete representation of the expression, or only part of it (i.e. either the expression overall is not in CNF/DNF, or it contains more than just sargeable terms). The "inclusivity" of the selectivity is whether the expression represents an inclusion or exclusion criteria with respect to the context of the expression.

The selectivity of an expression can be used to support query planning by identifying the search parameters that would be most "selective" for narrowing the set of subjects. This is used in particular in GroupDefinition and Group/$evaluate operations downstream in the clinical reasoning module to support group resources as cohort definitions, as well as to support Group-based evaluation for Libraries, PlanDefinitions, and Measures.

This initial implementation is functional with respect to:

1. Conjunctive queries only (all clauses and terms are conjunctive, disjunctive is not supported, it needs union rewrite, or at the very least data requirement duplication)
2. Inclusion and exclusion (i.e. terms with exists and not exists)
3. Parameters as RValues in terms
4. Inference through non-cardinality-changing conversion operators (same as data requirements inference generally)

The test cases cover these functions, but more test cases are planned for support (TestCase11.cql).


Src/java/cql-to-elm/src/commonMain/kotlin/org/cqframework/cql/cql2elm/CqlCompilerOptions.kt
```diff
@@ -276,6 +277,18 @@ class CqlCompilerOptions() {
         return this
     }
 
+    /**
+     * git Return this instance of CqlTranslatorOptions with addition of newly assigned
```
antvaset:
```suggestion
     * Return this instance of CqlTranslatorOptions with addition of newly assigned
```

Src/java/elm-fhir/src/main/kotlin/org/cqframework/cql/elm/requirements/ElmQuerySelectivity.kt
```diff
@@ -0,0 +1,84 @@
...
+            "inclusion"
+        ), // Indicates the query selectivity represents an inclusion criteria with respect to the
+        // context of the query
+        EXCLUSION(
+            "exclusion"
+        ), // Indicates the query selectivity represents an exclusion criteria with respect to the
+        // context of the query
+        INDETERMINATE(
+            "indeterminate"
+        ), // Indicates that whether the query selectivity represents inclusion or exclusion criteria
```
antvaset:
```suggestion
        ), // Indicates that the query selectivity represents inclusion or exclusion criteria
```

antvaset (approved):

sliver007 (approved):

PR #1758, opened 2026-05-05
Treat `null`s as empty lists in the list passed to `flatten()`

### LINKED ISSUE #1748: Incorrect interpretation of flatten{}
The Clinical Quality Language engine seems to be misinterpreting the behavior of flatten {}

define "flatten null": // returns `[ null, 1 ]`
    flatten(
        {
            null as List<System.Integer>,
            { 1 }
        }
    )
define "flatten {}": // returns `[ 1 ]`
    flatten(
        {
            {} as List<System.Integer>,
            { 1 }
        }
    )

The expectation is that null should be treated as an empty list

antvaset:
Closes #1748.

In the list passed in to `flatten()`, `null`s need to be treated as empty lists.

### DIFF
--- a/Src/java/engine/src/commonMain/kotlin/org/opencds/cqf/cql/engine/elm/executing/FlattenEvaluator.kt
+++ b/Src/java/engine/src/commonMain/kotlin/org/opencds/cqf/cql/engine/elm/executing/FlattenEvaluator.kt
@@ -19,9 +19,7 @@ object FlattenEvaluator {
         if (operand is Iterable<*>) {
             val resultList = mutableListOf<Any?>()
             for (element in operand) {
-                if (element == null) {
-                    resultList.add(null)
-                } else {
+                if (element != null) {
                     for (subElement in element as Iterable<*>) {
                         resultList.add(subElement)
                     }
--- a/Src/java/build-logic/build.gradle.kts
+++ b/Src/java/build-logic/build.gradle.kts
@@ -21,6 +21,7 @@ dependencies {
     )
     implementation("org.openrewrite.rewrite:org.openrewrite.rewrite.gradle.plugin:7.26.0")
     implementation("com.diffplug.spotless:spotless-plugin-gradle:8.4.0")
+    implementation("org.jetbrains.kotlinx:kover-gradle-plugin:0.9.8")
 }
 
 kotlin { jvmToolchain(17) }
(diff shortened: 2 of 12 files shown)


Src/java/engine/src/commonMain/kotlin/org/opencds/cqf/cql/engine/elm/executing/FlattenEvaluator.kt
```diff
@@ -19,9 +19,7 @@ object FlattenEvaluator {
         if (operand is Iterable<*>) {
             val resultList = mutableListOf<Any?>()
             for (element in operand) {
-                if (element == null) {
-                    resultList.add(null)
-                } else {
+                if (element != null) {
```
brynrhodes, line 22: We can't universally ignore nulls here, only in the case that the null is typed by the invocation as List-valued.

For example, this test case currently passes in all engines, and needs to continue to pass:

`flatten { { null }, { null } }` results in `{ null, null }`

This change would mean this test would no longer be passing.

And I'm actually not sure how we're not failing this test already? Looking at the tests in engine-fhir, this test is explicitly included and so shouldn't this be failing?

brynrhodes (changes requested):

Src/java/engine/src/commonMain/kotlin/org/opencds/cqf/cql/engine/elm/executing/FlattenEvaluator.kt
antvaset: Hi @brynrhodes , we only flatten "one-level deep", and the change only affects how the null-valued immediate sublists are handled (which we now ignore and effectively treat as empty lists). All the elements from the sublists are added to the result as is.

antvaset: Because the signature is `flatten(argument List<List<T>>) List<T>`, in the title I mean literally "Treat `null`s as empty lists in the list passed to `flatten()`" :)

brynrhodes: Gotcha, and that's why the other test is passing

brynrhodes (approved):

JPercival (approved):

PR #1759, opened 2026-05-05
`in` (ValueSet) for the `String` overload: check only the code value and throw if the resolved value set contains codes from multiple code systems

### LINKED ISSUE #1749: Observation.status missing system
The Clinical Quality Language engine has been giving incorrect results when using a ValueSet for Observation.status. We are getting "null" results due to the system being missing from a code type. The system being used in the ValueSet being called is the same one that is required by the [FHIR spec](https://hl7.org/fhir/R4/valueset-observation-status.html). Because of this, we would still expect the engine to return results if the Observation.status code is in the ValueSet, even though Observation.status is of type `code`. 
Here is a minimal repro of the issue:
```
library AAA version '2025.2.1'

using FHIR version '4.0.1'

include FHIRHelpers version '4.0.1' called FHIRHelpers

valueset "Final Observation": 'https://www.ncqa.org/fhir/ValueSet/hedis-final-observation-h009'

context Patient

define "Observations": [Observation]

define "Final Observations":
    "Observations"."With status"("Final Observation")

define fluent function "With status"(observations List<FHIR.Observation>, statusCodes System.ValueSet):
    (observations) observation
    where observation.status in statusCodes
```
When executing "Final Observations" we would expect a return, however we are getting a "null".

antvaset:
Closes #1749.

For the `String` overload, the `InValueSet` operator should return `true` if the resolved value set contains a code with an equivalent `code` value. It should throw if the resolved value set contains codes from multiple code systems.


JPercival (approved):

JPercival (commented):
I would expect the "ambiguous code behavior" to be behind the terminology provider API, not implemented as part of the engine.

antvaset:
Superseded by https://github.com/cqframework/clinical-reasoning/pull/1026.

PR #1776, opened 2026-06-04
Feature/cql debug

raleigh-g-thompson:
## Summary

Adds a `BreakpointHandler` interface to the CQL engine for step-through debugging. External callers set `State.breakpointHandler` to observe expression evaluation and optionally pause execution.

## New files

| File | Purpose |
|------|---------|
| `BreakpointAction.kt` | `enum { CONTINUE, PAUSE }` |
| `BreakpointHandler.kt` | Interface with 5 callbacks (all default no-op) |
| `BreakpointHandlerTest.kt` | 7 tests covering all callback paths |
| `BreakpointHandlerTest.cql` | Test fixture |

## Modified files

- **EvaluationVisitor.kt** — `onBeforeExpression`/`onAfterExpression` fire on every ELM node; `onExpressionDefEvaluated` fires on the define after its expression completes
- **State.kt** — `var breakpointHandler: BreakpointHandler?`

## Key fix

`onExpressionDefEvaluated` solves the define-value tracking gap: `onAfterExpression` fires on intermediate child nodes (Property, Equal, etc.), never on the `ExpressionDef` itself. The new callback always fires on the define with its final value.


raleigh-g-thompson:
compared to wrong base

PR #1786, opened 2026-06-08
Feature/add debugger support

### LINKED ISSUE #1783: Add breakpoint handler hooks for debugger support in the CQL engine
The CQL engine currently has no API for external tools (IDE debug adapters,
step-through debuggers) to observe or control expression evaluation at runtime.
The existing `DebugMap`/source-locator system is a post-hoc tracing mechanism —
it records what happened after evaluation completes, but provides no way to:

- Pause execution at a specific expression or define
- Inspect live variable bindings and intermediate values mid-evaluation
- Step into function calls or step over sub-expressions
- Identify what call site triggered a given define or function invocation

This makes it impossible to build a proper interactive debugger for CQL
(e.g., a VS Code Debug Adapter) on top of the engine.

raleigh-g-thompson:
### Summary

Adds a `BreakpointHandler` interface to the CQL engine that enables external debugger
tooling (e.g., VS Code debug adapters) to observe and control CQL evaluation at runtime.
Callers implement the interface and attach it to `engine.state.breakpointHandler` before
evaluation.

closes https://github.com/cqframework/clinical_quality_language/issues/1783

### New API — `BreakpointHandler` interface (`engine/debug/`)

| Callback | When called |
|---|---|
| `onBeforeExpression(elm, state) -> BreakpointAction` | Before every ELM expression evaluates. Return `PAUSE` to suspend execution. |
| `onAfterExpression(elm, state, value)` | After every ELM expression evaluates, with the result value. |
| `onExpressionDefEntered(elm, callSite, state)` | When a `define` or function body is entered. `callSite` is the `ExpressionRef` or `FunctionRef` that triggered the entry (or `null` for top-level entries). |
| `onExpressionDefEvaluated(elm, state, value)` | When a `define` or function body evaluation completes. |
| `waitForResume()` / `release()` | Pause/resume flow control for stepping. |

A `BreakpointAction` enum (`CONTINUE`, `PAUSE`) controls whether execution proceeds
or blocks until `release()` is called.

### Implementation details

- **`EvaluationVisitor.kt`**: `onBeforeExpression`/`onAfterExpression` hooked into
  `internalEvaluate` (the central expression dispatch);
  `onExpressionDefEntered`/`onExpressionDefEvaluated` wired into `visitExpressionDef`
  and `visitExpressionRef`.
- **`FunctionRefEvaluator.kt`**: Callbacks added around function body evaluation, plus
  `currentCallSite` tracking so the caller can identify which `FunctionRef` caused the
  invocation.
- **`State.kt`**: New `breakpointHandler` property and `@Volatile var currentCallSite`
  field for cross-visitor call-site propagation.

### Additional fix

**`RetrieveEvaluator.kt`**: Filters out `null` elements when the `codes` argument is
`Iterable<*>`, preventing NPEs during data retrieval when a code list contains null
entries. Also handles mixed `String`/`Code` iterables by wrapping strings appropriately.
Addresses the spec at
https://cql.hl7.org/04-logicalspecification.html#retrieve.

### Tests

`BreakpointHandlerTest.kt` validates:

- `onBeforeExpression` fires for every expression (not just defines)
- Live `State` is accessible with a non-empty stack
- Evaluation completes correctly with `CONTINUE` and produces expected results
  (`X=3`, `Y=12`, `Z=6`)
- `onAfterExpression` receives correct intermediate values (Add → 3, Multiply → 12)
- Pause/resume flow control blocks and unblocks correctly via `CompletableFuture`
- Independence from the existing `DebugMap`/tracing system
- `onExpressionDefEntered` fires for both `ExpressionRef` and `FunctionRef` call sites
- `onExpressionDefEntered`/`onExpressionDefEvaluated` are balanced (same set of defs)
- Compatible with `EnableDetailedTracing`

### DIFF
--- a/Src/java/engine/src/commonMain/kotlin/org/opencds/cqf/cql/engine/debug/BreakpointAction.kt
+++ b/Src/java/engine/src/commonMain/kotlin/org/opencds/cqf/cql/engine/debug/BreakpointAction.kt
@@ -1,0 +1,6 @@
+package org.opencds.cqf.cql.engine.debug
+
+enum class BreakpointAction {
+    CONTINUE,
+    PAUSE,
+}
--- a/Src/java/engine/src/commonMain/kotlin/org/opencds/cqf/cql/engine/debug/BreakpointHandler.kt
+++ b/Src/java/engine/src/commonMain/kotlin/org/opencds/cqf/cql/engine/debug/BreakpointHandler.kt
@@ -6,8 +6,12 @@ import org.opencds.cqf.cql.engine.execution.State
 
 interface BreakpointHandler {
     fun onBeforeExpression(elm: Element, state: State): BreakpointAction
+
     fun onAfterExpression(elm: Element, state: State, value: Any?) {}
+
     fun onExpressionDefEvaluated(elm: ExpressionDef, state: State, value: Any?) {}
+
     fun waitForResume() {}
+
     fun release() {}
 }
--- a/Src/java/engine/src/commonMain/kotlin/org/opencds/cqf/cql/engine/debug/BreakpointHandler.kt
+++ b/Src/java/engine/src/commonMain/kotlin/org/opencds/cqf/cql/engine/debug/BreakpointHandler.kt
@@ -1,0 +1,13 @@
+package org.opencds.cqf.cql.engine.debug
+
+import org.hl7.elm.r1.Element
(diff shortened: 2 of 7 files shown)


brynrhodes (approved):

bryantaustin13 (commented):
LGTM

bryantaustin13 (approved):

Issue #1788, opened 2026-06-17
Difference in weeks calculating incorrectly


bryantaustin13:
"expression": "difference in weeks between @2012-03-10T22:05:09 and @2012-03-24T07:19:33",
This returns 2 when it should be 1.

https://cql.hl7.org/09-b-cqlreference.html#difference states:
"The difference-between operator returns the number of boundaries crossed for the specified precision between the first and second arguments. If the first argument is after the second argument, the result is negative. The result of this operation is always an integer; any fractional boundaries are dropped."

From March 10 22:05:09 to March 23 22:05:09 = 13 days
From March 23 22:05:09 to March 24 07:19:33 = 9 hours 14 minutes 24 seconds
13 days 9 hours 14 minutes 24 seconds Which is not past the 2nd boundary, so it should be 1.

## How can we reproduce it?
Run the following in cql-tests-runner:
```
<?xml version="1.0" encoding="utf-8"?>
<tests xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://hl7.org/fhirpath/tests" xsi:schemaLocation="http://hl7.org/fhirpath/tests ../../testSchema/testSchema.xsd"
       name="CqlTypeOperatorsTest">
	<group name="SingleTest" version="1.0">
		<test name="DateTimeDifferenceWeeks3" version="1.4">
			<capability code="difference-in"/>
			<expression>difference in weeks between @2012-03-10T22:05:09 and @2012-03-24T07:19:33</expression>
			<output>1</output>
		</test>
	</group>
</tests>
```
It will fail as actual returned is 2.

Issue #1765, opened 2026-05-07
DateTime Operator "Add" for year and weeks treating 52 weeks as a calendar year


bryantaustin13:
## What happened?
DateTime arithmetic with weeks is incorrect. The engine appears to treat 52 weeks as 1 calendar year rather than 364 days, producing incorrect results, especially around leap-day cases.

## How can we reproduce it?

run this test in cql-tests-runner:
```
<?xml version="1.0" encoding="utf-8"?>
<tests xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://hl7.org/fhirpath/tests" xsi:schemaLocation="http://hl7.org/fhirpath/tests ../../testSchema/testSchema.xsd"
	   name="CqlDateTimeOperatorsTest">
	<group name="SingleTest" version="1.0">
		<test name="DateTimeAddYearInWeeks" version="1.4">
			<capability code="date-time-arithmetic"/>
			<expression>DateTime(2018, 5, 23) + 52 weeks</expression>
			<output>DateTime(2019, 5, 22)</output>
			<!-- EXPECT: The resultant day of the week to match the original day of the week. -->
			<!-- 2018-05-23 is a Wednesday -->
			<!-- 2019-05-22 is a Wednesday -->
		</test>
	</group>
</tests>
```
It returns false instead of true. These tests are all the same:
DateTimeAddYearInWeeks
DateTimeAddYearInWeeks
DateTimeLeapDayAddYearInWeeks

PR #1790, opened 2026-07-01
Fix DateTime add/subtract of weeks: 52 weeks is 364 days, not a calendar year

c-schuler:
### Problem

`DateTime`/`Date` arithmetic drifted by a day whenever a quantity of **52 or more weeks** was added or subtracted. For example:

`DateTime(2018, 5, 23) + 52 weeks   // returned 2019-05-23, should be 2019-05-22`

52 weeks is a fixed 364 days (and should preserve the day of week - Wed -> Wed), but the engine was returning a result 365 days later (the calendar anniversary).

### Root cause

TemporalHelper.weeksToDays special-cased blocks of 52 weeks as 365-day calendar years:

```
if (weeks >= 52) { years = weeks / 52; weeks -= years * 52 }
return weeks * 7 + (years * 365)
```

This is correct for weeks < 52 (weeks * 7) but jumps by an extra day at every 52-week boundary (51 weeks → 357 days, 52 weeks → 365). A week is a fixed 7-day duration, so the year rollup has no basis in the spec.

### Fix

fun weeksToDays(weeks: Int): Int = weeks * 7

Weeks are converted to days and added to the day component, letting the existing date logic walk the calendar (so leap-day/leap-year cases land correctly).

This matches the normative spec - CQL Author's Guide, Date/Time Arithmetic:

    "The week, positive or negative, is multiplied by 7, and the resulting value to the day component, respecting calendar month and calendar year lengths."

(and the reference's "1 week = 7 days"), rather than any "52 weeks ~ 1 year" assumption.

Validation

- All six DateTime add/subtract-weeks conformance tests (cql-tests) now pass, including the leap-day and leap-year variants (DateTimeAddYearInWeeks, DateTimeLeapDayAddYear, DateTimeLeapYearAddYearInWeeks, and the Subtract counterparts).
- Full :engine:jvmTest suite passes with no regressions.

Scope

Fixes the add/subtract weeks path only. The "difference in weeks" issue (#1788 separate operator (DifferenceBetween) and is not addressed here.

Closes #1726, #1765


antvaset (approved):
Nice!

JPercival (approved):

PR #1793, opened 2026-07-06
Fix equality of empty lists ({} = {} -> true) (#1721)

c-schuler:
## Problem

`{} = {}` returned `null` instead of `true` (and `{} != {}` returned `null` instead of `false`). Per the spec, the `=` operator for lists returns true when the lists "have the same elements by value, in the same order" — two empty lists are vacuously equal.

## Root cause

`EqualEvaluator.listsEqual` had an over-broad guard that returned `null` whenever *either* list was empty:

```kotlin
if (!leftIterator.hasNext() || !rightIterator.hasNext()) {
    return null
}
```
This conflated "empty list" with "null". It was introduced in the Equal/Equivalent refactor (#1668). The existing loop/tail logic already handles empty lists correctly, so the guard was both unnecessary and wrong.

Fix

Removed the guard. Behavior is now:

| expression | result |
|---|---|
| {} = {}  | true |
| {} != {} |  false|
| {null} = {null}  | true (nulls considered equal |
|  {} = null / null = {} | null (null operand) |

Null-element handling is unchanged

The removed guard only fired for empty lists (zero elements); lists that contain nulls have hasNext() == true, so they were never affected. The existing element-wise logic continues to implement the spec's two clauses correctly - a null matched against another null is equal, a null against a non-null (or a trailing null on a length mismatch) yields null:

| expression            | result |
|-----------------------|--------|
| {1, null} = {1, null} | true   |
| {1, null} = {1, 2}    | null   |
| {1} = {1, null}       | null   |
| {1} = {1, 2}          | false  |

EquivalentEvaluator.listEquivalent has no analogous guard, so this bug was isolated to EqualEvaluator.

Engine test corrections

The engine's own tests asserted the buggy null for {} = {}; updated to match the conformance suite (EqualEmptyListAndEmptyList -> true, NotEqualEmptyAndEmpty -> false):

- ListOperatorsTest.kt: EqualEmptyListAndEmptyList, NotEqualEmptyAndEmpty.
- CqlInternalTypeRepresentationSuiteTest.kt: equal({}, EMPTY_LIST) (parameterized over timezones).

Validation

- Full :engine:jvmTest: 228 tests, 0 failures.
- CQL Tests conformance: EqualEmptyListAndEmptyList and NotEqualEmptyAndEmpty now pass; no regressions.

Fixes #1721


c-schuler:
Closing this as the cql-tests test is found to be in error: https://github.com/cqframework/cql-tests/issues/129

PR #1805, opened 2026-07-09
Fix ClassCastException in cql-to-elm-cli with modelinfo

### LINKED ISSUE #1803: ClassCastException in cql-to-elm-cli when using modelinfo
## What happened?

Running cql-to-elm-cli while specifying a ModelInfo with `--model` always results in a ClassCastException:

```
Exception in thread "main" java.lang.ClassCastException: class org.cqframework.cql.cql2elm.cli.Main$$Lambda$13/0x000000700101f340 cannot be cast to class org.hl7.cql.model.ModelInfoProvider (org.cqframework.cql.cql2elm.cli.Main$$Lambda$13/0x000000700101f340 and org.hl7.cql.model.ModelInfoProvider are in unnamed module of loader 'app')
 	at org.cqframework.cql.cql2elm.cli.Main.getModelInfoProvider(Main.kt:46)
 	at org.cqframework.cql.cql2elm.cli.Main.main(Main.kt:330)
```

## What did you expect to happen?

Translation should succeed (assuming otherwise valid inputs)

## How can we reproduce it?
Run the cql-to-elm CLI with any input and any valid modelinfo file. For an example with resources already in this repo:
```
cd Src/java
./gradlew :cql-to-elm-cli:installDist
./cql-to-elm-cli/build/install/cql-to-elm-cli/bin/cql-to-elm-cli --input ../../Examples/2zc_VTE3.cql --model ./quick/src/main/resources/org/hl7/fhir/fhir-modelinfo-4.0.1.xml
```

## Environment

- Version / commit: latest on main as of today `313da315`
- OS / runtime: macOS 26.5, JDK 17.0.14

dehall:
This PR fixes #1803 - a ClassCastException when running the CLI and specifying a modelinfo.

The short explanation is that in Kotlin, unlike Java, anonymous lambdas cannot be cast to interfaces unless they are explicitly marked as functional interfaces (`fun interface`). So an alternative fix would be to mark `ModelInfoProvider` as a functional interface, but that may have broader implications. This approach instead creates an anonymous class that explicitly implements the interface.

To test this, run the cli and specify a modelinfo, eg here's an example with files already in the repo:

```sh
./gradlew :cql-to-elm-cli:installDist
./cql-to-elm-cli/build/install/cql-to-elm-cli/bin/cql-to-elm-cli  --model ./quick/src/main/resources/org/hl7/fhir/fhir-modelinfo-4.0.1.xml --input ./cql-to-elm/src/jvmTest/resources/org/cqframework/cql/cql2elm/fhir/r4/TestBundle.cql

```


dehall:
Apologies, I'm pretty sure the CI is all failing because this PR comes from a fork but I'm not sure what to do about it. I'm open to suggestions

antvaset (approved):

antvaset:
@dehall Looks good to me 👍  The PR only needs another approval and then it can be merged.

JPercival (approved):

cmoesel:
@brynrhodes @JPercival @antvaset - There are some failing CI, but they seem related to permissions/env issues, likely because the PR is coming from a branch. Can I force merge this or is there something real that we should be concerned about in the CI?

PR #1812, opened 2026-07-12
KMP project structure for `engine-fhir`

antvaset:
* Adopt KMP project structure for `engine-fhir`.
* Reuse `XjcTask` and configuration across the `engine-fhir` (KMP) and `quick` (non-KMP) projects.
* Link the output folder of the `xjc` task to the Java source sets declaratively in `engine-fhir` and `quick`.


engine-fhir/build.gradle.kts
```diff
@@ -1,42 +1,42 @@
...
-        schema = "${projectDir}/src/test/resources/org/hl7/fhirpath/testSchema"
-        extraArgs = listOf("-npa", "-p", "org.hl7.fhirpath.tests")
+        schema.set("${projectDir}/src/jvmTest/resources/org/hl7/fhirpath/testSchema")
+        extraArgs.set(listOf("-npa", "-p", "org.hl7.fhirpath.tests"))
+    }
+
+kotlin {
+    js { outputModuleName = "engine-fhir" }
+
+    @OptIn(ExperimentalWasmDsl::class) wasmJs { outputModuleName = "engine-fhir" }
```
c-schuler: I don't think we actually want to apply the KMP convention here as engine-fhir does not have JS sources or a commonMain. We want the layout here, but not the targets, right?

engine-fhir/build.gradle.kts
```diff
@@ -1,42 +1,42 @@
...
-    classDirectories.setFrom(
-        files(
-            "${projectDir}/../elm/build/classes/kotlin/jvm/main",
-            "${projectDir}/../cql-to-elm/build/classes/kotlin/jvm/main",
-            "${projectDir}/../engine/build/classes/kotlin/main",
-            "${projectDir}/../engine-fhir/build/classes/kotlin/main",
-        )
-    )
+java {
+    sourceSets { named("jvmMain") { java { srcDir(generateFhirPathTests.map { it.outputDir }) } } }
```
c-schuler: Do we want test schema classes wired into jvmMain and shipped in the production jar?

build-logic/src/main/kotlin/cql.fhir-kotlin-multiplatform-conventions.gradle.kts
```diff
@@ -0,0 +1,48 @@
+plugins { id("cql.kotlin-multiplatform-conventions") }
```
c-schuler: This is just a copy-paste of cql.fhir-conventions if I am not mistaken. Kind of contradicts the stated purpose of the PR:

> Reuse XjcTask and configuration across the engine-fhir (KMP) and quick (non-KMP) projects.

c-schuler (changes requested):
Looks mostly good. A couple of requested changes and questions below.

engine-fhir/build.gradle.kts
antvaset: Hi @c-schuler, I want to use `commonMain` inside the `engine-fhir` project for future multiplatform engine/FHIR code (coming after this PR). Those two JS/WASM lines will come into play when `commonMain` is in place.

engine-fhir/build.gradle.kts
antvaset: I went back and forth on this: The `cql.xjc-conventions` plugin currently does this currently, for both autogenerated classes and JAXB runtime dependencies:
```kt
dependencies {
  // ...
  api("jakarta.xml.bind:jakarta.xml.bind-api:4.0.1")
  api("codes.rafael.jaxb2_commons:jaxb2-basics-runtime:3.0.0")
}
```
I've kept it that way for avoid changing anything for Java consumers. The plugin does it that way because the autogenerated classes are indeed needed in the main JAR in the `quick` project.

build-logic/src/main/kotlin/cql.fhir-kotlin-multiplatform-conventions.gradle.kts
antvaset: The bulk of it copy-paste, but the new `cql.fhir-kotlin-multiplatform-conventions` plugin is KMP-specific. The HAPI FHIR dependencies are declared inside different blocks/scopes compared to the JVM-only `cql.fhir-conventions` plugin. It was possible to reuse the `xjc` configuration but not the dependencies section for HAPI FHIR.

engine-fhir/build.gradle.kts
c-schuler: I see, that makes sense. The only concern I would have is if a release is scheduled before the follow-up. If not, then I am happy to treat this as a non-blocker. We should avoid shipping empty engine-fhir-js / engine-fhir-wasm-js to Maven Central though.

build-logic/src/main/kotlin/cql.fhir-kotlin-multiplatform-conventions.gradle.kts
c-schuler: That's a fair point and you're right that the scoping differs... However the contents are still identical and shareable, right? The jetty resolutionStrategy, the bom, the hapi-fhir-structures-* deps with the same excludes, and the xpp3 pin are byte-for-byte the same in both plugins, and that list is exactly what will drift when we bump HAPI or add a structures module. It seems like what you did for cql.xjc-common-conventions could be used here (a shared plugin both variants apply with each supplying their own thin wrapper).

I'll defer to you on whether a third plugin is worth removing one duplicated list, but I don't think the differing wrapper block makes the contents un-shareable.

engine-fhir/build.gradle.kts
c-schuler: Agreed that for quick the generated classes belong in main and I understand wanting to preserve behavior for Java consumers in a structure-only PR.

Note that engine-fhir’s generated classes differ: they originate from src/jvmTest/resources/.../testSchema, are placed in package org.hl7.fhirpath.tests, and are used only by the FHIRPath tests. Moving them into jvmMain would include test-only bindings in engine-fhir’s production jar. The "quick needs it in main" justification doesn’t apply, and keeping the previous cql.xjc-conventions would preserve an existing wart rather than fix it. Should these remain in jvmTest? I can split this into a follow-up if you prefer to keep this PR purely structural. I just want the placement to be a conscious decision, not inherited.

JPercival (approved):

engine-fhir/build.gradle.kts
antvaset: That makes sense. I've removed the `js` and `wasmJs` blocks in the latest commit.

engine-fhir/build.gradle.kts
antvaset: I've changed up the `xjc-conventions` plugin to only carry the common `xjc`  configuration and the runtime dependencies can be wired up where they are needed (`main`/`test` in JVM-only projects or `jvmMain`/`jvmTest` in KMP projects). In `engine-fhir`, the generated sources and runtime dependencies are scoped to `jvmTest` and in `quick`, they are scoped to `main`.

build-logic/src/main/kotlin/cql.fhir-kotlin-multiplatform-conventions.gradle.kts
antvaset: I've changed this up based on your comments and created the `cql.fhir-common-conventions` plugin that includes the shared `configurations` setting that is common across JVM and KMP setups. As for the `dependencies` block, it doesn't seem possible to cleanly deduplicate the content using the DSL only. The JVM-only `cql.fhir-conventions` plugin will also be removed once the KMP project structure is applied to the rest of the projects.

engine-fhir/build.gradle.kts
c-schuler: Thanks for removing the outputModuleName lines. However, those only set the JS module name; the js/wasm targets themselves come from cql.kotlin-multiplatform-conventions. So they're still active and a publishToMavenLocal --dry-run on this commit still shows:

> :engine-fhir:publishJsPublicationToMavenLocal
> :engine-fhir:publishWasmJsPublicationToMavenLocal

So we'd still ship empty engine-fhir-js / engine-fhir-wasm-js to Maven Central until the commonMain work lands.

Totally fine to keep the targets as groundwork. Could we just disable the two publications until there's real multiplatform source? Something like:

```kotlin
publishing {
    publications.matching { it.name == "js" || it.name == "wasmJs" }
        .configureEach { /* skip until engine-fhir has commonMain sources */ }
}
```

Apologies for holding things up...

antvaset: No problem. I've changed the `cql.kotlin-multiplatform-conventions` plugin to only register JS/WASM targets if the relevant sources are present.

c-schuler (approved):
Looks great!

PR #1820, opened 2026-08-06
Make `createHapiInstance()` method public in `engine-fhir`

antvaset:
This allows the `createHapiInstance()` method to be used in the `FhirParametersConverter` in CR.


c-schuler (changes requested):
This results in an NPE if an external caller provides an unresolvable type. Internally that was tolerable, but we should be more defensive now that it is a public API. Something like: `resolveType(typeName) ?: throw UnknownType("Could not resolve type $typeName")`

c-schuler (approved):

antvaset:
Thanks for checking this and approving @c-schuler 🙌

JPercival (approved):

JPercival (approved):

Issue #1768, opened 2026-05-27
@cqframework/cql: subpath exports ship .mjs with no "types", so CqlTranslator methods (e.g. toJson) aren't visible to consumers


aks129:
## Summary

The npm package `@cqframework/cql` (4.0.0-beta.1) defines subpath `exports` that map only to `.mjs` files, with no corresponding `types` entry. As a result, TypeScript consumers importing from `@cqframework/cql/cql-to-elm` do not get the type definitions in `kotlin/cql-to-elm.d.ts`, even though that file declares the full `CqlTranslator` surface.

In practice this means methods that genuinely exist at runtime — e.g. `CqlTranslator.toJson()` — are reported by `tsc` / the Angular compiler as not existing on the type, forcing downstream code into `as unknown as { ... }` casts.

## Reproduction

`node_modules/@cqframework/cql/package.json`:

```json
"exports": {
  "./cql": "./kotlin/cql.mjs",
  "./cql-to-elm": "./kotlin/cql-to-elm.mjs",
  "./elm": "./kotlin/elm.mjs",
  "./shared": "./kotlin/shared.mjs"
}
```

The matching `kotlin/cql-to-elm.d.ts` does declare:

```ts
export declare class CqlTranslator {
    toXml(elmLibraryWriterProvider?: any): string;
    toJson(elmLibraryWriterProvider?: any): string;   // <-- exists at runtime + in .d.ts
    // ...
}
```

Consumer code:

```ts
import { CqlTranslator } from '@cqframework/cql/cql-to-elm';
const t = CqlTranslator.fromText(cql, libraryManager);
t.toXml();   // OK
t.toJson();  // TS2339: Property 'toJson' does not exist on type 'CqlTranslator'
```

`toXml()` happens to resolve in some configs but `toJson()` does not, depending on module-resolution mode — both should resolve consistently from the published type definitions.

## Impact

CQL Studio (Angular 21, `moduleResolution` bundler) consumes `@cqframework/cql/cql-to-elm` for in-browser CQL→ELM translation. To call `toJson()` (needed to feed ELM JSON into a SQL-on-FHIR transpiler) we currently cast:

```ts
const elmJson = (translator as unknown as { toJson: () => string }).toJson();
```

This works but is fragile — a signature change upstream wouldn't be caught by the compiler.

## Suggested fix

Add `types` (or `typesVersions`) entries to the subpath exports so each `.mjs` is paired with its `.d.ts`, e.g.:

```json
[...]

PR #1821, opened 2026-08-12
Generate type definitions for the `@cqframework/cql` NPM package

antvaset:
See https://github.com/cqframework/clinical_quality_language/issues/1768#issuecomment-4898553304.

The solution is to generate type definitions for every module (subpath export) in the `@cqframework/cql` package. The generated `cql.exports.d.mts`, `cql-to-elm.exports.d.mts`, `engine.exports.d.mts`, etc. re-export relevant type declarations from `engine.d.mts` generated by the Kotlin/JS compiler.

The `:npm-cql:generateTypeDefinitions` task is integrated into the Gradle build. The final npm package looks like this https://github.com/antvaset/cql-npm/tree/05d945b.

### DIFF
--- a/build-logic/src/main/kotlin/GenerateTypeDefinitions.kt
+++ b/build-logic/src/main/kotlin/GenerateTypeDefinitions.kt
@@ -1,0 +1,64 @@
+import java.io.File
+import org.gradle.api.DefaultTask
+import org.gradle.api.file.DirectoryProperty
+import org.gradle.api.provider.ListProperty
+import org.gradle.api.provider.Property
+import org.gradle.api.tasks.Input
+import org.gradle.api.tasks.InputDirectory
+import org.gradle.api.tasks.OutputDirectory
+import org.gradle.api.tasks.TaskAction
+
+/** Generates TypeScript declaration files (.exports.d.mts) for the CQL NPM package. */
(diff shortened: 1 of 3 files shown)


JPercival (approved):

c-schuler (approved):
Very nice!

One thing I would point out is there is potential for ELM vs. runtime collisions here.  The core types each have two declarations: the ELM AST node and the runtime value type, which are flattened into one `engine.d.mts`. I verified with tsc 5.9: with skipLibCheck:false they error (TS2300); with skipLibCheck:true first-declaration-wins, so the runtime value types are silently unreachable via typings (importing Quantity from /engine resolves to the ELM Quantity). 

This is not necessarily in the context of this PR, but we should track as a separate issue at the least.

antvaset:
Thanks @c-schuler, I've wanted to address the clashing `@JsExport`-ed names after this is merged https://github.com/cqframework/clinical_quality_language/pull/1822 and make `@JsName` overrides a new configuration option for the `XsdKotlinGen` task. Kotlin's standard library actually exports `List` as `KtList` in `kotlin-kotlin-stdlib.mjs`, so we can similarly use `ElmList` for the ELM `List` and `CqlList` for the engine runtime `List`.

PR #1822, opened 2026-08-13
Improvements for the Gradle build and Kotlin code generation

antvaset:
- Replace the `cql.xsd-kotlin-multiplatform-gen-conventions` plugin with the `registerAntlrKotlinGrammarGeneration()` helper. The helper fixes the `@Suppress("UNSAFE_CALL")` annotations in the generated code alongside the task registration logic.
- Update the ANTLR Kotlin plugin and runtime.
- Replace the `FilesToStrings` task with `FileToString` that better tracks the task inputs and outputs using Gradle Properties.
- Update `XsdKotlinGenTask` to use Gradle input/output properties. The task is registered as `generateModelInfoKotlinSource` and `generateElmKotlinSource` in the `cql` and `elm` projects. The source sets point to the task outputs, so the tasks run as part of normal builds/compiles.


JPercival (approved):

c-schuler (approved):
Looks good!

One minor nit: build-logic/build.gradle.kts adds the antlr-kotlin plugin 1.0.13, but cql/build.gradle.kts pins the runtime 1.0.12.

antvaset:
Thanks @c-schuler I've made sure both are set to the same version.

PR #1830, opened 2026-08-26
Fix loading of compiled libraries

### LINKED ISSUE #1827: ELM JSON loading: isVersionCompatible compares the ELM translatorVersion against the loader's compatibilityLevel, so no library is ever version-compatible by default
**Found on:** the `kmp-fhir-providers` branch (5.1.0 KMP line); likely applies to the KMP path generally.

## Summary

When `LibraryManager` attempts to load precompiled ELM (content type JSON, tried before CQL compilation), the compatibility check compares the translator version recorded in the ELM's `CqlToElmInfo` annotation (e.g. `5.1.0`) against the compiler option `compatibilityLevel` (e.g. `1.5`). These are different quantities, so the comparison fails for every library — including one produced moments earlier by the same toolchain that is loading it. Expected: same-toolchain output should be loadable, or the check should compare like against like.

## Reproduction

[dhes/cql-v5-probe](https://github.com/dhes/cql-v5-probe) @ 25030bd, Part 3 of `Main.kt`: write ELM JSON with the branch's own writer, attempt to reload it with the same branch's `LibraryManager`.

## Why now

This is the first of three independent defects (filed separately) that together block precompiled-ELM execution — the `Library.content` model that clinical-reasoning and the DAK ecosystem use. With `cqf-fhir-cql` converting to Kotlin (cqframework/clinical-reasoning#1080) and the engine's KMP data-access work under way, whoever wires CR to the KMP engine will hit all three; filing them now so they're characterized before that work reaches them. Every KMP consumer currently works around the set by shipping CQL source and compiling at initialization.

### LINKED ISSUE #1829: ELM JSON loading: generateCompiledLibrary requires resolved def.resultType, which deserialization never rehydrates (and the writer omits resultTypeName on the auto-generated context def)
**Found on:** the `kmp-fhir-providers` branch (5.1.0 KMP line). Third — and fatal — member of the series blocking precompiled-ELM loading; see #1827 and #1828 for the first two.

## Summary

Even with #1827 and #1828 worked around, compiling the deserialized library fails: `generateCompiledLibrary` unconditionally dereferences each expression def's resolved `resultType`, but JSON deserialization only restores `resultTypeName` attributes — it does not rehydrate the resolved type objects — so the requirement can never be met from a round trip. Additionally, the writer does not emit `resultTypeName` on the auto-generated Patient context expression def, so the information is not even present in the serialized form.

This is the one that makes precompiled-ELM execution (the `Library.content` model clinical-reasoning and the DAK ecosystem rely on) impossible on the branch regardless of configuration.

## Suggested direction

Rehydrate result types from `resultTypeName` during deserialization (re-resolving against the ModelManager), or relax `generateCompiledLibrary` to tolerate unresolved result types when `EnableResultTypes` was not requested.

## Reproduction

[dhes/cql-v5-probe](https://github.com/dhes/cql-v5-probe) @ 25030bd, Part 3 of `Main.kt`. The writer-side omission is verifiable by inspecting the emitted JSON for the context def.

antvaset:
Closes #1827. Closes #1829.

* Move `TypeResolver` (converts ELM type names and type specifiers to CQL `DataType`s) from `elm-fhir` to `cql-to-elm` with added KMP support.
* Add `resultTypeName`/`resultTypeSpecifier` to ELM model context retrieve nodes by moving the `track()` call in the `Cql2ElmVisitor`. For those result types to be added to the ELM elements, `track()` needs to be called after the CQL result types are set.
* Fix `LibraryManager.isVersionCompatible()`. The compatibility check passes when the version of the currently running translator matches the version of the translator that the deserialized library was compiled with (the `translatorVersion` annotation value in the ELM library).
* Fix `LibraryManager.generateCompiledLibrary()`. To properly convert a deserialized ELM `Library` to a `CompiledLibrary`, the `resultTypeName`/`resultTypeSpecifier` of each ELM expression needs to be resolved into a CQL `DataType`. The library is only accepted if every expression has a resolvable result type. All models listed in the library's `usings` are loaded first (using model identifiers) to enable type resolution for ELM result types which use namespace URIs.
* Modernize `LibraryManagerTests` and `TranslationTests` along the way.

### DIFF
--- a/cql-to-elm/detekt-baseline.xml
+++ b/cql-to-elm/detekt-baseline.xml
@@ -1,10 +1,5 @@
 <?xml version='1.0' encoding='UTF-8'?>
 <SmellBaseline>
   <ManuallySuppressedIssues/>
-  <CurrentIssues>
-    <ID>MaxLineLength:LibraryManager.kt$LibraryManager$"Library ${library.identifier?.id} uses ${it.localIdentifier} model with URI ${it.uri} but resolved model has name ${model.modelInfo.name} and URI ${model.modelInfo.url}"</ID>
-    <ID>ReturnCount:TypeResolver.kt$TypeResolver$fun dataTypeToProfileQName(type: DataType?): QName?</ID>
-    <ID>ReturnCount:TypeResolver.kt$TypeResolver$fun getTypeUri(type: DataType?): String?</ID>
-    <ID>TooManyFunctions:TypeResolver.kt$TypeResolver</ID>
-  </CurrentIssues>
+  <CurrentIssues/>
 </SmellBaseline>
--- a/cql-to-elm/detekt-baseline.xml
+++ b/cql-to-elm/detekt-baseline.xml
@@ -1,0 +1,10 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<SmellBaseline>
+  <ManuallySuppressedIssues/>
+  <CurrentIssues>
+    <ID>MaxLineLength:LibraryManager.kt$LibraryManager$"Library ${library.identifier?.id} uses ${it.localIdentifier} model with URI ${it.uri} but resolved model has name ${model.modelInfo.name} and URI ${model.modelInfo.url}"</ID>
+    <ID>ReturnCount:TypeResolver.kt$TypeResolver$fun dataTypeToProfileQName(type: DataType?): QName?</ID>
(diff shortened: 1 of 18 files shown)


JPercival (approved):

c-schuler (approved):
Nice work!

PR #1833, opened 2026-09-01
Fix overrides in `R4FhirModelResolver.resolveType()`

antvaset:
* Fix type name overrides in `R4FhirModelResolver.resolveType()`. Previously, this method incorrectly returned `org.hl7.fhir.r4.model.CodeType` for the `MedicationAdministrationStatus`, `MedicationDispenseStatus`, `MedicationKnowledgeStatus`, and `Messageheader_Response_Request` type names. In HAPI FHIR R4, there are dedicated enums corresponding to these types, which should be returned instead:
```
org.hl7.fhir.r4.model.MedicationAdministration.MedicationAdministrationStatus
org.hl7.fhir.r4.model.MedicationDispense.MedicationDispenseStatus
org.hl7.fhir.r4.model.MedicationKnowledge.MedicationKnowledgeStatus
org.hl7.fhir.r4.model.MessageDefinition.MessageheaderResponseRequest
```
Without the overrides, the first three are resolved correctly (using same resolution logic as, e.g. `FHIR.AdministrativeGender`). `FHIR.Messageheader_Response_Request` still needs to be manually mapped to the corresponding HAPI FHIR R4 enum.
* Update tests to include result checks for `resolveType()` calls.


lukedegruchy (approved):
This is tested against cqis-spark and I see no more errors on AAB-Details.

c-schuler (approved):
Approved with a note that these issues also occur in the R5 model resolver - we should endeavor to capture that and apply the fix there as well.

PR #1834, opened 2026-09-02
Key evaluated resources by IDs

antvaset:
Currently, the evaluated resources are tracked in a `HashSet<Value>` where the hash needs to be built for every added resouce.

This PR changes `ExpressionResult.evaluatedResources` from `Set<Value>` to `Map<kotlin.String, Value>` where the resources are keyed by their string IDs.

For every evaluated resource, the ID is resloved using `ModelResolver.resolveId(evaluatedResource)`. For FHIR resources, the model resolver returns the [`Resource.id`](https://hl7.org/fhir/R4/resource-definitions.html#Resource.id) value.

### DIFF
--- a/engine/src/commonMain/kotlin/org/opencds/cqf/cql/engine/execution/State.kt
+++ b/engine/src/commonMain/kotlin/org/opencds/cqf/cql/engine/execution/State.kt
@@ -140,7 +140,11 @@ constructor(
      */
     @Volatile var currentCallSite: Element? = null
 
-    private val evaluatedResourceStack = ArrayDeque<MutableSet<Value?>>()
+    /**
+     * In each entry of the stack, the resources are keyed by their string IDs. Previously, the
+     * evaluated resources were stored in a set, which caused poor performance.
+     */
+    private val evaluatedResourceStack = ArrayDeque<MutableMap<kotlin.String, Value>>()
 
     val parameters = mutableMapOf<kotlin.String, Value?>()
     var contextValues = mutableMapOf<kotlin.String, kotlin.String?>()
@@ -502,13 +506,10 @@ constructor(
             return null
         }
 
-    val evaluatedResources: MutableSet<Value?>?
+    val evaluatedResources: Map<kotlin.String, Value>
         get() {
-            check(!evaluatedResourceStack.isEmpty()) {
-                "Attempted to get the evaluatedResource stack when it's empty"
-            }
-
             return this.evaluatedResourceStack.firstOrNull()
+                ?: error("Attempted to get the evaluatedResource stack when it's empty")
         }
 
     fun clearEvaluatedResources() {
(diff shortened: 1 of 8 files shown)


engine/src/commonMain/kotlin/org/opencds/cqf/cql/engine/execution/State.kt
```diff
@@ -537,14 +538,40 @@ constructor(
...
     }
 
+    /**
+     * Adds the resource to the top of the evaluated resource stack, keyed by its string ID. The
+     * resource is only stored if it is not null and has a non-null ID.
+     */
+    fun saveEvaluatedResource(resource: Value?) {
+        if (resource != null) {
+            val dataProvider = environment.resolveDataProvider(resource)
+            val id = dataProvider?.resolveId(resource)
```
c-schuler, line 548: Bare logical id as the map key collides across resource types. For instance Patient/123 and Condition/123. We really want a type-qualified ID here

engine/src/commonMain/kotlin/org/opencds/cqf/cql/engine/execution/State.kt
```diff
@@ -537,14 +538,40 @@ constructor(
         carryOverEvaluatedResourcesUpCallStack()
     }
 
+    /**
+     * Adds the resource to the top of the evaluated resource stack, keyed by its string ID. The
+     * resource is only stored if it is not null and has a non-null ID.
+     */
+    fun saveEvaluatedResource(resource: Value?) {
```
c-schuler, line 545: Also an issue with id-less resources now being dropped.

c-schuler (changes requested):
Looking good, but some definite issues to work out. Noted those below. Additionally, there is a breaking change for `ExpressionResult.evaluatedResources` API that should be tested downstream with clinical-reasoning.

engine/src/commonMain/kotlin/org/opencds/cqf/cql/engine/execution/State.kt
c-schuler: If we want a custom class, a small key class might make for a nice contract for downstream projects. Something like: 

```java
/** Uniquely identifies an evaluated resource across types (and models). */
@JsOnlyExport
data class ResourceKey(
    val model: kotlin.String,   // namespace URI, e.g. http://hl7.org/fhir
    val type: kotlin.String,    // e.g. "Patient"
    val id: kotlin.String,      // logical id, e.g. "123"
)
```

Build it in the provider, with a working default (no downstream break)

```java
// DataProvider interface — default composes existing pieces, so no implementer has to change:
fun resolveResourceKey(target: Value?): ResourceKey? {
    val namedType = getNamedTypeForCqlValue(target) ?: return null
    val id = resolveId(target) ?: return null
    return ResourceKey(namedType.getNamespaceURI(), namedType.getLocalPart(), id)
}
```

Then State.saveEvaluatedResource just does:

```java
fun saveEvaluatedResource(resource: Value?) {
    if (resource == null) return
    val key = environment.resolveDataProvider(resource)?.resolveResourceKey(resource) ?: return
    evaluatedResourceStack.first()[key] = resource
}
```

antvaset:
Superseded by https://github.com/cqframework/clinical_quality_language/pull/1837

PR #1837, opened 2026-09-04
Track evaluated resources by identity, leaving deduplication to the caller

c-schuler:
### Problem

Membership in the evaluated-resource set was decided by `Value.equals`, which for a structured model value is a deep comparison of the whole element tree. The cost is proportional to resource size and dominates retrieve-heavy evaluation.

### Why not dedupe by resource id

Deduplicating here requires the engine to know what makes two model values the same record. Neither ELM nor the ModelInfo defines that, so any key the engine picked (like FHIR's `Resource.id`) would be a model-specific rule compiled into model-agnostic code. `ModelResolver.resolveId` exists, but it isn't derived from the ModelInfo either.

### What this does

`State.pushEvaluatedResourceStack()` now allocates an `IdentitySet`, backed by the existing `createIdentityHashMap()` expect/actual so JVM, JS and wasm are all covered. The engine collects what was retrieved and takes no position on identity. Callers know their model and deduplicate on the appropriate key.

`ExpressionResult.evaluatedResources` keeps its `MutableSet<Value?>` type - **no API or ABI change**, so existing consumers link unchanged.

### Behaviour change

A record retrieved by several expressions now appears once per retrieve rather than once per record. The FHIR parser builds a fresh `ClassInstance` per retrieve, so identity collapses only literal re-adds of the same object.

Measured on `EvaluatedResourcesMultiLibComplexDepsTest`: 15 evaluated resources become 105 across 7 retrieves. Those are references, not resources (the objects already exist), but downstream must collapse them.

### Companion change

clinical-reasoning already deduplicates by resource type and logical id, and https://github.com/cqframework/clinical-reasoning/pull/1100 makes that a keyed lookup rather than an O(n^2) pairwise scan. **The two should land together**: this engine change with an unmodified CR is correct but runs that scan over ~7x the entries.

### DIFF
--- a/engine/src/commonMain/kotlin/org/opencds/cqf/cql/engine/execution/State.kt
+++ b/engine/src/commonMain/kotlin/org/opencds/cqf/cql/engine/execution/State.kt
@@ -140,7 +140,11 @@ constructor(
      */
     @Volatile var currentCallSite: Element? = null
 
-    private val evaluatedResourceStack = ArrayDeque<MutableSet<Value?>>()
+    /**
+     * In each entry of the stack, the resources are keyed by their string IDs. Previously, the
+     * evaluated resources were stored in a set, which caused poor performance.
+     */
+    private val evaluatedResourceStack = ArrayDeque<MutableMap<kotlin.String, Value>>()
 
     val parameters = mutableMapOf<kotlin.String, Value?>()
     var contextValues = mutableMapOf<kotlin.String, kotlin.String?>()
@@ -502,13 +506,10 @@ constructor(
             return null
         }
 
-    val evaluatedResources: MutableSet<Value?>?
+    val evaluatedResources: Map<kotlin.String, Value>
         get() {
-            check(!evaluatedResourceStack.isEmpty()) {
-                "Attempted to get the evaluatedResource stack when it's empty"
-            }
-
             return this.evaluatedResourceStack.firstOrNull()
+                ?: error("Attempted to get the evaluatedResource stack when it's empty")
         }
 
     fun clearEvaluatedResources() {
@@ -517,7 +518,7 @@ constructor(
     }
 
     fun pushEvaluatedResourceStack() {
-        evaluatedResourceStack.addFirst(HashSet<Value?>())
+        evaluatedResourceStack.addFirst(mutableMapOf())
     }
 
     /**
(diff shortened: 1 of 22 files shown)


antvaset (approved):

engine/src/commonMain/kotlin/org/opencds/cqf/cql/engine/util/IdentitySet.kt
```diff
@@ -0,0 +1,25 @@
...
+ * Backed by [createIdentityHashMap], which supplies a true identity map on the JVM.
+ */
+class IdentitySet<E> : AbstractMutableSet<E>() {
+
+    private val backing: MutableMap<E, Boolean> = createIdentityHashMap()
+
+    override val size: Int
+        get() = backing.size
+
+    override fun add(element: E): Boolean = backing.put(element, true) == null
```
lukedegruchy, line 16: I think there would be value in adding a direct unit test on this class

engine/src/commonMain/kotlin/org/opencds/cqf/cql/engine/execution/State.kt
```diff
@@ -517,7 +518,18 @@ constructor(
...
+        // defines that -- any key the engine picked (FHIR's Resource.id, say) would be a
+        // model-specific rule compiled into model-agnostic code. The previous HashSet decided
+        // membership with Value.equals, a deep comparison of the whole element tree on every
+        // insert, which is what made retrieve-heavy evaluation expensive.
+        //
+        // The retrieve path builds a fresh value per retrieve, so identity collapses only
+        // literal re-adds of the same object; a record retrieved by several expressions appears
+        // once per retrieve. Callers know their model and deduplicate on the right key --
+        // clinical-reasoning does so by resource type and id when it builds a MeasureReport.
+        evaluatedResourceStack.addFirst(IdentitySet<Value?>())
```
lukedegruchy, line 532: I think there's value in writing a unit test against this code path that fails if run against main production code

engine/src/commonMain/kotlin/org/opencds/cqf/cql/engine/util/IdentitySet.kt
lukedegruchy: I see you did this on a new commit.

engine/src/commonMain/kotlin/org/opencds/cqf/cql/engine/execution/State.kt
lukedegruchy: Looks like new tests were added.

lukedegruchy (approved):

